PR #5, opened 2020-06-12
Protect against call to setState after unmount

wilmaier:


calebmshafer:
@wilmaier can you run `rush change`?  Once this is merged I can do the first publish from this repo.

PR #10, opened 2020-07-01
Update to support dark/light themes

bsteinbk:
Replace hard-coded color specifications with themeable color specifications.


calebmshafer:
LGTM

PR #20, opened 2020-07-17
Update readme

archana-maharjan:
This PR contains the update in ReadMe file that was missed in previous PR, and also fixed the Repository Url for markup-frontstage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,12 @@ With the repository containing many different packages related to iModel.js, eac
 
 List of packages within this repository:
 
-| Name                         | Folder                 |
-| ---------------------------- | ---------------------- |
-| @bentley/imodel-select-react | packages/imodel-select |
-| @bentley/tree-widget-react   | packages/tree-widget   |
-|                              |                        |
+| Name                             | Folder                     |
+| -------------------------------- | -------------------------- |
+| @bentley/imodel-select-react     | packages/imodel-select     |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,11 +8,12 @@ With the repository containing many different packages related to iModel.js, eac
...
 List of packages within this repository:
 
-| Name                         | Folder                 |
-| ---------------------------- | ---------------------- |
-| @bentley/imodel-select-react | packages/imodel-select |
-| @bentley/tree-widget-react   | packages/tree-widget   |
-|                              |                        |
+| Name                             | Folder                     |
+| -------------------------------- | -------------------------- |
+| @bentley/imodel-select-react     | packages/imodel-select     |
```
aruniverse, line 13: @calebmshafer should we start linking to the readmes of the pkgs like so:
```
| Name                         | Folder                                                     |
| ---------------------------- | ---------------------------------------------------------- |
| @bentley/imodel-select-react | [packages/imodel-select](packages/imodel-select/README.md) |

calebmshafer: Yea, that sounds like a good suggestion to me.  Save people a few clicks

aruniverse: its not technically the folder, but leads to where they can find out more about the pkg. we could also add npm badges right next to it. im thinking like how rush does it with their readme https://github.com/microsoft/rushstack/blob/master/README.md

archana-maharjan: Updated please!

PR #21, opened 2020-07-20
Fixing inf recursion isssue in property grid data provider

zjzoltek:


packages/property-grid/src/api/PropertyGridDataProvider.ts
```diff
@@ -8,10 +8,11 @@ import { PropertyCategory, PropertyData } from "@bentley/ui-components";
 import { PresentationPropertyDataProvider } from "@bentley/presentation-components";
 export class PropertyDataProvider extends PresentationPropertyDataProvider {
   // tslint:disable-next-line:naming-convention
+  private _parentIsFieldFavorite = this.isFieldFavorite;
```
aruniverse, line 11: The fix was just moving this line up?

zjzoltek: yup. I didn't realize that it was caching the parent's version of the `isFieldFavorite` function, so when I moved it below the `isFieldFavorite` override it was no longer caching the parent version, but the child version (which relies on the parent version)

PR #23, opened 2020-07-28
Tree fixes

diegopinate:
1. Added invert functionality to categories tree to be consistent with models tree
2. Added forced 100% height on content wrapper to avoid issues when using iframes to host design review in Teams.

### DIFF
--- a/packages/tree-widget/src/components/trees/CategoriesTree.tsx
+++ b/packages/tree-widget/src/components/trees/CategoriesTree.tsx
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
- * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
- * See LICENSE.md in the project root for license terms and full copyright notice.
- *--------------------------------------------------------------------------------------------*/
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
 import * as React from "react";
 import {
   IModelApp,
--- a/packages/tree-widget/src/components/trees/CategoriesTree.tsx
+++ b/packages/tree-widget/src/components/trees/CategoriesTree.tsx
@@ -1,19 +1,24 @@
 /*---------------------------------------------------------------------------------------------
-* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
-*--------------------------------------------------------------------------------------------*/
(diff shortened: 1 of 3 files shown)


packages/tree-widget/src/components/trees/CategoriesTree.tsx
```diff
@@ -1,19 +1,24 @@
...
+  CategoryTree,
+  toggleAllCategories,
+  getCategories,
+} from "@bentley/ui-framework";
 import { IconButton } from "../IconButton";
 import { SearchBar } from "../search-bar/SearchBar";
 import { useTreeFilteringState } from "../TreeFilteringState";
 import "./CategoriesTree.scss";
 import { TreeWidget } from "../../TreeWidget";
+import { CategoryVisibilityHandler } from "@bentley/ui-framework/lib/ui-framework/imodel-components/category-tree/CategoryVisibilityHandler";
```
aruniverse, line 21: Could you import `CategoryVisibilityHandler` from the root level barrel file in ui-framework, so just add it to the existing imports from ui-framework

packages/tree-widget/src/components/trees/CategoriesTree.tsx
```diff
@@ -53,6 +58,40 @@ export function CategoriesTreeComponent(props: CategoriesTreeComponentProps) {
     );
   }, [props.iModel, filteredProvider]);
 
+  const invert = React.useCallback(async () => {
+    const activeView = IModelApp.viewManager.getFirstOpenView();
+    const ids = await getCategories(props.iModel, activeView, filteredProvider);
+    if (!activeView) {
```
aruniverse, line 64: wouldn't you want to check if `activeView` is defined prior to calling `getCategories`?

diegopinate: That call takes an undefined viewport, but yes, in this case it makes sense to do it after. I'll fix it. Thx

packages/tree-widget/src/components/trees/CategoriesTree.tsx
diegopinate: Nope, is internal atm, would need to update core first.

packages/tree-widget/src/components/trees/CategoriesTree.tsx
diegopinate: Fixed.

packages/tree-widget/src/components/trees/CategoriesTree.tsx
aruniverse: damn, thats unfortunate.

PR #27, opened 2020-07-30
Decouple tree header

diegopinate:
Tree Header component for other widgets to use


packages/tree-widget/src/components/header/TreeHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface TreeHeaderComponentProps {
+  showAll?: () => void;
+  hideAll?: () => void;
+  invert?: () => void;
+  searchOptions: SearchOptions;
+}
+
+export function TreeHeaderComponent(props: TreeHeaderComponentProps) {
+  const { searchOptions } = props;
```
aruniverse, line 16: nit pick, you can destructure all the props, and might as well move it above like so 

```tsx
export function TreeHeaderComponent({ searchOptions, showAll, hideAll, searchOptions }: TreeHeaderComponentProps) {
```

aruniverse: Also, should this be a patch or a new minor bump? Since you're exporting a new component maybe its a minor bump? I'm not too bothered with how it is tho.

diegopinate: Not a fan of that myself, but sure. :)

packages/tree-widget/src/components/header/TreeHeader.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from "react";
```
calebmshafer, line 1: copyright header

aruniverse: i should look into adding the linter to the monorepo

calebmshafer: Yeah, I've been working on trying to adapt what kevin has in the itwin-viewer to be a little bit more broad for use in a lot of our repos.  I can send you the PR I have in progress for iModel.js

diegopinate: Forgot :) fixed

PR #29, opened 2020-08-07
Added 2D & 3D Visibility Toggles

cortneysmithbentley:
Added icon buttons to tree header for 2D and 3D visibility toggles. This included some logic added to the ModelTree as well as adding a label property to the icon button (to display "2D" and "3D" next to the new buttons).

### DIFF
--- a/packages/tree-widget/src/components/header/TreeHeader.tsx
+++ b/packages/tree-widget/src/components/header/TreeHeader.tsx
@@ -74,7 +74,7 @@ export function TreeHeaderComponent({
         )}
         {toggle2D && toggle2DIcon && (
           <IconButton
-            className={"tree-widget-models-tree-toolbar-icon"}
+            className={"tree-widget-header-tree-toolbar-icon"}
             key="view-2d-btn"
             icon={toggle2DIcon}
             title={TreeWidget.translate("toggle2DViews")}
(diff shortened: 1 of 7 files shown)


packages/tree-widget/src/components/header/TreeHeader.tsx
```diff
@@ -13,50 +13,86 @@ export interface TreeHeaderComponentProps {
...
-        </div>
-      </SearchBar>
-  )
-}
\ No newline at end of file
+          />
+        )}
+        {toggle2D && toggle2DIcon && (
+          <IconButton
+            className={"tree-widget-models-tree-toolbar-icon"}
```
aruniverse, line 77: will the 2d, and 3d toggle only be available for the models tree? right now this TreeHeader component will be used by all the trees. So using `models` in the classname here isnt great imo

packages/tree-widget/src/components/header/TreeHeader.tsx
```diff
@@ -13,50 +13,86 @@ export interface TreeHeaderComponentProps {
...
+            key="view-2d-btn"
+            icon={toggle2DIcon}
+            title={TreeWidget.translate("toggle2DViews")}
+            onClick={toggle2D}
+            label={TreeWidget.translate("label2D")}
+          />
+        )}
+        {toggle3D && toggle3DIcon && (
+          <IconButton
+            className={"tree-widget-models-tree-toolbar-icon"}
```
aruniverse, line 87: same comment as above

PR #30, opened 2020-08-19
Property grid enhancements

diegopinate:
1. MultiElementPropertyGrid class that lets user inspect the selection set, and inspect properties of each elements
2. First pass at making the property grid an extension (needs work)


packages/property-grid/src/PropertyGridManager.ts
```diff
@@ -71,3 +80,32 @@ export class PropertyGridManager {
     return category;
   }
 }
+
+/** Extension object for loading on runtime */
+class PropertyGridExtension extends Extension {
```
aruniverse, line 85: Should the extension work be extracted into its own file?

packages/property-grid/src/components/ElementList.scss
```diff
@@ -0,0 +1,57 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+@import "~@bentley/ui-core/lib/ui-core/style/index";
+@import "~@bentley/ui-core/lib/ui-core/text/index";
+@import "@bentley/ui-core/lib/ui-core/index";
```
aruniverse, line 7: missing the `~` in import for line 7. If you import the barrel on line 7, i dont think you need the ones on 5 and 6

packages/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -78,25 +86,25 @@ interface PropertyGridState {
...
 
-    this._dataProvider = new PropertyDataProvider(props.iModelConnection, props.rulesetId, props.enableFavoriteProperties);
-
-    if (this.props.accessToken) {
-      this._requestContext = new AuthorizedClientRequestContext(
-        this.props.accessToken,
-      );
-    }
+    this._dataProvider =
+      props.dataProvider !== undefined
```
aruniverse, line 101: You could use `??` here instead of the ternary op

packages/property-grid/src/components/PropertyGridUiItemsProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  UiItemsProvider,
+} from "@bentley/ui-abstract";
+
+import { PropertyGrid, PropertyGridProps } from "./PropertyGrid";
+import React = require("react");
+import { connect } from "react-redux";
+import { FrameworkState } from "@bentley/ui-framework";
+
+// Map framework state to property grid props
+const mapStateToProps = (state: FrameworkState): PropertyGridProps => {
```
aruniverse, line 15: anyway to make sure that the consumer has setup the UiFramework redux store? Should probably have default values or throw an error if not

packages/property-grid/package.json
```diff
@@ -92,6 +95,7 @@
     "classnames": "^2.2.5",
     "i18next": "^10.2.2",
     "react-redux": "^7.2.0",
```
aruniverse, line 97: forgot to mention earlier, does this need to be a dev+peer dep now?

jason-crow: react-redux does need to be a peerDep

diegopinate: yep, changing in a PR, fixing all other comments too.

diegopinate: fixed.

packages/property-grid/src/components/PropertyGridUiItemsProvider.tsx
diegopinate: Added check but not in the mapStateToProps. The check is in the PropertyGridManager.ts initialize function. StateManager class can be checked to ensure UiFramework store is initialized properly.

packages/property-grid/src/components/ElementList.scss
diegopinate: Fixed.

packages/property-grid/src/PropertyGridManager.ts
diegopinate: Moved.

packages/property-grid/src/components/PropertyGridUiItemsProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+  AbstractWidgetProps,
+  StagePanelLocation,
+  StagePanelSection,
+  UiItemsProvider,
+} from "@bentley/ui-abstract";
+import { FrameworkState } from "@bentley/ui-framework";
+
+import { PropertyGrid, PropertyGridProps } from "./PropertyGrid";
+
+import React = require("react");
```
aruniverse, line 14: any reason this is being imported with a require?

packages/property-grid/src/components/ElementList.tsx
```diff
@@ -0,0 +1,222 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * Licensed under the MIT License. See LICENSE.md in the project root for license terms.
+ *--------------------------------------------------------------------------------------------*/
+
+import React = require("react");
```
aruniverse, line 6: same here re using the require

packages/property-grid/package.json
```diff
@@ -86,12 +92,13 @@
     "cpx": "^1.5.0",
     "rimraf": "^3.0.2",
     "tslint": "^5.11.0",
-    "typescript": "~3.7.4"
+    "typescript": "~3.7.4",
+    "react-redux": "^7.2.0",
+    "react-spring": "^8.0.27",
```
aruniverse, line 97: i actually think you can leave react-spring as a direct dep

packages/property-grid/src/components/PropertyGridUiItemsProvider.tsx
diegopinate: No reason, it was done this way by the "add auto import". I'll fix.

packages/property-grid/package.json
diegopinate: Oh yeah, just checked its deps, looks like it won't be an issue. I'll change it to a direct dep.

packages/property-grid/src/components/PropertyGridUiItemsProvider.tsx
diegopinate: fixed

packages/property-grid/package.json
```diff
@@ -29,55 +31,58 @@
...
+    "@bentley/ui-components": "^2.3.3",
+    "@bentley/ui-core": "^2.3.3",
+    "@bentley/ui-framework": "^2.3.3",
+    "@bentley/ui-ninezone": "^2.3.3",
+    "@bentley/webgl-compatibility": "^2.3.3",
     "react": "^16.8.0",
-    "react-dom": "^16.8.0"
+    "react-dom": "^16.8.0",
+    "react-redux": "^7.2.0",
+    "redux": "^4.0.3"
```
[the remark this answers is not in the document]
diegopinate: cool. fixed. thx for checking!

PR #34, opened 2020-08-24
Applying package updates.

jason-crow:


packages/imodel-select/package.json
```diff
@@ -2,34 +2,43 @@
...
-    "cover": ""
-  },
   "keywords": [
     "Bentley",
     "iModel Select React",
     "iModel.js"
   ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/imodeljs/viewer-components-react/tree/imodel-select/packages/imodel-select"
```
archana-maharjan, line 12: Repo Url doesn't exist, I believe it should be https://github.com/imodeljs/viewer-components-react/tree/master/packages/imodel-select

jason-crow: nice catch!

PR #35, opened 2020-08-24
add copyright linter

aruniverse:
Add copyright linter


common/scripts/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@bentley/viewer-components-react-scripts",
```
calebmshafer, line 2: So I don't think you need a package.json.. I was testing this out in the iModel.js repo and pretty sure I figured out a way to get around it.  At least for the copyright header case.

calebmshafer: Let me dig up that branch, I didn't get around to moving forward with it yet but would like to soon.

aruniverse: If we can get around that, thatd be great. I just copied what the viewer was doing. Just need to make sure lint-staged is installed.

aruniverse: Updated, removed the unnecessary pkg, and tested it out locally. This should be added to all the other imjs repos as well

PR #40, opened 2020-08-24
modify the build-publish.yaml to separate rush version and rush publi…

jason-crow:
…sh steps

### DIFF
--- a/common/config/azure-pipelines/templates/build-publish.yaml
+++ b/common/config/azure-pipelines/templates/build-publish.yaml
@@ -2,8 +2,10 @@ steps:
   - checkout: self
     persistCredentials: true
   - template: ./build.yaml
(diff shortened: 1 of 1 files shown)


aruniverse:
@jason-crow what was the need for this? its been a few months, so i dont expect you to remember. but i might close this if it deems unnecessary

calebmshafer:
@aruniverse I think this was due to the issue with package.json ordering.

aruniverse:
> @aruniverse I think this was due to the issue with package.json ordering.

what do you think, still needed?

calebmshafer:
@aruniverse I think we're fine skipping this for now. We haven't hit the issue in a while :)

jason-crow:
agreed, yeah I couldn't even recall the basis of this

PR #41, opened 2020-08-24
update .gitattributes to converge with microsoft/rushstack's

jason-crow:


calebmshafer:
@jason-crow did no files change when you did this?

aruniverse:
@calebmshafer  @jason-crow Same story as #40 ? Cool if i close?

calebmshafer:
@aruniverse fine by me.

PR #43, opened 2020-08-25
Create codeql.yml

calebmshafer:
The iModel.js org has access to the Code Scanning beta offered by GitHub.  Setup a basic scan that will run on every PR and on a weekly basis.

This is very similar, if not the same, as what will be found on LGTM, https://lgtm.com/projects/g/imodeljs/viewer-components-react/?mode=list. 

Learn more about how to config it and monitor the results [here](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,54 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 15 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,54 @@
...
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      # Override language selection by uncommenting this and choosing your languages
+      # with:
+      #   languages: go, javascript, csharp, python, cpp, java
```
aruniverse, line 35: this monorepo will only have js/ts files so maybe we should explicitly select which languages to use.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
```
aruniverse, line 39: looks like we can remove the autobuild step

.github/workflows/codeql.yml
calebmshafer: Yea, I went back and forth on cleaning up some of these defaults. I wanted to see how this went before going through and cleaning it up but you're right.

.github/workflows/codeql.yml
calebmshafer: good point, I'll remove it

PR #44, opened 2020-08-25
imodel-content-tree: Initial commit

grigasp:
Initial commit of the new "everything tree". 

Some things that I would like improved:
- [ ] I find it very hard to explain the resulting hierarchy (in README.md) in a way that's both understandable and detailed enough. Maybe adding an image of a real hierarchy would be more helpful, but need to find a good dataset for that.
- [ ] I'd like to add at least one integration test that produces the hierarchy for a real iModel. The iModel should be small enough to be included in the repository, but also have a complex-enough structure (subjects, models, categories) to exercise all the rules. And of course it has to be public to be included here.


grigasp:
@calebmshafer, the linter fails due to the issue i emailed you about

calebmshafer:
> @calebmshafer, the linter fails due to the issue i emailed you about

Yes, the release of 2.5.1 that should happen shortly which will contain the fix for the custom eslint-rules not being in the build-tools package.

calebmshafer:
>   I'd like to add at least one integration test that produces the hierarchy for a real iModel. The iModel should be small enough to be included in the repository, but also have a complex-enough structure (subjects, models, categories) to exercise all the rules. And of course it has to be public to be included here.

@grigasp with regards to the sample data, is that something you could easily create in a test iModel or do we need a seed iModel?

calebmshafer:
> > @calebmshafer, the linter fails due to the issue i emailed you about
> 
> Yes, the release of 2.5.1 that should happen shortly which will contain the fix for the custom eslint-rules not being in the build-tools package.

@grigasp the 2.5.1 packages are available.  Can you test them out and see if that fixes the issue?

packages/imodel-content-tree/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@bentley/imodel-content-tree",
```
aruniverse, line 2: @calebmshafer , should this pkg also use the `-react` suffix we've been using for the others?

calebmshafer: Yes, good catch.  @grigasp can you add that to the name of the package?  It's a standard we've been enforcing within this repo.

aruniverse: You'll have to update the CODEOWNERS, version-policies.json, rush.json, and readme. Maybe a few more spots too

common/config/rush/version-policies.json
```diff
@@ -18,5 +18,10 @@
     "policyName": "property-grid",
     "definitionName": "individualVersion",
     "lockedMajor": 1
+  },
+  {
+    "policyName": "imodel-content-tree",
+    "definitionName": "individualVersion",
+    "lockedMajor": 0
```
aruniverse, line 25: nbd, but we're going to version this as 0.x? will this be a beta pkg?

packages/imodel-content-tree/.npmignore
```diff
@@ -0,0 +1,8 @@
+# start off ignoring everything, and then add back only the files we want
+*
+!lib/**/*.d.ts
```
aruniverse, line 3: could you add an entry for `!lib/**/*.d.ts.map`?

packages/imodel-content-tree/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "copy:assets": "cpx \"./src/**/*.scss\" ./lib && cpx \"./public/**/*\" ./lib/public",
+    "cover": "",
+    "lint": "eslint ./src/**/*.{ts,tsx} 1>&2",
+    "test": ""
+  },
+  "dependencies": {
+    "classnames": "^2.2.5"
+  },
+  "devDependencies": {
+    "@bentley/bentleyjs-core": "^2.5.1",
```
aruniverse, line 34: I could be mistaken, but these dev deps should probably be pinned down

common/config/rush/version-policies.json
grigasp: We're still not 100% certain how the hierarchy should look like. @Carl-Hinkle said he'd like DR to consume this and see the results and go from there. So yes, this is still beta - we're likely to change the hierarchy that the component shows.

packages/imodel-content-tree/package.json
calebmshafer: I don't know if this is necessary, I actually meant to talk to everyone else about this. I think we could move to using `^` everywhere since that's what our peerDeps state as well.

calebmshafer: Resolving this. @aruniverse we can chat separately about what to do here.

grigasp:
@calebmshafer any idea what the failures are about?

aruniverse:
> @calebmshafer any idea what the failures are about?

It's not coming from your pr which is strange. 
https://dev.azure.com/imodeljs/imodeljs/_build/results?buildId=1828&view=logs&j=572c5e49-83d5-5271-390a-e6dc77f89c6b&t=c696d233-fcbb-5de2-6396-4122a0838c2b

image

grigasp:
> > @calebmshafer any idea what the failures are about?
> 
> It's not coming from your pr which is strange.
> https://dev.azure.com/imodeljs/imodeljs/_build/results?buildId=1828&view=logs&j=572c5e49-83d5-5271-390a-e6dc77f89c6b&t=c696d233-fcbb-5de2-6396-4122a0838c2b
> 
> image

It failed because my `rush update` pulled the latest version of pretty-format which requires typescript to be at least 3.8. 

First I tried to lower the version of dependency that requires `pretty-format`, but found out that the chain is pretty long (`markup-frontstage-react` -> `@testing-library/react` -> `@testing-library/dom` -> `pretty-format`) and `@testing-library/dom` uses `^` for `pretty-format`, so I would've had to lower the version to a very old one. 

So instead I upgraded the typescript version to `~3.8`. This seems to have revealed an issue in `markup-frontstage-react` where it was setting non-existing attribute for `ColorPickerButton` (`activeColor` vs `initialColor`).

aruniverse:
> > > @calebmshafer any idea what the failures are about?
> > 
> > 
> > It's not coming from your pr which is strange.
> > https://dev.azure.com/imodeljs/imodeljs/_build/results?buildId=1828&view=logs&j=572c5e49-83d5-5271-390a-e6dc77f89c6b&t=c696d233-fcbb-5de2-6396-4122a0838c2b
> > image
> 
> It failed because my `rush update` pulled the latest version of pretty-format which requires typescript to be at least 3.8.
> 
> First I tried to lower the version of dependency that requires `pretty-format`, but found out that the chain is pretty long (`markup-frontstage-react` -> `@testing-library/react` -> `@testing-library/dom` -> `pretty-format`) and `@testing-library/dom` uses `^` for `pretty-format`, so I would've had to lower the version to a very old one.
> 
> So instead I upgraded the typescript version to `~3.8`. But there seem to have been other dependency updates that caused `markup-frontstage-react` tests to fail, so this became a never-ending story.
> 
> I dropped all the merge commits, rebased my changes on master and made sure the shrinkwrap is updated only with dependencies that are needed for my new package while keeping dependencies of other packages as-is. That seems to have solved the issues.
> 
> Someone definitely needs to re-generate the shrinkwrap so it pulls the latest dependencies and then fix all the issues, otherwise it'll be a pain to make any updates here, since at this moment we're using conflicting dependencies (pretty-format requires TS 3.8, we're using 3.7)

Thanks Grigas, will look into it. Might be due to me setting ts ~3.7.4 as the common version to use across the monorepo

PR #48, opened 2020-08-28
Feat/upgrade ui component in markup

archana-maharjan:
This PR consists of fix for breaking changes caused by "@bentley/ui-components" v2.5.0. The consumer that are using imodelJs packages of v2.5.0 or above were seeing issues on ColorPickerButton due to the change in property name from "activeColor" to "initialColor". Hence, version upgrade to most of the "@bentley" peer dependencies required upgrade.

After bumping the version, ran into pretty-format issue, bumping the typescript version solved it.

Finally, unit test broke after all of these upgrades, so removed some of the unit test from MarkupSettingPanel where it requires mock of both "ui-components" as well as "ui-framework". I've added a not to cover those removed unit test from end2end test at my side.


common/config/rush/pnpm-lock.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   '@rush-temp/tree-widget-react': 'file:projects/tree-widget-react.tgz'
```
aruniverse, line 12: @calebmshafer , we need to remove you as a codeowner to the the lockfile

calebmshafer: Yea feel free to open a PR today so I can be removed before going on vacation next week.  Also, should probably find someone, in addition to me, who can approve all any change to the repo.

PR #49, opened 2020-08-28
add vcr team as default codeowners. mainly so approvals for lockfiles…

aruniverse:
add imjs/vcr team as default codeowner for the monorepo. mainly so approvals for lockfile arent dependent on only one person


aruniverse:
fyi @jason-crow , @wgoehrig . You guys should be able to approve any pr

calebmshafer:
I think this still requires me because of what's on `master`.  We can check once this is approved.

PR #51, opened 2020-09-29
Enabled property group nesting in property-grid package

cortneysmithbentley:
PBI: https://dev.azure.com/bentleycs/beconnect/_workitems/edit/444702/

Enabled property group nesting, unless the data provider is passed in through props (in which case we don't want to overwrite what is passed in).


aruniverse (commented):
Missing rush change

aruniverse (approved):

diegopinate (approved):

PR #55, opened 2020-10-09
Feat/support view2d

archana-maharjan:
This PR contains view2D support feature in addition to Markup Frontstage React package. The SavedViewData provided by the MakrupFrontstage used to only support SpatielView, now with this new addition, it also supports SheetView and DrawingView. In order to not break existing consumers, I've added 2D view properties in SavedViewData, instead of creating BaseSavedViewData(with both common 3Dview and 2Dview), (3DSavedViewData, 2DSavedViewData)inheriting from BaseSavedViewData.

### DIFF
--- a/packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
+++ b/packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
@@ -3,6 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
+
 import {
   EmphasizeElements,
   EmphasizeElementsProps,
@@ -220,7 +221,10 @@ export class MarkupFrontstageProvider extends FrontstageProvider {
   private _svgRect = (svg: string | SVGSVGElement) => {
     if (typeof svg === "string") {
       const dom = new DOMParser().parseFromString(svg, "image/svg+xml");
-      if (dom.getElementsByTagName("parsererror").length > 0) {
+      if (
+        dom.getElementsByTagName("parsererror") &&
+        dom.getElementsByTagName("parsererror").length > 0
+      ) {
         throw new UiError(
           MarkupFrontstage.loggerCategory(MarkupFrontstageProvider),
           `MarkupData.svg is invalid: ${
--- a/packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
+++ b/packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
@@ -2,6 +2,7 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
+
 import {
   EmphasizeElements,
   EmphasizeElementsProps,
@@ -34,7 +35,7 @@ import React from "react";
 import { MarkupFrontstage } from "../../MarkupFrontstage";
 import MarkupSettingsPanel from "../toolbar/MarkupSettingsPanel";
 import { MarkupToolWidget } from "../toolbar/MarkupToolWidget";
-import { createSavedViewData } from "../../services/SavedView";
+import { createMarkupSavedViewData } from "../../services/SavedView";
 import {
   AddMarkupEvent,
   MarkupFrontstageConstants,
@@ -179,14 +180,17 @@ export class MarkupFrontstageProvider extends FrontstageProvider {
       MarkupFrontstageConstants.VIEW_LAYOUT_ID
     );
     const currentViewPort = contentControl?.viewport;
-    const savedView = currentViewPort && createSavedViewData(currentViewPort);
+    const savedView =
+      currentViewPort && createMarkupSavedViewData(currentViewPort);
     const markupData = await MarkupApp.stop();
     try {
       if (savedView) {
         savedView.markup = markupData.svg;
-        savedView.emphasizedElementsProps = JSON.stringify(
-          this._emphasizedElementsProps
-        );
+        savedView.emphasizedElementsProps = savedView.is2d
+          ? savedView.emphasizedElementsProps
+          : this._emphasizedElementsProps
+          ? JSON.stringify(this._emphasizedElementsProps)
+          : undefined;
         this.onAddMarkupEvent.raiseEvent({
           savedView,
           thumbImage: markupData.image,
@@ -220,7 +224,7 @@ export class MarkupFrontstageProvider extends FrontstageProvider {
         throw new UiError(
           MarkupFrontstage.loggerCategory(MarkupFrontstageProvider),
           `MarkupData.svg is invalid: ${
-          dom.getElementsByTagName("parsererror")[0].textContent
+            dom.getElementsByTagName("parsererror")[0].textContent
           }`
         );
       } else {
@@ -238,9 +242,9 @@ export class MarkupFrontstageProvider extends FrontstageProvider {
     if (view) {
       const markupData = this._svg
         ? {
-          rect: this._svgRect(this._svg),
-          svg: this._svg,
-        }
+            rect: this._svgRect(this._svg),
+            svg: this._svg,
+          }
         : undefined;
       if (this._emphasizedElementsProps) {
         this._createEmphasizedElements(this._emphasizedElementsProps, view);
--- a/packages/markup-frontstage/src/services/SavedView.ts
+++ b/packages/markup-frontstage/src/services/SavedView.ts
@@ -3,6 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
+
 import {
   Code,
   ColorDef,
@@ -19,6 +20,7 @@ import {
   SheetViewState,
   SpatialViewState,
   Viewport,
+  ViewState2d,
 } from "@bentley/imodeljs-frontend";
 
 import {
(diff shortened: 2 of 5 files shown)


packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
```diff
@@ -220,7 +224,7 @@ export class MarkupFrontstageProvider extends FrontstageProvider {
         throw new UiError(
           MarkupFrontstage.loggerCategory(MarkupFrontstageProvider),
           `MarkupData.svg is invalid: ${
-          dom.getElementsByTagName("parsererror")[0].textContent
+            dom.getElementsByTagName("parsererror")[0].textContent
```
kckst8, line 227: should we check if this element exists and has length before accessing textContent?

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -71,9 +80,72 @@ export const parseFromEmphasizeElements = (
...
+ * Creates a drawing saved view.
+ * @param vp the name of the view port.
+ */
+const createDrawingSavedView = (vp: Viewport): SavedViewData => {
+  const viewState = vp.view as DrawingViewState;
+  if (!viewState) {
+    throw new Error("Invalid viewport");
+  }
+
+  const categorySelectorProps = viewState.categorySelector.toJSON();
```
kckst8, line 92: maybe ensure that categorySelector and displayStyle are defined?

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -71,9 +80,72 @@ export const parseFromEmphasizeElements = (
...
+ * Creates a sheet saved view
+ * @param vp the name of the view port.
+ */
+const createSheetSavedView = (vp: Viewport): SavedViewData => {
+  const viewState = vp.view as SheetViewState;
+  if (!viewState) {
+    throw new Error("Invalid viewport");
+  }
+  const categorySelectorProps = viewState.categorySelector.toJSON();
+  const viewDefinitionProps = viewState.toJSON();
```
kckst8, line 119: same here

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -71,9 +80,72 @@ export const parseFromEmphasizeElements = (
 };
 
 /**
- * Creates saved view data
+ * Creates a drawing saved view.
+ * @param vp the name of the view port.
+ */
+const createDrawingSavedView = (vp: Viewport): SavedViewData => {
```
kckst8, line 86: this and createSheet... functions are very similar. Could they be combined and parameterized? Even if you leave the separate functions but they both call the single function with different parameters

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
 
   const data: SavedViewData = {
     alwaysDrawn,
-    neverDrawn,
     categories,
+    changeSetId,
+    displayStyleProps: JSON.stringify(viewState.displayStyle.toJSON()),
```
kckst8, line 224: ensure defined

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
 
   return data;
 };
+
+/**
+ * Determines whether or not given saved view is spatial.
+ * @param view name of the Saved View.
+ */
+export const isSpatialSavedView = (view: SavedViewData) => {
+  return (view.is2d === undefined || !view.is2d) && "models" in view;
```
kckst8, line 248: shouldn't !view.is2d suffice here? That should cover undefined and false

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
+};
+
+/**
+ * Determines whether or not given view is Drawing.
+ * @param view name fo the Saved View.
+ */
+export const isDrawingSavedView = (view: SavedViewData) => {
+  return view.is2d !== undefined && view.is2d && !("sheetProps" in view);
```
kckst8, line 256: see comment above. I think view.is2d would meet both conditions here

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
+};
+
+/**
+ * Determines whether or not given view is Sheet.
+ * @param view name of the Saved View.
+ */
+export const isSheetSavedView = (view: SavedViewData) => {
+  return view.is2d !== undefined && view.is2d && "sheetProps" in view;
```
kckst8, line 264: same

JoeZman (approved):

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
+};
+
+/**
+ * Creates a drawing view state from the data object.
+ * @param iModelConnection the name of the IModel Connection.
+ * @param savedView the naem of the Saved View.
+ */
+export const createDrawingViewState = async (
```
kckst8, line 272: Similar comment as above. Could these be combined into a single parameterized function?

kckst8 (changes requested):

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +219,125 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
+  await viewState.load();
+  return viewState;
+};
+
+/**
+ * Creates a markup saved view data from the viewport, it could return a
+ * View2d(Sheet or Drawings) or a View3d(Spatial views) object
+ * @param vp the name of the view port.
+ */
+export const createMarkupSavedViewData = (vp: Viewport): SavedViewData => {
```
diegopinate, line 335: Why do you need this special markup one at all? Isn't it the same as createSavedView?

archana-maharjan: CreateSavedViewData returns spatial viewState, I was thinking of using createSavedViewData to return (spatial, sheet and drawing) it'll break consumer's since that function is accessible. Hence I named createMarkupSavedViewData instead of CreateSavedViewData, please let me know if you have any good name suggestion?

packages/markup-frontstage/src/components/ui/MarkupFrontstageProvider.tsx
archana-maharjan: element exist check added please!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: ensured please!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: thanks, added null optional chaining!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: pulled the common into separate function, and parameterized the exposed function, used exposed function although it was not necessary for drawings, please let me know if you're good with it?

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: thanks, done please!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: added !!view.is2d to cover undefined as well.

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: done please

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: combined both function to parameterize create2dViewState!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: done please!

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +220,97 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
 
   const data: SavedViewData = {
     alwaysDrawn,
-    neverDrawn,
     categories,
+    changeSetId,
+    displayStyleProps: jsonStringify(viewState.displayStyle?.toJSON()),
```
kckst8, line 225: this will still break if viewState is undefined since its properties being accessed before the new function. Maybe just return at the beginning if viewState is undefined?

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -147,22 +220,97 @@ export const createSavedViewData = (vp: Viewport): SavedViewData => {
...
   return data;
 };
+
+/**
+ * Determines whether or not given saved view is spatial.
+ * @param view name of the Saved View.
+ */
+export const isSpatialSavedView = (view: SavedViewData) => {
+  // view.is2d can be undefined for spatialSavedView that are added before 2D saved view feature is added.
+  return !!view.is2d && "models" in view;
```
kckst8, line 250: did you mean to add the double negation here?

packages/markup-frontstage/src/services/SavedView.ts
```diff
@@ -71,9 +82,70 @@ export const parseFromEmphasizeElements = (
...
+    const viewState = vp.view as SheetViewState;
+    const sheetSize = viewState.sheetSize;
+    const sheetProps: SheetProps = {
+      width: sheetSize.x,
+      height: sheetSize.y,
+      model: viewState.model,
+      classFullName: SheetViewState.classFullName,
+      code: Code.createEmpty(),
+    };
+    const common2dViewProps = create2dViewProps(vp);
```
kckst8, line 138: Do we need a new const here? Couldn't we just use the common2dViewProps that was already declared?

kckst8 (changes requested):

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: yes sir!

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: good catch, removed the line 👍

packages/markup-frontstage/src/services/SavedView.ts
archana-maharjan: viewState is checked at line 151 for undefined, I believe it should not break at the time displayStyle is being accessed, but I can add optional null chaining before it's property being accessed!

archana-maharjan: done please!

PR #56, opened 2020-10-13
[project] imodel-react-hooks

MichaelBelousov:
extracted from the it1 monorepo with some changes. Currently used in at least two known applications, which will need to migrate to the new name and public location.


common/config/rush/version-policies.json
```diff
@@ -1,27 +1,32 @@
-[
```
MichaelBelousov, line 1: it looks like probably the line endings changed on this file, I can rebase and undo the change if needed

MichaelBelousov:
and I just realized I haven't done the changelog stuff. Sorry about that 😞 I'll fix it

PR #65, opened 2020-10-29
Default to iModelApp i18n before throwing an error

kckst8:
This will allow the component to be used in use cases where IModelApp has already been initialized and an additional async initialize call is not optimal

### DIFF
--- a/packages/imodel-select/src/IModelSelect.ts
+++ b/packages/imodel-select/src/IModelSelect.ts
@@ -2,11 +2,13 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
+
 /** @module Common */
 
 import * as i18next from "i18next";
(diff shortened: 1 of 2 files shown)


packages/imodel-select/src/IModelSelect.ts
```diff
@@ -21,7 +23,8 @@ export class IModelSelect {
    */
   public static async initialize(i18n: I18N): Promise<void> {
     IModelSelect._i18n = i18n;
-    return IModelSelect._i18n.registerNamespace(IModelSelect.i18nNamespace).readFinished;
+    return IModelSelect._i18n.registerNamespace(IModelSelect.i18nNamespace)
```
calebmshafer, line 26: Wouldn't we occasionally run into the issue where this doesn't finish before we try and get the `i18n` instance which would cause the `IModelApp.i18n` to override whatever this call is trying to do?

kckst8: _i18n would still be defined prior to this though? It's defined in ln25. This code is just registering a new namespace I thought?

PR #66, opened 2020-10-30
Add a codeowners file sanity check workflow

calebmshafer:
The current codeowners file isn't working correctly due to a mis-configuration.  This workflow has been helpful in finding issues with the one in iModel.js.

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -1,0 +1,24 @@
+name: "Codeowners Validator"
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+    - '.github/CODEOWNERS'
+    - '.github/workflows/sanity.yaml'
+  schedule:
+    # Runs at 08:00 UTC every day
+    - cron:  '0 8 * * *'
+
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository, which is validated in the next step
+      - uses: actions/checkout@v2
+      - name: GitHub CODEOWNERS Validator
+        uses: mszostok/codeowners-validator@v0.5.0
+        with:
+          checks: "files,duppatterns,syntax"
+          # GitHub access token is required only if the `owners` check is enabled
+          github_access_token: "${{ secrets.OWNERS_VALIDATOR_GITHUB_SECRET }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/sanity.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository, which is validated in the next step
+      - uses: actions/checkout@v2
+      - name: GitHub CODEOWNERS Validator
+        uses: mszostok/codeowners-validator@v0.5.0
+        with:
+          checks: "files,duppatterns,syntax"
```
aruniverse, line 23: id be interested in seeing `notowned` added to the list here

aruniverse: granted, its experimental.  
and to enable experimental check set EXPERIMENTAL_CHECKS=notowned environment variable.

aruniverse (approved):

aruniverse:
files from rush change will be annoying 
image

.github/workflows/sanity.yml
calebmshafer: Yea, I had the `owners` enabled in iModel.js for a little while but it was failing due to a few people not being in the Org. I think we may hit the same issue here since a few of the people in the file aren't technically in the Org.

As for `notowned`, I could add that but I figured it's pretty much covered by `*` but if we get a more complicated config then it may make sense.

calebmshafer:
> files from rush change will be annoying
> image

Yes, it will be.  I haven't yet figured out how to handle this in the iModel.js repo either. What I'd really like to do is not require any reviewers for those files and have the reviewers at the actual package level review them.

calebmshafer:
@aruniverse just updated the way we handle the common/changes to avoid invalidating the codeowners file.

aruniverse:
> @aruniverse just updated the way we handle the common/changes to avoid invalidating the codeowners file. Unfortunately it doesn't look like you can ignore certain files so instead I just set everyone who is a reviewer for a package as a reviewer for the entire folder.
> 
> Let me know what you think.

Yea this should work, list might eventually get long and at that point I guess we can make another team like `@imodeljs/viewer-components-owners` or something. We already have the other team vc-reviewers team we could update, but lets cross that bridge once we get there?

calebmshafer:
@aruniverse sounds good.  I was thinking that's what I may do for iModel.js since we have a couple dozen reviewers overall so once I get a good setup there I'll copy it over.

The main thing I want to avoid when doing that is notifying everyone in the list that they need to review that specific change.

PR #69, opened 2020-11-20
Fix ordering of pkg deps

calebmshafer:
Update the ordering of dependencies in the package.json to avoid the Rush failure.


aruniverse (approved):
Maybe, also update the codeowners file for pkgs with only one owner to fallback to the vcr team? I thought thats what it did, but guess not. This is just for a fail safe when we run into rush issues with the publishing like this again

calebmshafer:
> Maybe, also update the codeowners file for pkgs with only one owner to fallback to the vcr team? I thought thats what it did, but guess not. This is just for a fail safe when we run into rush issues with the publishing like this again

@aruniverse sounds good to me.  Just updated it for all of the packages that only had one.

The CODEOWNERS doesn't propagate up the hierarchy of reviewers but is instead explicit and only goes down the folder structure.  Unless there is an explicit reviewer for any given sub-folder or file.

mdastous-bentley (approved):

PR #70, opened 2020-11-23
Consuming fix for property grid labels

cortneysmithbentley:
Consuming fix for property grid labels by updating @bentley/ui-components package and other related packages to 2.9.2 (fix is in 2.6, this is the latest non-dev release. was unsure which version to use)


packages/property-grid/package.json
```diff
@@ -54,7 +54,7 @@
     "@bentley/presentation-frontend": "2.3.3",
     "@bentley/product-settings-client": "2.3.3",
     "@bentley/ui-abstract": "2.3.3",
-    "@bentley/ui-components": "2.3.3",
+    "@bentley/ui-components": "2.9.2",
```
aruniverse, line 57: you cant bump just one core pkg, its all or nothing. so please bump all the dev deps to 2.9.2, and the peer deps to ^2.9.2

aruniverse (changes requested):
Maybe make this a minor bump rather than a patch

packages/property-grid/package.json
cortneysmithbentley: gotcha didn't realize I missed some, will do.

cortneysmithbentley: Ok I definitely did this for the rest of the packages (just forgot the peer deps) and it didn't get committed, so I'm not totally useless. 👍  it's all pushed now

packages/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -28,6 +28,7 @@ import {
   PropertyGrid as CorePropertyGrid,
```
aruniverse, line 28: Is this import even being used anymore? If not, you can remove it. PropertyGrid as CorePropertyGrid

aruniverse (approved):

aruniverse:
you also need to run `rush update` to update the lockfile.

aruniverse (approved):

PR #75, opened 2020-12-09
Geotools/geo search icon

mdastous-bentley:
Cosmetic changes: New button icon and label.


packages/geo-tools/package.json
```diff
@@ -46,6 +47,7 @@
     "@bentley/imodeljs-frontend": "^2.9.1",
     "@bentley/imodeljs-i18n": "^2.9.1",
     "@bentley/itwin-client": "^2.9.1",
+    "@bentley/react-scripts": "3.4.6",
```
aruniverse, line 50: any reason for adding this? i assume you were adding this to pickup the sprite-loader, but looks like that a dependency has also been added

mdastous-bentley: "@bentley/react-scripts" is referenced is tsconfig.json.  Without it, its impossible to import SVG files.

aruniverse: I believe what you really need from this pkg is the following type declaration. 

```
declare module '*.svg?sprite' {
  const src: string;
  export default src;
}
```

I recommend you create a `geotools/declaration.d.ts` file with the above snippet, and remove the dep on react-scripts. Could you try that and let me know if that works?

aruniverse (approved):
Approving for now, will revisit later and confirm what the right way to handle svgs are

aruniverse:
@calebmshafer  failing the ado pipelines with the rush change verify step, which i assume is screwed up now from the failing rush publishes earlier due to the pkg.json ordering

calebmshafer:
> @calebmshafer failing the ado pipelines with the rush change verify step, which i assume is screwed up now from the failing rush publishes earlier due to the pkg.json ordering

@aruniverse yes, there is probably PR that needs to be merged with the necessary changes to the changelogs. I'll take a look in a few minutes.

hnn0003 (approved):

PR #78, opened 2020-12-11
revert back the version policy of markup frontstage

archana-maharjan:
Changed the version policy back to 1, hopefully it'll help package release with new major version.


common/config/rush/version-policies.json
```diff
@@ -7,7 +7,7 @@
   {
     "policyName": "markup-frontstage",
     "definitionName": "individualVersion",
-    "lockedMajor": 2
+    "lockedMajor": 1
```
calebmshafer, line 10: Can you remove the lockedMajor entirely from this policy?  That should let rush change bump the major version.  Not sure if that's the best way to handle this but would at least let you move forward.

calebmshafer (changes requested):

common/config/rush/version-policies.json
archana-maharjan: removed please! after that I was able to generate new correct changelog with rush change with "major" option.

PR #81, opened 2020-12-11
add @imodeljs/viewer-components-reviewers as a failsafe for all pkgs

aruniverse:


.github/CODEOWNERS
```diff
@@ -13,13 +13,13 @@
 # Assigning everyone who owns a package to the common/changes review
 /common/changes @wilmaier @calebmshafer @JoeZman @archana-maharjan @aruniverse @diegopinate @JValiunas @michaalx @grigasp @michaelbelousov
```
calebmshafer, line 14: While we're switching things around, let's fix this. I don't like everyone being added just because there's a changelog.  Can you update it to explicitly set each package? (Gross but the only way I know to do it)

calebmshafer (changes requested):

.github/CODEOWNERS
aruniverse: Just to confirm, you wanted to revert changes made in #66

aruniverse: Hmm idk if this will work. With the major bump for markup-frontstage, it doesnt show up in this [dir](https://github.com/imodeljs/viewer-components-react/tree/master/common/changes/%40bentley) causing the validator to fail

calebmshafer: Yea, I've been running into this when trying to make a change in iModel.js repo.  Not entirely sure how i want to handle it.

Technically it should be a valid codeowners file and it's only failing the validator that I have in place.  We can just remove that check from the validator.

aruniverse: So i removed the `files` check. Added the `owners` check. but thats failing and im guessing the token has expired/needs to get updated. 
image

calebmshafer: We should update it to match the iModel.js one to be, 

```
with:
  checks: "duppatterns,syntax"
  experimental_checks: "notowned"
  # GitHub access token is required only if the `owners` check is enabled
  github_access_token: "${{ secrets.OWNERS_VALIDATOR_GITHUB_SECRET }}"
```

It does not currently include the OWNERS one but that's something I need to fix.

calebmshafer: I'll add whatever token we need to have.

calebmshafer:
@aruniverse lets remove the owners validation for now since everyone isn't in the org yet.

.github/workflows/sanity.yml
```diff
@@ -17,8 +17,9 @@ jobs:
       # Checks-out your repository, which is validated in the next step
       - uses: actions/checkout@v2
       - name: GitHub CODEOWNERS Validator
-        uses: mszostok/codeowners-validator@v0.5.0
+        uses: mszostok/codeowners-validator@v0.6.0
         with:
-          checks: "files,duppatterns,syntax"
+          checks: "duppatterns,owners,syntax"
```
aruniverse, line 22:
```suggestion
          checks: "duppatterns,syntax"
```

calebmshafer (approved):

PR #82, opened 2020-12-14
fix search styling issues

kckst8:
in UI 2.0 with resizable widgets, absolute positioning is causing overlap. Also adjusted transition time:

image


aruniverse (approved):
Tested with ui 1.0 and still looks fine right?

kckst8:
> Tested with ui 1.0 and still looks fine right?

Yep. I changed styles in it1 and it looked fine. Famous last words

PR #86, opened 2020-12-21
Geotools/geo address search label update

mdastous-bentley:


packages/geo-tools/public/locales/en/GeoTools.json
```diff
@@ -1,6 +1,6 @@
 {
   "geoAddressSearch": {
-    "label": "Address Search",
+    "label": "Search Address",
```
calebmshafer, line 3: Let's hold this until I fix the current publishing issues after @aruniverse's fix.  I'll have a PR up in a couple minutes.

calebmshafer (changes requested):

packages/geo-tools/public/locales/en/GeoTools.json
calebmshafer: Once #87 is merged this should be good to go.

calebmshafer (approved):

aruniverse (approved):

PR #88, opened 2020-12-22
lint package.json to prevent ci failures

aruniverse:
Publish fails often due to rush not liking the order of deps in pkg.json or line endings.

See 
- https://github.com/microsoft/rushstack/issues/1795
- https://github.com/microsoft/rushstack/issues/1543
- https://github.com/microsoft/rushstack/issues/1503

### DIFF
--- a/packages/geo-tools/package.json
+++ b/packages/geo-tools/package.json
@@ -32,6 +32,23 @@
     "pseudolocalize": "betools pseudolocalize --englishDir ./public/locales/en --out ./lib/public/locales/en-PSEUDO",
     "test": "mocha --opts ./mocha.opts \"./lib/test/**/*.test.js\""
   },
+  "eslintConfig": {
+    "extends": "./node_modules/@bentley/build-tools/.eslintrc.js"
+  },
+  "nyc": {
+    "branches": 25,
+    "check-coverage": true,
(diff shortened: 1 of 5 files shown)


packages/geo-tools/package.json
```diff
@@ -32,6 +32,23 @@
     "pseudolocalize": "betools pseudolocalize --englishDir ./public/locales/en --out ./lib/public/locales/en-PSEUDO",
     "test": "mocha --opts ./mocha.opts \"./lib/test/**/*.test.js\""
   },
+  "eslintConfig": {
```
calebmshafer, line 35: Is there a way to `sort-package-json` to not put the configs above the dep sections (devDeps, peerDeps, and deps) ?  I always dislike having the configs not start after that, it's hard to ever see the dependencies of a particular package. Plus the configs are something that's mostly only important for the devs of the package not the consume.

I could get over it if there isn't a way though :)

aruniverse: There is a way to specify the order, but I was worried the order of the fields might also cause an issue with the publish step. Waiting to hear back from rush team.

aruniverse:
this hasnt been an issue for a while; and theres been no progress in rush to get this done so closing

PR #89, opened 2021-01-04
Feat/add create view state props

archana-maharjan:
Added SearializeViewStateProps.


archana-maharjan:
Closing this for now, to see if ViewState.toProps() from https://github.com/imodeljs/imodeljs/pull/491 will do the job without above function.

PR #90, opened 2021-01-05
Added support for UK postal code by querying places.

mdastous-bentley:


packages/geo-tools/src/IModelGeoView.ts
```diff
@@ -12,29 +13,31 @@ export class IModelGeoView {
     let result: Range2d | undefined;
 
     const vp = IModelApp.viewManager?.selectedView;
-    if (vp?.iModel?.contextId === undefined || vp?.iModel?.iModelId === undefined)
+    if (vp === undefined) {
       return result;
```
aruniverse, line 17: why not just return undefined here and on line 28 instead of creating a useless var on line 13?

aruniverse (approved):

PR #91, opened 2021-01-05
Fixed double scrollbars in property grid widget

cortneysmithbentley:
Set overflow: hidden for DIV with the extra scrollbars. Still shows single scrollbar for inner div as expected.


aruniverse (approved):

packages/property-grid/src/components/PropertyGrid.scss
```diff
@@ -6,6 +6,10 @@
 @import "~@bentley/ui-core/lib/ui-core/style/index";
 @import "~@bentley/ui-core/lib/ui-core/text/index";
 
+.pcomponents-property-grid-with-unified-selection {
```
aruniverse, line 9: where is this `pcomponents-` coming from?

cortneysmithbentley: That's the class defined in the presentation-components class. It's what's leading to the extra scroll bars, so I'm overriding it.

aruniverse: ok and i guess the class is coming from @bentley/presentation-components?

aruniverse (approved):

PR #99, opened 2021-02-19
imodel-content-tree: Handle group information elements

grigasp:


packages/imodel-content-tree/package.json
```diff
@@ -31,16 +31,16 @@
...
-    "@bentley/build-tools": "^2.5.1",
-    "@bentley/imodeljs-common": "^2.5.1",
-    "@bentley/imodeljs-frontend": "^2.5.1",
-    "@bentley/presentation-common": "^2.5.1",
-    "@bentley/presentation-components": "^2.5.1",
-    "@bentley/presentation-frontend": "^2.5.1",
-    "@bentley/ui-abstract": "^2.5.1",
-    "@bentley/ui-components": "^2.5.1",
-    "@bentley/ui-core": "^2.5.1",
+    "@bentley/bentleyjs-core": "^2.12.0",
```
calebmshafer, line 34: Is this bump required? If not, we should leave it as low as possible.

calebmshafer (changes requested):

packages/imodel-content-tree/package.json
grigasp: The primary reason I bumped this was to get all the latest performance improvements. I suppose we could as well leave that up to consumers of this package. What do you think?

aruniverse (approved):

calebmshafer (commented):
LGTM

PR #100, opened 2021-02-23
Fix scss issue found in Dart Sass

calebmshafer:
I moved Desktop Starter up to the latest version of Dart Sass and found the following error.  I just added a space and everything seems to be okay :).  Someone with more knowledge of scss can tell me if that was the "correct" solution.

```
SassError: Expected identifier.
   ╷
39 │     -webkit-transition: color background-color.2s ease;
```

### DIFF
--- a/packages/imodel-select/src/components/IModelList.scss
+++ b/packages/imodel-select/src/components/IModelList.scss
@@ -37,9 +37,9 @@ $imodel-select-cards-border-color: $buic-foreground-disabled;
     border: 1px solid #ccc;
     font-size: 18px;
     cursor: pointer;
-    -webkit-transition: color background-color.2s ease;
-    -moz-transition: color background-color.2s ease;
-    -o-transition: color background-color.2s ease;
+    -webkit-transition: color background-color .2s ease;
+    -moz-transition: color background-color .2s ease;
+    -o-transition: color background-color .2s ease;
     transition: color background-color 0.2s ease;
   }
 
--- a/packages/imodel-select/src/components/IModelList.scss
+++ b/packages/imodel-select/src/components/IModelList.scss
@@ -37,9 +37,6 @@ $imodel-select-cards-border-color: $buic-foreground-disabled;
     border: 1px solid #ccc;
     font-size: 18px;
     cursor: pointer;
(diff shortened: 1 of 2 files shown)


packages/imodel-select/src/components/IModelList.scss
```diff
@@ -37,9 +37,9 @@ $imodel-select-cards-border-color: $buic-foreground-disabled;
     border: 1px solid #ccc;
     font-size: 18px;
     cursor: pointer;
-    -webkit-transition: color background-color.2s ease;
-    -moz-transition: color background-color.2s ease;
-    -o-transition: color background-color.2s ease;
+    -webkit-transition: color background-color .2s ease;
```
MichaelBelousov, line 40: sass/scss should generate the vendor specific prefixes for us when compiled to css. I would actually just get rid of all three properties since we already have a correct `transition` property

MichaelBelousov (changes requested):
let's just remove the vendor-specific prefixes since sass makes them redundant

packages/imodel-select/src/components/IModelList.scss
MichaelBelousov: I'm *pretty* sure that auto-prefixing is what happens, but I don't think sass does it, it might be a postcss plugin the react-script's default webpack configuration, I'll need to double check...

calebmshafer: Done, thanks!

calebmshafer:
> let's just remove the vendor-specific prefixes since sass makes them redundant

@MichaelBelousov sounds good, just tested it out in Desktop Starter and that works.  Thanks again!

MichaelBelousov:
> let's just remove the vendor-specific prefixes since sass makes them redundant

So sass *does not* make them redundant. Consumers of imodelselect get the raw scss as imports to the raw ts, so they technically control the import process of the scss. Since we can reasonably expect them to use CRA or otherwise add autoprefixer to their pipeline if they choose not to, I am comfortable with removing the vendor specific prefixes anyway.

MichaelBelousov (approved):

wilmaier (approved):

PR #104, opened 2021-03-23
Tree widget resize

NancyMcCallB:
Reported by a user in imodeljs git repo: https://github.com/imodeljs/imodeljs/discussions/1005

The tree widget can get into a flickering refresh state when it is resized. I solved this by changing the components-selectable-content class from display: flex to display: block. This seems to keep it out of a endless resize loop when the scroll bars are added.

I tested in docked and undocked states and did not see any problems.

### DIFF
--- a/packages/tree-widget/src/components/TreeWidgetComponent.scss
+++ b/packages/tree-widget/src/components/TreeWidgetComponent.scss
@@ -13,7 +13,7 @@
   flex-direction: column;
 
   .components-selectable-content {
-    display: flex;
+    display: block;
     flex-direction: column;
     height: 100%;
     width: 100%;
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

packages/tree-widget/src/components/TreeWidgetComponent.scss
```diff
@@ -13,7 +13,7 @@
   flex-direction: column;
 
   .components-selectable-content {
-    display: flex;
+    display: block;
     flex-direction: column;
```
bsteinbk, line 17: FYI - no need for flex-direction if display is block.

aruniverse: ah good catch bill, i just looked at the changed line. will fix whenever theres a new update to the pkg

PR #113, opened 2021-05-17
Open source measure-tools package

aruniverse:
Open source measure-tools pkg (direct port/no changes to code)
Renamed to measure-tools-react, ver will be reset to 1.0.0 on publish
Cleaned up dev deps, using `^` rather than `~` for core deps


common/changes/@bentley/measure-tools-react/measure-tools_2021-05-17-20-42.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@bentley/measure-tools-react",
+      "comment": "Initial release of bentley/measure-tools-react!",
+      "type": "minor"
```
aruniverse, line 6: @calebmshafer i was hoping to go with major release, 1.0.0 but when i did rush change i didnt see major as an option. im guessing its cause if the locked major in the version policy. Any recommendations? publish this to make sure it works before 1.0.0 or?

packages/measure-tools/README.md
```diff
@@ -0,0 +1,162 @@
...
+```typescript
+    await IModelApp.startup(opts);
+    await MeasureTools.startup();
+```
+
+After that, most of your code changes will be setting up your frontstage to consume the default measurement tools (such as measure distance, polygon, location, etc).
+
+**Note:** Startup can be called any number of times but will only startup once. Some other libraries that use measure-tools may have already initialized it for you!
+
+# What to add in your iTwin-Viewer based app (UI 2.0)
```
calebmshafer, line 19: Can you refer to it as "iTwin App UI" rather than "UI 2.0"?  Trying to remove the very useless name of "UI 2.0" from all docs :)

packages/measure-tools/README.md
```diff
@@ -0,0 +1,162 @@
...
+      ]}
+    />
+  );
+```
+
+# What to add in your frontstage (UI 1.0)
+
+### Tools
+
+Similar to  ```CoreTools``` in iModel.js, measure-tools has a collection of item definitions organized under the ```MeasureToolDefinitions``` singleton. The application can arrange or group the default tool buttons any way it wants.
```
calebmshafer, line 43: `iModel.js` -> `iTwin.js`. Can you update all occurrences?

packages/measure-tools/src/MeasureTools.ts
```diff
@@ -0,0 +1,100 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+
```
calebmshafer, line 6: Double new lines

packages/measure-tools/package.json
```diff
@@ -0,0 +1,115 @@
...
+  },
+  "main": "lib/measure-tools.js",
+  "typings": "lib/measure-tools",
+  "scripts": {
+    "build": "npm run copy:assets && npm run pseudolocalize && tsc 1>&2 && npm run webpack:tests",
+    "copy:assets": "cpx \"./src/**/*.*css\" \"./lib\" && cpx \"./public/**/*\" ./lib/public",
+    "pseudolocalize": "betools pseudolocalize --englishDir ./public/locales/en --out ./lib/public/locales/en-PSEUDO",
+    "watch": "tsc --w",
+    "clean": "rimraf lib .rush/temp/package-deps*.json",
+    "lint": "tslint --project . 1>&2",
```
calebmshafer, line 18: Any reason to not update to use eslint? Maybe we just take a separate pass through the whole repo...

calebmshafer (changes requested):

common/changes/@bentley/measure-tools-react/measure-tools_2021-05-17-20-42.json
calebmshafer: I'd push out a 0.x version and then do the major version bump.

When you do go to bump it, the version policy lock will need to be updated.

packages/measure-tools/package.json
aruniverse: I thought of that, there were some other pkgs using tslint, and thought itd be best to do it all at once.

packages/measure-tools/README.md
aruniverse: Should i remove UI 1.0 thats mentioned below?

packages/measure-tools/package.json
```diff
@@ -0,0 +1,115 @@
...
+    "test": "node ./node_modules/@bentley/certa/bin/certa.js -c certa.json -r chrome",
+    "cover": "nyc npm test",
+    "madge": "madge --circular --extensions ts,tsx src",
+    "webpack:tests": "cross-env NODE_OPTIONS=--max-old-space-size=4096 webpack --config webpack.config.js 1>&2"
+  },
+  "keywords": [
+    "Bentley",
+    "BIM",
+    "iModel",
```
calebmshafer, line 27: Can you add `iTwin` and `iTwinjs` to the tags? I'll probably do that for all the others as well.

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Copyright © Bentley Systems, Incorporated. All rights reserved.
+
+The measure-tools-react package provides measurement tools and React components to access within an iTwin Viewer.
+
+Every iTwin.js application has initialization code where they must call ```IModelApp.startup()```. This library has a similar initialization entrypoint that must be called to configure localization and register tools.
+Call the  ```MeasureTools.startup() ``` **AFTER** the IModelApp one.
+
+```typescript
+    await IModelApp.startup(opts);
```
calebmshafer, line 11: nit: spacing looks strange in markdown preview.

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+    await MeasureTools.startup(opts);
+```
+
+After that, most of your code changes will be to consume the default measurement tools (such as measure distance, polygon, location, etc).
+
+**Note:** Startup can be called any number of times but will only startup once. Some other libraries that use measure-tools may have already initialized it for you!
+
+## What to add in your iTwin-Viewer based app (iTwin App UI)
```
calebmshafer, line 19: nit: remove `-`

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+}
+```
+
+![Measurement Widget](./docs/images/MTDocs-Widget.gif)
+
+# Advanced Topics
+
+## Action Toolbar
+
+Measurements support a right click context menu called the "Action Toolbar". By default it is not enabled, however it is completely customizable by the application. Similar to how tools are constructed by creating item definitions,
```
calebmshafer, line 105: Can you link to more docs on "Action Toolbar", is that on itwinjs.org anywhere?

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+      uiProviders={[
+        new MeasureToolsUiItemsProvider(),
+      ]}
+    />
+  );
+```
+
+## What to add in your frontstage
+
+If you are not using the iTwin-Viewer or iTwin App UI, you can add the measurement tools and measurement property widget to your frontstage like so:
```
calebmshafer, line 41: Is this going to describe the UI 1.0 way of doing it, or is it agnostic to the UI layer?

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+If multiple measurements are selected, the actions apply to all the measurements.
+
+Below is an example to enable the toolbar with a default action provider. Actions are:
+
+- Open Properties (selects the measurement)
+- Delete
+- Toggle display of axes (distance measurement only)
+- Toggle lock state (locked measurements by default are not cleared)
+
+ ```typescript
```
calebmshafer, line 118: nit: leading space

packages/measure-tools/README.md
```diff
@@ -0,0 +1,164 @@
...
+        }}
+        syncEventIds={[MeasurementSyncUiEventId.MeasurementSelectionSetChanged]}
+        stateFunc={MeasurementWidget.generateWidgetStateFuncWithMemory()}
+        />,
+    ]}
+    />
+}
+```
+
+![Measurement Widget](./docs/images/MTDocs-Widget.gif)
```
calebmshafer, line 99: Any idea how the gifs will show up in the README on npmjs.org since they're not published?

calebmshafer (changes requested):

bsy-nicholasw (approved):

packages/measure-tools/README.md
aruniverse: AFAIK, this is a custom feature implemented by civil not in core, but i do think this would be useful to be added there. @bsteinbk thoughts?

packages/measure-tools/README.md
aruniverse: This is the way to add to your ui-1.0 based app. Technically it will work in ui-2.0 as well, but we don't encourage this, devs should be using UiItemProviders to add to their frontstage instead of directly like we do here in 1.0. Also i dont believe we technically expose the term "ui-1.0" externally so i didn't want to add to more confusion. I guess we could technically remove this chunk altogether.... Thoughts?

packages/measure-tools/README.md
aruniverse: I dont believe they would show up on npmjs, i dont like the idea of publishing the gifs and making the deliverable even larger. Maybe we could just say if gifs don't appear look at the actual readme in github?

packages/measure-tools/README.md
bsteinbk: We have the concept of a popup toolbar, but I no nothing about an "Action Toolbar". I also noticed a few uses of "document." to add and remove event listeners.  It is best to get the ownerDocument from an HTMLElement, so this feature is compatible with use in popout windows now available in UI 2.0.  If you continue to use PopupToolbar.tsx you should get the ownerDocument from this._wrapperRef.  See core popup toolbar info at https://github.com/imodeljs/imodeljs/blob/693b71c55a0525c42f6621a2262d50006b9abd8f/docs/learning/ui/abstract/UiAdmin.md#L99 and example use at https://github.com/imodeljs/imodeljs/blob/df437ba429b7fb1395eeae1c34a67baf9b9f7970/test-apps/ui-test-app/src/frontend/tools/AccuDrawPopupTools.tsx#L253 .

packages/measure-tools/README.md
calebmshafer: > I guess we could technically remove this chunk altogether.... Thoughts?

That's what I was debating when originally writing the comment. I'm in favor of doing that since it's pretty much how everything should be setup going forward if you're using our React UI. If you're not using our React UI pkgs then this doesn't help you much.

packages/measure-tools/README.md
aruniverse: I'll add an enhancement to add support for the popout windows, and address it in a later pr. thanks Bill :)

packages/measure-tools/README.md
aruniverse: i updated the gifs to point to github, and according to this [SO post](https://stackoverflow.com/questions/63759426/how-to-show-gif-in-npm-packages-readme-file) should work, we can check once the first ver is released.

calebmshafer (approved):

a-gagnon (approved):

rush.json
```diff
@@ -50,6 +50,12 @@
       "packageName": "@bentley/geo-tools-react",
       "projectFolder": "packages/geo-tools",
       "versionPolicyName": "geo-tools"
+    },
+    {
+      "packageName": "@bentley/measure-tools-react",
+      "projectFolder": "packages/measure-tools",
+      "versionPolicyName": "measure-tools",
+      "shouldPublish": false
```
aruniverse, line 58: @calebmshafer are you cool with this? Or should we wait until tomorrow to decide what to do with this pkg?

calebmshafer: this works with me. Lets get this merged in and then we can figure out what else to do.

PR #115, opened 2021-06-07
Markup: React to breaking changes in core, wrt to popups

aruniverse:
Popups now require setting the parent document.


aruniverse:
This will be a major bump so i'll update it accordingly and remove the major ver lock.

@archana-maharjan , if there are other changes youd like to make in this major ver bump, lets do so. I see theres a deprecated `createMarkupSavedViewData` we are using which we should probably remove.

jason-crow (approved):

archana-maharjan:
> This will be a major bump so i'll update it accordingly and remove the major ver lock.
> 
> @archana-maharjan , if there are other changes youd like to make in this major ver bump, lets do so. I see theres a deprecated `createMarkupSavedViewData` we are using which we should probably remove.

@aruniverse please test in DesignReview to see if removal of "createMarkupSavedViewData" is not breaking, if so green flag from my side to remove it :)

bsteinbk (approved):

aruniverse:
> This will be a major bump so i'll update it accordingly and remove the major ver lock.
> 
> @archana-maharjan , if there are other changes youd like to make in this major ver bump, lets do so. I see theres a deprecated `createMarkupSavedViewData` we are using which we should probably remove.

Discussed with @calebmshafer , will leave as a minor bump. @archana-maharjan we'll schedule a pbi to remove the deprecated stuff in here before the next major

packages/property-grid/src/components/ElementList.tsx
```diff
@@ -4,7 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 
 import * as React from "react";
-import * as classnames from "classnames";
```
bsteinbk, line 7: @aruniverse - Since you are messing with classnames you can update to latest package "classnames": "^2.3.1", and remove the related types files as the new package delivers types.  That is what I did in imodeljs.

bsteinbk (approved):

calebmshafer (approved):

PR #116, opened 2021-06-09
Feature/Property Grid Filtering

usjpin:
adding feature to filter records on the property-grid component using FilteringPropertyDataProvider, as well as using HOC to support unified selection.


packages/property-grid/package.json
```diff
@@ -52,11 +52,11 @@
     "@bentley/itwin-client": "^2.6.0",
     "@bentley/orbitgt-core": "^2.6.0",
     "@bentley/presentation-common": "^2.6.0",
-    "@bentley/presentation-components": "^2.6.0",
+    "@bentley/presentation-components": "^2.16.2",
```
aruniverse, line 55: did you need to bump to 2.16.2? whats the lowest core ver your changes are compatible with?

packages/property-grid/package.json
```diff
@@ -52,11 +52,11 @@
...
     "@bentley/presentation-frontend": "^2.6.0",
     "@bentley/product-settings-client": "^2.6.0",
     "@bentley/ui-abstract": "^2.6.0",
-    "@bentley/ui-components": "^2.6.0",
+    "@bentley/ui-components": "^2.16.2",
```
aruniverse, line 59: if you bump one core ver, then you need to bump them all

aruniverse: so if it is 2.16.2 you need, you need to bump all deps on 2.6.0 -> 2.16.2

packages/property-grid/src/components/PropertyGridWithUnifiedSelection.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return result;
+  }
+}
+
+interface PropertyGridWithUnifiedSelectionProps extends VirtualizedPropertyGridWithDataProviderProps {
+  filterer: PropertyDataFiltererBase;
+}
+
+export function PropertyGridWithUnifiedSelection(props: PropertyGridWithUnifiedSelectionProps): JSX.Element {
+  const { isOverLimit /*, numSelectedElements*/ } = usePropertyDataProviderWithUnifiedSelection(
```
aruniverse, line 48: is numSelectedElements not needed? If so, id remove that and the code block on L60-L66 thats commented out

aruniverse (commented):
CI build is failing on missing rush changes. 

@jason-crow @calebmshafer , should this be a major bump cause of the core ver bump or?

calebmshafer:
@aruniverse I think it should be.

packages/property-grid/src/components/PropertyGridWithUnifiedSelection.tsx
usjpin: That return value has not been added to the latest package release, and so I had just commented it out to include it as necessary when the newer version comes out.

aruniverse: is that coming out in 2.17? id personally rather not bump the major back to back, and would just hold this pr until the ver is out if thats cool with @jason-crow

jason-crow: @aruniverse is does appear to be in the 2.17 release branch:
https://github.com/imodeljs/imodeljs/blob/release/2.17.x/presentation/components/src/presentation-components/propertygrid/UseUnifiedSelection.tsx

however, wouldn't this just be undefined until 2.17? in otherwords, couldn't you skirt the major bump by just checking 
if (numSelectedElements !== undefined && numSelectedElements === 0) 
and just keep it uncommented with a comment saying 'numSelectedElements will be undefined why used by versions older than 2.17 of presentation-components'

aruniverse: i like that! as long as @calebmshafer  is in agreement, we can do this

jason-crow: will probably want to cast it like this to make it not throw a lint error:
as UsePropertyDataProviderWithUnifiedSelectionResult & { numSelectedElements: number | undefined }

calebmshafer: Do we intend to drop the minimum version again and then do what @jason-crow suggested?  If so, I agree I like it to avoid a major version bump.

aruniverse: Talking to Utkarsh, sounded like what he needs in this pr is the `usePropertyDataProviderWithUnifiedSelection` hook added in [2.6.0](https://github.com/imodeljs/imodeljs/blob/383386291a45f0b3435bcb44e83840ee02f31f5a/presentation/components/CHANGELOG.md#updates-9) and `FilteringPropertyDataProvider` added in [2.7.0](https://github.com/imodeljs/imodeljs/commit/5c325007bce78d8ce0ce74e505f113b8b5ce31cb#diff-45915886c77b97687147d5b46c5258952c9999ab40d5a6796610dfc5c51a0f2e) _I think_, so we'll probably have to bump to at least ^2.7.0

aruniverse (changes requested):

packages/property-grid/src/components/PropertyGridWithUnifiedSelection.tsx
```diff
@@ -0,0 +1,69 @@
...
+export function PropertyGridWithUnifiedSelection(props: PropertyGridWithUnifiedSelectionProps): JSX.Element {
+  const { isOverLimit , numSelectedElements } = (usePropertyDataProviderWithUnifiedSelection as any) (
+    { dataProvider: props.dataProvider as IPresentationPropertyDataProvider },
+  );
+  const filteringDataProvider = new FilteringPropertyDataProvider(props.dataProvider, props.filterer);
+  const autoExpandingFilteringDataProvider = new AutoExpandingPropertyDataProvider(filteringDataProvider);
+  if (isOverLimit) {
+    return (
+      <FillCentered>
+        {IModelApp.i18n.translate("Sample:property-grid.too-many-elements-selected")}
```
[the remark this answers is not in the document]
usjpin: just added those labels, had completely missed them

aruniverse (approved):
looks fine to me, but will also want @jason-crow and @calebmshafer approval before this merges in

jason-crow (approved):

PR #118, opened 2021-07-01
update build command to allow warnings in a successful build to fix ci

aruniverse:
This should fix the ci and unblock  #116

### DIFF
--- a/common/changes/@bentley/geo-tools-react/successful_warnings_2021-07-08-18-30.json
+++ b/common/changes/@bentley/geo-tools-react/successful_warnings_2021-07-08-18-30.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@bentley/geo-tools-react",
+      "comment": "add rebuild script",
+      "type": "none"
+    }
+  ],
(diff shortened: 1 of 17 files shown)


common/changes/@bentley/geo-tools-react/successful_warnings_2021-07-08-18-30.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@bentley/geo-tools-react",
+      "comment": "add rebuild script",
+      "type": "none"
```
aruniverse, line 6: @calebmshafer i had to manually modify these files to set the type to none because the only options available were path or minor. 
Will this be an issue?

jason-crow (approved):

common/changes/@bentley/geo-tools-react/successful_warnings_2021-07-08-18-30.json
calebmshafer: Making it none will cause it not to bump the version of the package.

PR #119, opened 2021-07-09
Added a UiProvider for Tree Widget

spap975:
Using existing code and props from TreeWidgetControlOptions, I created a UiProvider which functions the same as the Tree Widget.  I am more than happy to reduce the size of my changes if there are better options for implementation.


packages/tree-widget/package.json
```diff
@@ -62,8 +62,8 @@
     "@types/react-dom": "^16.8.0",
     "@types/react-redux": "^7.0.1",
     "cpx": "^1.5.0",
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0",
+    "react": "^16.13.0",
```
aruniverse, line 65: Did you need to bump these versions?

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+  SpatialContainmentTree,
+  ClassGroupingOption
+} from "@bentley/ui-framework";
+import React from "react";
+import { TreeWidgetComponent } from "./TreeWidgetComponent";
+import { CategoriesTreeComponent, ModelsTreeComponent, TreeWidgetControlOptions } from "../tree-widget-react";
+import { IModelConnection, Viewport } from "@bentley/imodeljs-frontend";
+import { SelectableContentDefinition } from "@bentley/ui-components";
+
+/** Provides the property grid widget to zone 9 */
```
aruniverse, line 19: Comment is irrelevant here, not property-grid widget, and there is non zone 9 in ui-2.0

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+export class TreeUiItemsProvider implements UiItemsProvider {
```
aruniverse, line 20: Maybe should be called `TreeWidgetUiItemsProvider`? Idk, probably too nit picky.
@calebmshafer thoughts?

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+  UiItemsProvider,
+} from "@bentley/ui-abstract";
+import {
+  UiFramework,
+  SpatialContainmentTree,
+  ClassGroupingOption
+} from "@bentley/ui-framework";
+import React from "react";
+import { TreeWidgetComponent } from "./TreeWidgetComponent";
+import { CategoriesTreeComponent, ModelsTreeComponent, TreeWidgetControlOptions } from "../tree-widget-react";
```
aruniverse, line 15: Could you import these 3 from the actual module and not the barrel file?

aruniverse: Pretty sure we'll hit circular import issues; or atleast we do in other pkgs when we import things from the same pkg from the barrel

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { IModelConnection, Viewport } from "@bentley/imodeljs-frontend";
+import { SelectableContentDefinition } from "@bentley/ui-components";
+
+/** Provides the property grid widget to zone 9 */
+export class TreeUiItemsProvider implements UiItemsProvider {
+  public readonly id = "TreeUiitemsProvider";
+  public static i18n: I18N;
```
aruniverse, line 22: this appears to be unused, would remove

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+          render: this._categoriesTreeReplacement ? this._categoriesTreeReplacement : () => categoriesTreeComponent,
+        },
+        {
+          label: UiFramework.translate("visibilityWidget.containment"),
+          id: "spatial-containment-tree",
+          render: this._spatialTreeReplacement ? this._spatialTreeReplacement : () => spatialContainmentComponent,
+        },
+      ];
+
+      if (this._additionalTrees) trees.push(...this._additionalTrees);
```
aruniverse, line 114:
```suggestion
      if (this._additionalTrees) {
          trees.push(...this._additionalTrees);
      }
```

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
```diff
@@ -0,0 +1,127 @@
...
+              ? ClassGroupingOption.Yes
+              : ClassGroupingOption.No
+          }
+          {...this._spatialTreeProps}
+        />
+      );
+
+      const trees: SelectableContentDefinition[] = [
+        {
+          label: UiFramework.translate("visibilityWidget.modeltree"),
```
aruniverse, line 98: I guess using `UiFramework.translate` is fine, but id rather we use `IModelApp.i18n.translate`, or use `TreeUiItemsProvider.i18n.translate`

aruniverse (commented):
Please update the pkg readme with instructions on how to add the UiProivder into a viewer. I would argue we should remove whatever documentation we have for ui-1.0.

You're missing rush changes. This should be a minor bump.

packages/tree-widget/package.json
spap975: yeah I was running into issues when testing the provider in the web-viewer-test

aruniverse:
@spap975 Take a look at https://github.com/imodeljs/imodeljs/pull/1620, Bill add support for UiProviders to add widgets in 1.0. 
@kckst8 is that worth adding in here you think? It will help teams transition from ui-1.0 -> 2.0.

aruniverse:
@bsteinbk could you take a pass over this pr since your our UiItemsProvider expert? :)

packages/tree-widget/package.json
aruniverse: I think thats because you copied the entire pkg over into that repo, id revert these changes in this file since we know it works because its being used in itwin/viewer-react

packages/tree-widget/src/components/TreeUiItemsProvider.tsx
aruniverse: Actually, should use TreeWidget.i18n

aruniverse:
> @spap975 Take a look at [imodeljs/imodeljs#1620](https://github.com/imodeljs/imodeljs/pull/1620), Bill add support for UiProviders to add widgets in 1.0.
> @kckst8 is that worth adding in here you think? It will help teams transition from ui-1.0 -> 2.0.

Update: Im going to say ignore this comment, we'll nee to bump the core ver of this pkg and i dont want to do that in this pr.

packages/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -0,0 +1,131 @@
...
+              ? ClassGroupingOption.Yes
+              : ClassGroupingOption.No
+          }
+          {...this._spatialTreeProps}
+        />
+      );
+
+      const trees: SelectableContentDefinition[] = [
+        {
+          label: TreeWidget.translate("visibilityWidget.modeltree"),
```
aruniverse, line 100: These strings are missing.... 
Could you add them to https://github.com/imodeljs/viewer-components-react/blob/master/packages/tree-widget/public/locales/en/TreeWidget.json

aruniverse: Please update https://github.com/imodeljs/viewer-components-react/blob/master/packages/tree-widget/src/components/TreeWidgetControl.tsx such that it doesn't use `UiFramework.translate`, but rather `TreeWidget.translate`

packages/tree-widget/package.json
aruniverse: For some reason github says theres still changes to this file, might be line endings or something.

could you run:
```
git checkout origin/master packages/tree-widget/package.json
```

packages/tree-widget/README.md
```diff
@@ -180,3 +180,20 @@ export class TreeWidgetControl extends WidgetControl {
   }
 }
 ```
+
+### Using the TreeWidgetUiItemsProvider
+
+The `TreeWidgetUiItemsProvider` functions like the `TreeWidgetControl` but is instead passed through the `uiProviders` prop of the `<Viewer>` component.
```
aruniverse, line 186: We still need to initialize TreeWidget. So add that to the docs here, and update TreeWidget.initialize to be a lazy init, just in case it tries to get initialized more than once.

packages/tree-widget/README.md
```diff
@@ -180,3 +180,20 @@ export class TreeWidgetControl extends WidgetControl {
...
+
+```ts
+<Viewer
+  authConfig={{ oidcClient: AuthorizationClient.oidcClient }}
```
aruniverse, line 190: i'd honestly remove all the other unused props here. take a look at the snippet i added here https://github.com/imodeljs/viewer-components-react/tree/master/packages/measure-tools#what-to-add-in-your-itwin-viewer-based-app-itwin-app-ui

aruniverse:
Did the tree's actually show up for you? I keep getting an infinite spinner... 
image
Might be due to how i added this into  the web-viewer-test...
How did you obtain the iModelConnection that you passed into the UiItemsProvider?
If im not mistaken this is what i saw you had when you shared your screen earlier but i dont like using `!` to assert that `UiFramework.getIModelConnection()`  is defined.

image

Makes me think that we should make the iModelConnection property optional for the UiItemsProvider constructor, and if its not defined get the connection directly from UiFramework in there

bsteinbk (approved):

packages/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -0,0 +1,152 @@
...
+        },
+      ];
+
+      if (this._additionalTrees) {
+        trees.push(...this._additionalTrees);
+      }
+
+      widgets.push({
+        id: "tree",
+        label: "Tree View",
```
aruniverse, line 143: this should be localized

packages/tree-widget/src/TreeWidget.ts
```diff
@@ -13,12 +13,17 @@ import { UiError, getClassName } from "@bentley/ui-abstract";
  */
 export class TreeWidget {
   private static _i18n?: I18N;
+  private static _initialized: boolean;
 
   /**
    * Called by IModelApp to initialize the Tree Widget
    * @param i18n - The internationalization service created by the IModelApp.
    */
   public static async initialize(i18n: I18N): Promise<void> {
```
aruniverse, line 22: could you make this optional(the arg), and default to IModelApp.i18n if i18n is undefined?

packages/tree-widget/README.md
```diff
@@ -180,3 +180,15 @@ export class TreeWidgetControl extends WidgetControl {
...
 ```
+
+### Using the TreeWidgetUiItemsProvider
+
+The `TreeWidgetUiItemsProvider` functions like the `TreeWidgetControl` but is instead passed through the `uiProviders` prop of the `<Viewer>` component. `TreeWidget` must be initialized before use.
+
+```ts
+<Viewer
+  ...
+  onIModelAppInit={()=>{TreeWidget.initialize(IModelApp.i18n)}}
```
aruniverse, line 191: Initialize is an async call

packages/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -0,0 +1,152 @@
...
+} from "@bentley/ui-framework";
+import React from "react";
+import { TreeWidgetComponent } from "./TreeWidgetComponent";
+import { CategoriesTreeComponent } from "./trees/CategoriesTree";
+import { ModelsTreeComponent } from "./trees/ModelsTree";
+import { IModelConnection, Viewport } from "@bentley/imodeljs-frontend";
+import { SelectableContentDefinition } from "@bentley/ui-components";
+import { TreeWidget } from "../TreeWidget";
+
+export interface TreeWidgetUiItemsProviderProps {
```
aruniverse, line 21: Redefining everything in TreeWidgetControlOptions seems unnecessary for when all you're trying to do is making the iModelConnection optional.

I __think__ this will suffice, please test it out

```ts
export interface TreeWidgetUiItemsProviderProps extends TreeWidgetControlOptions {
   iModelConnection?: IModelConnection;
}
```

packages/tree-widget/README.md
```diff
@@ -180,3 +180,15 @@ export class TreeWidgetControl extends WidgetControl {
   }
 }
 ```
+
+### Using the TreeWidgetUiItemsProvider
+
+The `TreeWidgetUiItemsProvider` functions like the `TreeWidgetControl` but is instead passed through the `uiProviders` prop of the `<Viewer>` component. `TreeWidget` must be initialized before use.
```
aruniverse, line 186: nit: should be `<Viewer />` not `<Viewer>`.

packages/tree-widget/README.md
```diff
@@ -180,3 +180,15 @@ export class TreeWidgetControl extends WidgetControl {
   }
 }
 ```
+
+### Using the TreeWidgetUiItemsProvider
```
aruniverse, line 184: @calebmshafer @kckst8 

How do we feel about this? Should this be moved up to above the readme section for ui-1.0? Do we want to remove the old way from the readme?

We need to start standardizing how we document these Extensions/UiProivders

aruniverse (commented):
looks good. Still waiting to hear back about if this actually worked for you or if you just got an infinite spinner

packages/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
spap975: I tried doing this before making the new interface and i got an error, that's why i made this one

aruniverse: What was the error?

spap975: Interface 'TreeWidgetUiItemsProviderProps' incorrectly extends interface 'TreeWidgetControlOptions'.
  Types of property 'iModelConnection' are incompatible.
    Type 'IModelConnection | undefined' is not assignable to type 'IModelConnection'.
      Type 'undefined' is not assignable to type 'IModelConnection'.

spap975:
> Did the tree's actually show up for you? I keep getting an infinite spinner...
> image
> Might be due to how i added this into the web-viewer-test...
> How did you obtain the iModelConnection that you passed into the UiItemsProvider?
> If im not mistaken this is what i saw you had when you shared your screen earlier but i dont like using `!` to assert that `UiFramework.getIModelConnection()` is defined.
> 
> image
> 
> Makes me think that we should make the iModelConnection property optional for the UiItemsProvider constructor, and if its not defined get the connection directly from UiFramework in there



> Did the tree's actually show up for you? I keep getting an infinite spinner...
> image
> Might be due to how i added this into the web-viewer-test...
> How did you obtain the iModelConnection that you passed into the UiItemsProvider?
> If im not mistaken this is what i saw you had when you shared your screen earlier but i dont like using `!` to assert that `UiFramework.getIModelConnection()` is defined.
> 
> image
> 
> Makes me think that we should make the iModelConnection property optional for the UiItemsProvider constructor, and if its not defined get the connection directly from UiFramework in there

I switched it so the iModelConnection is loaded in the UiProvider, if the user passes a different iModelConnection through props it'll prioritize that and default to UiFramework.getIModelConnection()

packages/tree-widget/README.md
```diff
@@ -180,3 +180,19 @@ export class TreeWidgetControl extends WidgetControl {
...
+
+The `TreeWidgetUiItemsProvider` functions like the `TreeWidgetControl` but is instead passed through the `uiProviders` prop of the `<Viewer />` component. `TreeWidget` must be initialized before use.
```
kckst8, line 186:
```suggestion
The `TreeWidgetUiItemsProvider` functions like the `TreeWidgetControl` but is instead passed through the `uiProviders` prop of the iTwin Viewer's `<Viewer />` component. `TreeWidget` must be initialized before use.
```

packages/tree-widget/README.md
kckst8: I think we remove the 1.0 example entirely. Nobody should be using that any longer

packages/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
aruniverse: Ok guess you can only extend and make an optional prop required, not vice versa :/

Anyway, can type the arg in the constructor as `Partial<TreeWidgetControlOptions>` like so: 
image

packages/tree-widget/README.md
```diff
@@ -2,181 +2,28 @@
...
+The tree-widget-react package provides a UiProvider class - `TreeWidgetUiItemsProvider` - which can be passed into the `uiProviders` prop of the iTwin Viewer's `<Viewer />` component to provide access to the following features:
 
 - Spatial Containment Tree
 
 - Models Tree
 
 - Categories Tree
 
-The package also provides the component which underlies `TreeWidgetControl` - `TreeWidgetComponent` - which you can wrap within your own custom widget controller and pass in your own custom trees to display.
+The package also provides the widget controller class - `TreeWidgetControl` - as well as the underlying component - `TreeWidgetComponent` - which you can wrap within your own custom UiProvider and pass in your own custom trees to display.
```
aruniverse, line 13: tehcnically can't use `TreeWidgetControl` in a UiProvider(ui-2.0), that's more for adding directly into a FrontstageDef (ui-1.0)

I would probably just drop the mention of the Control class.
```suggestion
The package also provides the underlying component - `TreeWidgetComponent` - which you can wrap within your own custom UiProvider and pass in your own custom trees to display.
```

aruniverse (approved):

PR #120, opened 2021-07-12
React to breaking changes in imjs 2.17

aruniverse:
React to breaking changes in 2.17


common/changes/@bentley/markup-frontstage-react/update-rush-pnpm_2021-07-01-19-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@bentley/markup-frontstage-react",
+      "comment": "Update iTwin.js minimum dependencies to ^2.17.0",
+      "type": "minor"
```
calebmshafer, line 6: do we want to make this more than just a minor bump?

aruniverse: Yea i was wondering about that too, we didnt the last time because theres deprecated code in here that we haven't removed yet, so unless we want to carry forward the deprecated code into the next major then sure.

jason-crow (approved):

common/changes/@bentley/markup-frontstage-react/update-rush-pnpm_2021-07-01-19-33.json
jason-crow: seems like the only correct approach here would be to make these major, otherwise if someone had a ^ and didnt update to 2.17 they would be broken right?

aruniverse: Updated to major. fyi @archana-maharjan

PR #123, opened 2021-07-21
Property Grid - Null Values Visibility State Persistence

usjpin:
Added persistence to the product setting service for the state of null value visibility in the property grid.


usjpin:
There seem to be two rush change files added to this PR (from my last filtering property grid PR and from this one). Should I be removing the older one?

aruniverse:
> There seem to be two rush change files added to this PR (from my last filtering property grid PR and from this one). Should I be removing the older one?

Since you forked the repo, i think you need to set another remote branch to pull the latest from the actual repo. Then rebase your pr, on top of that. That should ideally remove the extra rush change and extra commits

PR #125, opened 2021-08-06
Add breakdown trees

MeghaCBorty:
Created a new package @bentley/breakdown-trees-react and added the breakdown trees there.


packages/breakdown-trees/src/Views/TreeWithRuleset.tsx
```diff
@@ -0,0 +1,242 @@
...
+  }
+  const filterApplied = filteredNodeLoader !== nodeLoader;
+  const visibleNodes = useVisibleTreeNodes(filteredNodeLoader.modelSource);
+  const viewPort = IModelApp.viewManager.selectedView;
+  const visibilityHandler = props.enableVisibility ? useVisibilityHandler(filteredNodeLoader.dataProvider.rulesetId, filteredNodeLoader.dataProvider, filteredNodeLoader.modelSource, viewPort) : undefined;
+  const unifiedSelectionEventHandler = useSelectionTrackingUnifiedSelectionTreeEventHandler({ nodeLoader: filteredNodeLoader, collapsedChildrenDisposalEnabled: false, onNewSelectionSetCallback, visibilityHandler });
+
+  const toolBarSection = props.searchTools ? searchBar : functionsToolbar;
+  const selectionSet = props.iModel.selectionSet.elements;
+  React.useMemo(() => {
```
jason-crow, line 177: shouldn't this be a useEffect ?

packages/breakdown-trees/src/Views/TreeWithRuleset.tsx
```diff
@@ -0,0 +1,242 @@
...
+  React.useEffect(() => () => previous.current?.dispose(), []);
+
+  return React.useMemo(() => {
+    if (previous.current)
+      previous.current.dispose();
+
+    const handler = createVisibilityHandler(rulesetId, treeDataProvider, modelSource, activeView);
+    previous.current = handler;
+    return handler;
+  }, [rulesetId, activeView]);
```
jason-crow, line 220: missing treeDataProvider, modelSource in dependency array

packages/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -0,0 +1,190 @@
...
+    queries.push(`SELECT sensor.ECInstanceId as id, sensor.UserLabel as label, sensor.CodeValue as code, sensor.Origin as origin, link.SourceECInstanceId as composingId, sensor.BACnet_CO2Sensor as airQualitySensorId, sensor.AKS_Nummer as aks_number
+    FROM RevitDynamic._300S_Kanal_Luftqualitatsfuhler_QPM sensor
+    JOIN Bis.ElementRefersToElements link ON sensor.ECInstanceId = link.TargetECInstanceId`);
+
+    for (const query of queries) {
+      try {
+        rows = rows.concat(await this.executeQuery(iModel, query));
+      } catch (e) {
+        // tslint:disable-next-line:no-console
+        console.log(e);
```
jason-crow, line 121: would suggest considering replacing  console.log with a Logger.logError

packages/breakdown-trees/src/Views/visibility/SectioningUtil.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+      if (this._spaceDecorator) {
+        this._spaceDecorator.clearDecorations();
+      }
+      return false;
+    }
+    const storyRange = await DataLink.queryStoryRange(imodel, storyId, clipAtSpaces ? clipAtSpaces : false);
+    if (storyRange === undefined)
+      return false;
+    // const planeSet = ConvexClipPlaneSet.createRange3dPlanes(storyRange, false, false, false, false, false, false);
```
jason-crow, line 79: maybe cleanup commented code or add a note explaining it

packages/breakdown-trees/src/Views/visibility/SectioningUtil.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  private static async addSpaceDecorations(imodel: IModelConnection, roomIds: any) {
+    if (this._spaceDecorator) {
+      await this._spaceDecorator.addDecorationsForSpaces(imodel, roomIds, this._handleSpaceDecorationClicked);
+    }
+  }
+
+  private static _handleSpaceDecorationClicked = (ev: BeButtonEvent, decoration?: SpaceLabelDecoration) => {
+    // tslint:disable-next-line:no-console
+    console.log(`Event: ${ev}, Marker: ${decoration}`);
```
jason-crow, line 143: would suggest replacing with Logger.logInfo

packages/breakdown-trees/src/Views/visibility/SpaceLabelDecoration.ts
```diff
@@ -0,0 +1,65 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
```
jason-crow, line 5: prefer to not have 2 consecutive blank lines

packages/breakdown-trees/src/Views/visibility/SpaceLabelDecorator.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  public showDecorations() {
+    this._roomLabelDecoration.forEach((decoration) => {
+      decoration.show();
+    });
+  }
+
+  public async addDecorationsForSpaces(imodel: IModelConnection, roomIds: any, onClickHandler: DecorationClickedHandler) {
+    const roomProps: any[] = await imodel.elements.getProps(roomIds[0]);
```
jason-crow, line 50: prefer to not use `any`

packages/breakdown-trees/src/Views/visibility/SpaceLabelDecorator.ts
```diff
@@ -0,0 +1,164 @@
...
+
+      const decorationData = this.createDecorationData(room, roomId, roomOrigin, massProps);
+      spaceDecorations.push(decorationData);
+
+    }
+
+    this.addSpaceDecorations(spaceDecorations, minHeight, onClickHandler);
+  }
+
+  private createDecorationData(room: any, roomId: any, roomOrigin: number[], massProps: MassPropertiesResponseProps): SpaceDecorationCreationProps | undefined {
```
jason-crow, line 94: most of the comments i'v left on `any` have been essentially just recommendations, however using roomId as typed `any` seems especially wrong, i do get how this code is likely years old, but this may be a good opportunity to modernize

jason-crow (changes requested):

packages/breakdown-trees/src/Views/ClassificationsTree.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const classificationsTree = React.useMemo(() => (<ControlledTreeWrapper
+    iModel={props.iModel}
+    loadedRuleset={classificationRules as Ruleset}
+    dataProvider={dataProvider}
+    treeName={CLASSIFICATIONS_TREE_NAME} treeNodeIconMapper={functionIconMapper} optionItems={optionItems}
+    searchTools={true}
+    displayGuids={props.displayGuids}
+    setIsDisplayGuids={props.setIsDisplayGuids} enableVisibility={props.enableVisibility ? props.enableVisibility : false}
+  />), [props.iModel, props.displayGuids, props.setIsDisplayGuids]);
```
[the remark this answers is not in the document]
MeghaCBorty: Added missing dependencies wherever required and also used props.iModel.key for non-primitives props.iModel.

packages/breakdown-trees/src/Views/visibility/SpaceLabelDecorator.ts
MeghaCBorty: Fixed the code to avoid any.

packages/breakdown-trees/src/Views/visibility/DataLink.ts
MeghaCBorty: Found some unused functions . so deleted them.

jason-crow (approved):

packages/breakdown-trees/package.json
```diff
@@ -0,0 +1,141 @@
+{
+  "name": "@bentley/breakdown-trees-react",
```
aruniverse, line 2: @calebmshafer should this be itwin scoped since this repo is going to be deprecated soon?

packages/breakdown-trees/package.json
```diff
@@ -0,0 +1,141 @@
...
+    "@bentley/imodeljs-frontend": "^2.15.0",
+    "@bentley/imodeljs-i18n": "^2.15.0",
+    "@bentley/imodeljs-markup": "^2.15.0",
+    "@bentley/imodeljs-quantity": "^2.15.0",
+    "@bentley/itwin-client": "^2.15.0",
+    "@bentley/presentation-common": "^2.15.0",
+    "@bentley/presentation-components": "^2.15.0",
+    "@bentley/presentation-frontend": "^2.15.0",
+    "@bentley/presentation-testing": "^2.15.0",
+    "@bentley/presentation-backend": "^2.15.0",
```
aruniverse, line 55: really need presentation-backend in this pkg?

packages/breakdown-trees/package.json
```diff
@@ -0,0 +1,141 @@
...
+    "@bentley/product-settings-client": "^2.15.0",
+    "@bentley/ui-abstract": "^2.15.0",
+    "@bentley/ui-components": "^2.15.0",
+    "@bentley/ui-core": "^2.15.0",
+    "@bentley/ui-framework": "^2.15.0",
+    "@bentley/ui-ninezone": "^2.15.0",
+    "@bentley/webgl-compatibility": "^2.15.0",
+    "@bentley/webpack-tools": "1.14.1",
+    "@bentley/measure-tools-react": "workspace:*",
```
aruniverse, line 64: idk if we want to be explicitly depending on other pkgs in the repo like this. might be ok, but probably want to use ^ for the peer dep

packages/breakdown-trees/package.json
calebmshafer: yes, but when it does I think we should bundle that with it consuming 3.0 and the new `@itwin` based packages. Otherwise we'll just introduce another breaking change.

MeghaCBorty:
Was getting the build error in pipeline after merging with latest master and was not able to fix that. So created a fresh branch  from the latest master and copied the package (which already took care of all the comments in this PR) there. Here is the link to the duplicate and completed PR
https://github.com/imodeljs/viewer-components-react/pull/141

PR #129, opened 2021-08-18
Update rush and pnpm

calebmshafer:
Update to the latest pnpm version and turn on use of workspaces. In addition updated to the latest Rush version.

The latest Rush version required me to change up the way copyright linter is used to follow all the other repos and update the minimum Node version supported to Node >12.17.

Also, to avoid having to update all packages to the latest Typescript due to a change in `@bentley/build-tools` I locked all packages down to a specific version.


aruniverse (commented):
didnt get a chance to pull and test locally, but lgtm

aruniverse (approved):

PR #130, opened 2021-08-18
Update measurement tools with internal updates

calebmshafer:
Pull over the internal updates made to the tool as not everyone is using this version yet.

After this change, we can hopefully make that the case and all new updates will be made directly to this package.

### DIFF
--- a/packages/measure-tools/src/api/FeatureTracking.ts
+++ b/packages/measure-tools/src/api/FeatureTracking.ts
@@ -1,6 +1,6 @@
 /*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
+* See COPYRIGHT.md in the repository root for full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
 import { BeUiEvent, GuidString, Logger } from "@bentley/bentleyjs-core";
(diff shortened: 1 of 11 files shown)


packages/measure-tools/src/api/FeatureTracking.ts
```diff
@@ -1,6 +1,6 @@
 /*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
+* See COPYRIGHT.md in the repository root for full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
 import { BeUiEvent, GuidString, Logger } from "@bentley/bentleyjs-core";
```
aruniverse, line 6: hate that this exists, we should chat about this later

packages/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -102,7 +102,7 @@ export class MeasureAreaTool extends MeasurementToolBase<AreaMeasurement, Measur
           hints.setRotation(mat.inverse()!);
       }
     }
-    hints.sendHints(true);
+    hints.sendHints(false);
```
aruniverse, line 105: what are the points of hints if we dont show them? 
@a-gagnon @bsy-nicholasw

aruniverse (approved):

packages/measure-tools/src/api/FeatureTracking.ts
calebmshafer: lol yea, I'll clean all this up in a separate PR.

packages/measure-tools/src/tools/MeasureAreaTool.ts
aruniverse: spoke with nicholas about this, users need to manually turn on accusnap to see these now

PR #131, opened 2021-08-18
Update measure tools to use eslint and raise itwin.js minimum

calebmshafer:


packages/measure-tools/src/api/MeasureToolsLoggerCategory.ts
```diff
@@ -11,6 +11,7 @@ export enum MeasureToolsLoggerCategory {
   Root = "Measure-Tools",
 }
 
+// eslint-disable-next-line @typescript-eslint/no-redeclare
```
aruniverse, line 14: where is this being redeclared?

aruniverse (approved):

aruniverse:
did you bump the core ver? or mean to?

packages/measure-tools/src/api/MeasureToolsLoggerCategory.ts
calebmshafer: That's due to the namespace and enum. I really don't understand why both exist but thought that'd be a different PR, and later problem to fix.

calebmshafer:
> did you bump the core ver? or mean to?

@aruniverse yup, just pushed that change. A couple of the changes I made to clean up the lint were added in 2.19.

calebmshafer:
@aruniverse mind taking another quick look? Pushed a few more updates while I was already updating the minimum version needed.

Also, do we want to try and get a release out with the cherry-picked changes and the 2.17 minimum prior to updating all the way to 2.19?

aruniverse:
> @aruniverse mind taking another quick look? Pushed a few more updates while I was already updating the minimum version needed.
> 
> Also, do we want to try and get a release out with the cherry-picked changes and the 2.17 minimum prior to updating all the way to 2.19?

new changes look fine to me, i dont think theres much usage of this pkg, so id go straight to 2.19, but whatever is easiest for you

calebmshafer:
@aruniverse that's easier for me. We'll go straight there. Thanks!

PR #137, opened 2021-08-24
PropertyGrid UiProvider now shows multi element property grid

aruniverse:


packages/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -9,10 +9,11 @@ import {
   StagePanelSection,
   StageUsage,
   UiItemsProvider,
-  WidgetState
+  WidgetState,
 } from "@bentley/ui-abstract";
 import * as React from "react";
-import { FunctionalPropertyGridWidget } from "./components/FunctionalPropertyGridWidget";
+
+import { FunctionalMultiElementPropertyGrid } from "./components/functional";
```
kckst8, line 16: do we need to name this "functional"? Theoretically you should be able to refactor a class component into a functional component without breaking changes. Do we need both?

packages/property-grid/src/components/functional/FunctionalElementList.tsx
```diff
@@ -0,0 +1,96 @@
...
+    React.useState<SimpleTableDataProvider>();
+
+  const labelsProvider: PresentationLabelsProvider = React.useMemo(() => {
+    return new PresentationLabelsProvider({
+      imodel: iModelConnection,
+    });
+  }, [iModelConnection]);
+
+  React.useEffect(() => {
+    const onMount = async () => {
```
kckst8, line 39: this could/will execute more often than on mount....not sure the name makes sense?

packages/property-grid/src/components/functional/FunctionalElementList.tsx
```diff
@@ -0,0 +1,96 @@
...
+    async (rowIterator: AsyncIterableIterator<RowItem>, _replace: boolean) => {
+      for await (const row of rowIterator) {
+        if (row.extendedData?.key) {
+          onSelect(row.extendedData.key);
+          return false;
+        }
+      }
+      return false;
+    },
+    [onSelect]
```
kckst8, line 58: should we add a note that this prop should be a memoized function? Otherwise this function could be updated unnecessarily with a function as a dependency I think

packages/property-grid/src/PropertyGridUiItemsProvider.tsx
aruniverse: Main reason i did this was i didn't want to break exisiting applications and make this a major ver bump with 3.0 around the corner + scope change. The class based components currently have some DR based functionality like featureTracking and debugLogs that i didnt port over. I was originally thinking removing all those and the class based components in the new pkg, but if we want i can make the change now.

kckst8: ok that makes sense. Just not sure we should name components based on their component type. Wonder if another name makes sense that we don't have to change again once the class component is inevitably deprecated...Is it just different because it is missing those few features? If so, maybe something like BasicMultiElementPropertyGrid?

aruniverse: Yea basically... (pun inteded 😁), will update component names.

aruniverse: Went back and added that one featureTracking call i didnt port over, and then just replaced the class components with the functional one entirely. Added some deprecation tags.

packages/property-grid/src/types.ts
```diff
@@ -27,16 +27,14 @@ export interface OnSelectEventArgs {
   contextMenuArgs: PropertyGridContextMenuArgs;
 }
 
-// Should be used for FunctionalPropertyGridWidget (we get iModelConnection and projectId using hooks)
+// Should be used for BasicPropertyGridWidget (we get iModelConnection and projectId using hooks)
```
aruniverse, line 30:
```suggestion
```

kckst8 (approved):

packages/property-grid/src/components/ElementList.tsx
```diff
@@ -130,93 +125,78 @@ const createDataProvider = async (
...
+  React.useEffect(() => {
+    const createAndSetDp = async () => {
+      const dp = await createDataProvider(labelsProvider, instanceKeys);
+      setDataProvider(dp);
+    };
 
-  public componentWillUnmount() {
-    this._unmounted = true;
-  }
+    createAndSetDp().catch(console.log);
```
jason-crow, line 150: prefer Logger.LogError over console.log

aruniverse: done

PR #140, opened 2021-08-30
PropertyGrid: Fix missing labels and animation bug with MultiElement PG

aruniverse:


kckst8 (approved):
havent tested but code LGTM

calebmshafer (approved):

PR #142, opened 2021-09-16
stop delivering psuedo localized strings, add devcontainer to use with workspaces

aruniverse:
- stop delivering psuedo localized strings
- add devcontainer to use with workspaces


calebmshafer (approved):
LGTM

PR #143, opened 2021-09-28
Styling updates

JoeZman:
Several styling changes:
1. Force header to be a consistent size (50px), this way transitions look better.
2. Show ellipsis for clipped text.
3. Align back button better to the left.
4. Show Element Lists in the header.

### DIFF
--- a/packages/property-grid/src/components/ElementList.tsx
+++ b/packages/property-grid/src/components/ElementList.tsx
@@ -170,6 +170,8 @@ export const ElementList = ({
     [onSelect]
   );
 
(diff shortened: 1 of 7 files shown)


packages/property-grid/src/components/ElementList.tsx
```diff
@@ -170,6 +170,8 @@ export const ElementList = ({
     [onSelect]
   );
 
+  const title = `${PropertyGridManager.translate("element-list.title")} (${instanceKeys.length})`;
```
aruniverse, line 173: Just want to confirm this change here.
Wouldn't this be different from what data-viz and version-compare display? (the result count at the bottom)

aruniverse (approved):

PR #148, opened 2021-10-20
imodel-content-tree: Load InformationContentElement nodes under geometric models

grigasp:


aruniverse:
Grigas who is using this package? Is it just DR?
cc @jason-crow

PR #149, opened 2021-11-12
Export data trees

MeghaCBorty:
Export data trees to use from consuming applications.

### DIFF
--- a/packages/breakdown-trees/package.json
+++ b/packages/breakdown-trees/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bentley/breakdown-trees-react",
-  "version": "0.1.2",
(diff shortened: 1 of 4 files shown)


packages/breakdown-trees/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bentley/breakdown-trees-react",
-  "version": "0.1.2",
+  "version": "0.1.3",
```
aruniverse, line 3: Hey you don't need to manually bump this

MeghaCBorty: Reverted.

aruniverse (approved):

PR #151, opened 2021-11-16
3.0

aruniverse:
Initial pass at updating the vcr repo to iTwin.js 3.0

We need to continue to maintain both 2.x and 3.0 versions of packages, so restructured the directory structure.

We will now have:

|_packages\\
|____archive\\
|____bentley\\
|____itwin\\

Where pkgs under the bentley directory are still on 2.x and need to be upgraded. Need to re-add the 2.x vers of tree-widget, property-grid, and measure-tools in a separate pr. Wanted to keep history between 2.x and 3.0 so just used git mv to track the rename and keep history


packages/archive/imodel-select/src/components/IModelSelector.scss
```diff
@@ -2,87 +2,87 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-
-@import "~@bentley/ui-core/lib/ui-core/index";
```
calebmshafer, line 6: Are any of the changes here real changes or just line endings?

aruniverse: i think line endings, this pkg shouldve been untouched (besides the rename), let me look into it

packages/bentley/breakdown-trees/package.json
```diff
@@ -57,7 +57,7 @@
     "@bentley/ui-ninezone": "^2.19.9",
     "@bentley/webgl-compatibility": "^2.19.9",
     "@bentley/webpack-tools": "1.14.1",
-    "@bentley/measure-tools-react": "workspace:*",
+    "@bentley/measure-tools-react": "^0.5.0",
```
calebmshafer, line 60: I guess this is needed because we don't have the 2.x version in there repo anymore?

aruniverse: correct, will revert this once i add the 2.x ver in a separate pr

packages/bentley/geo-tools/package.json
```diff
@@ -29,7 +29,8 @@
     "rebuild": "npm run clean && npm run build",
     "copy:assets": "cpx \"./src/**/*.{*css,json,svg}\" \"./lib\" && cpx \"./public/**/*\" \"./lib/public\"",
     "cover": "nyc npm test",
-    "lint": "eslint ./src/**/*.{ts,tsx} 1>&2",
+    "lint": "",
+    "lint:real": "eslint ./src/**/*.{ts,tsx} 1>&2",
```
calebmshafer, line 33: Did you not want to run lint in the CI?

aruniverse: i didnt want to touch the bentley pkgs, not sure why they fail to lint, besides the eslint-plugin ver being upgraded

packages/archive/imodel-select/src/components/IModelSelector.scss
calebmshafer: okay cool, I see that with a bunch of the /bentley/ packages as well. Not sure what happened with the line endings... I hate line endings lol  We may need to change this repo to have the same gitattributes as the itwinjs-core repo but better.

packages/bentley/geo-tools/package.json
calebmshafer: hm okay, we can follow up after this PR to see if we can fix it.

packages/itwin/measure-tools/package.json
```diff
@@ -0,0 +1,130 @@
...
+    "@itwin/appui-layout-react": "rc",
+    "@itwin/appui-react": "rc",
+    "@itwin/components-react": "rc",
+    "@itwin/core-bentley": "rc",
+    "@itwin/core-common": "rc",
+    "@itwin/core-frontend": "rc",
+    "@itwin/core-geometry": "rc",
+    "@itwin/core-quantity": "rc",
+    "@itwin/core-react": "rc",
+    "@itwin/core-telemetry": "rc",
```
calebmshafer, line 58: Does telemetry need to be a per dep? We can follow up on this but did want to point it out.

aruniverse: I believe we're just using the type in this PKG so it can be a direct dep. Will update and trim deps in a separate pr

packages/itwin/tree-widget/src/components/trees/TreeWithRuleset.tsx
```diff
@@ -114,30 +113,38 @@ export const ControlledTreeWrapper: React.FC<ControlledTreeProps> = (
...
+  const treeModel = useTreeModel(modelSource);
 
   return (
-    <ControlledTree
-      visibleNodes={visibleNodes}
-      nodeLoader={nodeLoader}
-      treeEvents={unifiedSelectionEventHandler}
-      selectionMode={SelectionMode.Extended}
-    />
+    <div ref={ref} style={{ width: "100%", height: "100%" }}>
```
calebmshafer: Why was this update needed?

calebmshafer: I see this update all over. I assume it's to make the component more adaptive?

aruniverse: one of the breaking changes early on, https://github.com/iTwin/itwinjs-core/blob/master/docs/changehistory/NextVersion.md#controlledtree-api-changes

packages/itwin/property-grid/src/PropertyGridManager.ts
```diff
@@ -3,39 +3,26 @@
...
-  private static _featureFlags: PropertyGridManagerFeatureFlags = {
-    enablePropertyGroupNesting: false,
-  };
   /**
    * Called by IModelApp to initialize PropertyGridManager
    * @param i18n - The internationalization service created by the IModelApp.
    */
-  public static async initialize(
-    i18n: I18N,
-    featureFlags?: PropertyGridManagerFeatureFlags
```
calebmshafer, line 29: Is there no need to have feature flags anymore?

aruniverse: i moved this handling into the props of the property grid itself, im not sure why we ever maintained them separately, really ugly

calebmshafer (approved):

PR #152, opened 2021-11-18
Measure tools 3.0

aruniverse:


packages/measure-tools/src/widgets/PopupToolbar.tsx
```diff
@@ -2,9 +2,10 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
+/* eslint-disable deprecation/deprecation */
```
aruniverse, line 5: Hey @bsteinbk , lets chat about this component later

kckst8 (approved):

PR #153, opened 2021-11-18
Deprecate imodel-select & markup-frontstage

aruniverse:


README.md
```diff
@@ -6,18 +6,23 @@ This repository contains a set of npm packages that deliver React utilities (Com
...
+
+The following packages have been deprecated and will no longer be maintained:
+| Name                               | Folder                                                         | Changelog | Version                                                                                                                                                |
+| ---------------------------------- | -------------------------------------------------------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| @bentley/imodel-select-react       | [packages/imodel-select](./packages/imodel-select/)            | [changelog](./packages/imodel-select/CHANGELOG.md) | [![npm version](https://badge.fury.io/js/%40bentley%2Fimodel-select-react.svg)](https://badge.fury.io/js/%40bentley%2Fimodel-select-react)             |
```
calebmshafer, line 24: Do we want to point to the admin component that should be used instead of this one?

aruniverse: Yea i was wondering if i should do that here in the root readme or in the pkg readme, but then again i dont think with the way i have it setup it won't publish a new version of the pkg. 

I'll update to add a tab for replacements (pointing to admin-comp-react)

aruniverse: Do we actually want to publish a new ver saying this pkg has been deprecated?

calebmshafer (approved):
Let @skirby1996 know if you want us to run the `npm deprecate` script for it.

We'll just need the message you want to send along with it.

PR #159, opened 2021-12-08
tree-widget rc

aruniverse:
bump tree-widget to core rc, try to publish new pre-release under itwin scope


aruniverse:
@calebmshafer lets talk 3.0 vcr pipelines/publishing tomorrow

calebmshafer:
> @calebmshafer lets talk 3.0 vcr pipelines/publishing tomorrow

@aruniverse sounds good. I've been debating if we want to somehow try and have both the 2.x and 3.x versions live side-by-side. That way we don't need to create a release branch in order to release the older versions. What do you think about doing that? The hardest part will probably be to get dependencies to behave correctly.

.github/workflows/publishPreRelease.yml
```diff
@@ -0,0 +1,38 @@
...
+        node-version: ${{ matrix.node-version }}
+    - run: |
+        set -e
+        echo 'Installing...'
+        node common/scripts/install-run-rush.js install
+        echo 'Building...'
+        node common/scripts/install-run-rush.js rebuild
+        echo 'Publishing...'
+        git config user.name "iTwinVCRWorkflow"
+        git config user.email "itwinviewer@users.noreply.github.com"
```
aruniverse, line 34: @calebmshafer is there a preferred git user to use?

calebmshafer: good point, let me try and switch that. It may be the cause of our publishing issues.

calebmshafer: Actually, I wouldn't have a separate workflow for this. I'd try to do it in the same one.

kckst8 (approved):

PR #160, opened 2021-12-09
Fix: info button not appearing in property grid when selecting multiple elements 

cortneysmithbentley:
Problem was that the function that figures out if there's multiple elements selected is registered as a listen event, but that event isn't triggered during the initial opening of the widget (i.e. if you select multiple elements which opens the property grid widget, vs having the widget already opened and selecting multiple elements as a new selection). 

Adding a direct call to the onSelectionChange function after registering the event listener ensures it's still called when the widget is first opened.


jason-crow (approved):

aruniverse (approved):
Ah thx for taking this on; only occurred in ui1.0 so never noticed.

PR #164, opened 2022-01-06
property-grid rc

aruniverse:
done:
- rename pkg to itwin/property-grid-react
- remove deprecated props
- remove PropertyGridManager.featureFlags & refactor how to enable nested property  grid
- remove dead code
- remove sharing functionality temporarily (no way to do shared settings atm in 3.0)


packages/property-grid/README.md
```diff
@@ -1,39 +1,58 @@
...
 
 Copyright © Bentley Systems, Incorporated. All rights reserved.
 
 The property-grid-react package provides React components to access the property grid within an iModel.
 
-## Sample usage
+## Usage
 
-### PropertyGridWidget props
+### What to add in your ITwin AppUI based application
```
kckst8, line 9:
```suggestion
### What to add in your iTwin AppUI based application
```

packages/property-grid/README.md
```diff
@@ -1,39 +1,58 @@
...
+  disableUnifiedSelection?: boolean;
 }
 ```
 
 ### Adding your own options to the property grid context menu
 
-It is possible to provide additional funcitonality to your users via the context menu of the property grid. You can add your own options to this context menu via the `additionalContextMenuOptions` prop which can be passed in through the `applicationData` property on a 9-zone `<Widget>` element which has the `PropertyGridWidgetControl` as its controller.
-
-This `additionalContextMenuOptions` props expects an array filled with `ContextMenuItemInfo` objects, which has the following members:
+It is possible to provide additional funcitonality to your users via the context menu of the property grid. This `additionalContextMenuOptions` props expects an array filled with `ContextMenuItemInfo` objects, which has the following members:
```
kckst8, line 55:
```suggestion
It is possible to provide additional functionality to your users via the context menu of the property grid. This `additionalContextMenuOptions` prop expects an array filled with `ContextMenuItemInfo` objects, which has the following members:
```

packages/property-grid/package.json
```diff
@@ -1,33 +1,41 @@
 {
-  "name": "@bentley/property-grid-react",
-  "version": "2.2.4",
-  "description": "This package offers a React widget which provides access to the property grid within an iModel. Designed for use in an iModeljs 9-zone application",
+  "name": "@itwin/property-grid-react",
+  "version": "0.1.0",
```
kckst8, line 3: should this be a dev version?

packages/property-grid/src/components/ElementList.tsx
```diff
@@ -177,9 +177,11 @@ export const ElementList = ({
       className={classnames("property-grid-react-element-list", rootClassName)}
     >
       <div className="property-grid-react-element-list-header">
+        {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
```
kckst8, line 180: could you add a role instead of disabling? https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md

packages/property-grid/src/components/ElementList.tsx
```diff
@@ -192,6 +194,7 @@ export const ElementList = ({
       </div>
       <div className="property-grid-react-element-list-container">
         {dataProvider && (
+          /* eslint-disable-next-line deprecation/deprecation */
```
kckst8, line 197: future upgrade to iTwin UI I assume?

packages/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -170,31 +170,34 @@ export const MultiElementPropertyGrid = (props: Partial<PropertyGridProps>) => {
...
+              : "translate(-100%,0)",
+          }}
+          enter={{ transform: "translate(0,0)" }}
+          leave={{
+            transform: !animationForward
+              ? "translate(100%,0)"
+              : "translate(-100%,0)",
+          }}
+        >
+          {/* eslint-disable-next-line react/display-name */}
```
kckst8, line 189: should probably disable this globally. Doesnt make a ton of sense for functional components IMO

kckst8 (approved):

packages/property-grid/src/components/ElementList.tsx
aruniverse: yea i was going to tackle that next week if i get time

packages/property-grid/package.json
aruniverse: i think for this pkg an tree-widget, instead of doing pre-releases and its 0.x, just do minor bumps. need to talk this over with caleb before i publish.

packages/property-grid/src/components/MultiElementPropertyGrid.tsx
aruniverse: Idk i think this can be helpful even for functional components. If components don't have display names, theyll show up in dev tools as anonymous. It can happen if you do default unnamed exports (im not a fan of this to begin with, but have seen this in some of our codebases)

kckst8: fair enough. Should we add a name to this component then instead of disabling the rule?

aruniverse: probably should, i just couldnt figure out how to do so with this nested component.
im hoping to nuke this transition wraper tomorrow if i get a chance, i dont love that we pull in an entire library to do what should be a simple animation. also causes us to add a peer dep in the viewer 9react-spring)

kckst8 (approved):

PR #166, opened 2022-01-11
ran manual rush version --bump

aruniverse:
ran manual `rush version --bump` because the ci failed

- locally `rush check` fails even though `ensureConsistentVersions` is set to false :/


calebmshafer (approved):
Yes, this is gross. I'll see if I can spend some time this afternoon figuring out what's wrong with the pipeline.

aruniverse:
I want to see if we can commit some time and pr a fix in rush instead of waiting on them

PR #167, opened 2022-01-12
Applying package updates.

aruniverse:


common/config/azure-pipelines/templates/build.yaml
```diff
@@ -6,10 +6,10 @@ steps:
     inputs:
       versionSpec: "$(NodeVersion).x"
       checkLatest: true
-  - script: 'git config --local user.email 38288322+imodeljs-admin@users.noreply.github.com'
-    displayName: 'git config email'
-  - script: 'git config --local user.name imodeljs-admin'
-    displayName: 'git config name'
+  - script: |
+      git config user.email imodeljs-admin@users.noreply.github.com
```
aruniverse, line 10: just wanted to confirm this is the user that weve allowed to bypass pr reqs in the branch protection rule for master.
and theres nothing additional that needs to be done to "authenticate" as that user

calebmshafer: It should be but that is the part which doesn't seem to be working at the moment for certain repos. The iTwinjs core repo isn't having any trouble with it which has me a bit confused...

calebmshafer (approved):

PR #170, opened 2022-01-13
bump to latest rc, dev-185

aruniverse:
- bump to latest 3.0 rc (dev-185)
- trim deps


packages/itwin/tree-widget/package.json
```diff
@@ -29,69 +29,72 @@
...
-    "lint": "",
+    "lint": "eslint -f visualstudio \"./src/**/*.{ts,tsx}\" 1>&2",
+    "lint:fix": "npm run lint -- --fix",
     "pseudolocalize": "betools pseudolocalize --englishDir ./public/locales/en --out ./lib/public/locales/en-PSEUDO",
     "test": "",
     "clean": "rimraf lib .rush/temp/package-deps*.json",
     "rebuild": "npm run clean && npm run build"
   },
   "dependencies": {
+    "@itwin/core-i18n": "rc",
```
aruniverse, line 40: @calebmshafer could you review the changes in here? im a little iffy on the appui-layout-react removal.

packages/itwin/property-grid/package.json
```diff
@@ -38,13 +38,14 @@
     "rebuild": "npm run clean && npm run build"
   },
   "dependencies": {
+    "@itwin/core-i18n": "rc",
```
kckst8, line 41: should these be peerDeps?

aruniverse: after writing that, makes me question, if we could just get away with making it a dev dep since its only used for types

kckst8: i was thinking that as well....maybe test that? if it builds, should be fine

kckst8 (approved):

packages/itwin/property-grid/package.json
aruniverse: this seemed to work. any pkgs were just using as types i kept as a dev dep & removed as a peer dep.
also went through and removed any peers of peers as that should be unnecessary

packages/itwin/measure-tools/package.json
```diff
@@ -46,21 +47,18 @@
   ],
   "peerDependencies": {
     "@itwin/appui-abstract": "rc",
-    "@itwin/appui-layout-react": "rc",
```
calebmshafer, line 49: Since this is used to import "Toolbar" it needs to be either a real or peerDependency and given the static nature of pieces of the package I think it'll need to be a peer.

packages/itwin/tree-widget/src/components/search-bar/SearchBox.tsx
```diff
@@ -38,8 +38,6 @@ export interface SearchBoxProps extends CommonProps {
   resultCount: number;
   /** Callback to currently selected result/entry change */
   onSelectedChanged: (index: number) => void;
-  /** Specify that the <input> element should automatically get focus */
```
calebmshafer, line 41: Is this not a needed prop?

aruniverse: this component was only used internally, and we never used the prop. 

also, there was a linter warning re ux, and how we should never set autofocus directly so just removed it altogether

calebmshafer: okay, sounds good to me.

calebmshafer (approved):

PR #172, opened 2022-01-19
Attempt to use --local to persist the git email and user

calebmshafer:


aruniverse (approved):
could you also delete the duplicate rush change -verify we do in here? preferably the one after the install

calebmshafer:
> could you also delete the duplicate rush change -verify we do in here? preferably the one after the install

@aruniverse , yup, done.

PR #176, opened 2022-02-01
update scss to be pulled from cjs dir

aruniverse:
Needed for when you link this repo with the core repo


calebmshafer (approved):
That's unfortunate :(

PR #181, opened 2022-02-04
Fix so that the ZoneLocation is checked for UI 1.0

WillGrayMSU:
Currently there is an issue for this component in UI 1.0 where the Widgets won't show up.  This is because the ZoneLocation is not being checked by provideWidgets.


aruniverse (commented):
Hey Will, run `rush change` then select `patch`

PR #182, opened 2022-02-04
Fix for GroupingMatchingWidget when used in UI 1.0 applications

WillGrayMSU:
The Grouping Mapping Widget is not working in UI 1.0 contexts because of a defect where it is not checking for the zoneLocation and so does not return the widget to UI 1.0 when provideWidgets is called.

This PR adds a check for AbstractZoneLocation.CenterLeft (4) so that the widget will be returned when called by UiItemsProviders in UI 1.0 contexts.  This does not affect UI 2.0 uses.

### DIFF
--- a/common/changes/@itwin/grouping-mapping-widget/Fix-for-grouping-mapping-Ui-1.0_2022-02-04-22-07.json
+++ b/common/changes/@itwin/grouping-mapping-widget/Fix-for-grouping-mapping-Ui-1.0_2022-02-04-22-07.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/grouping-mapping-widget",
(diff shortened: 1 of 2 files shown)


common/changes/@itwin/grouping-mapping-widget/Fix-for-grouping-mapping-Ui-1.0_2022-02-04-22-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "This PR adds a check for AbstractZoneLocation.CenterLeft (4) so that the widget will be returned when called by UiItemsProviders in UI 1.0 contexts. This does not affect UI 2.0 usAdds es.",
```
aruniverse, line 5: not sure if something got cropped here

aruniverse (approved):

Guillar1 (approved):

PR #187, opened 2022-02-10
[Grouping&Mapping] Improved error handling

Guillar1:
Improvement to error handling and small tweaks to import mapping wizard

### DIFF
--- a/common/changes/@itwin/grouping-mapping-widget/property-menu-overflow-fix_2022-02-10-23-02.json
+++ b/common/changes/@itwin/grouping-mapping-widget/property-menu-overflow-fix_2022-02-10-23-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "Fixed property menu overflow",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/grouping-mapping-widget"
+}
--- a/common/changes/@itwin/grouping-mapping-widget/struct-fix_2022-02-12-01-09.json
+++ b/common/changes/@itwin/grouping-mapping-widget/struct-fix_2022-02-12-01-09.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "Fixed missing struct properties in group properties",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/grouping-mapping-widget"
+}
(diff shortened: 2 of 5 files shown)


packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -248,27 +259,94 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
+              );
+              hilitedElements.current.set(row.query ?? "", hiliteIds);
 
-            allIds = allIds.concat(ids);
+              allIds = allIds.concat(ids);
+            } catch {
+              const index = groups.findIndex((group) => group.id === row.id);
+              setSelectedRows((rowIds) => {
+                const selectedRowIds = { ...rowIds };
+                delete selectedRowIds[index];
```
aruniverse, line 279: might be me, but i dont like the use of `delete` here, couldnt you just use Array.prototype.filter?

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -248,27 +259,94 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
+  ): TableState => {
+    switch (action.type) {
+      case actions.toggleRowSelected: {
+        const newSelectedRows = {
+          ...selectedRows,
+        };
+        if (action.value) {
+          newSelectedRows[action.id] = true;
+        } else {
+          delete newSelectedRows[action.id];
```
aruniverse, line 324: another one

Guillar1: Sure, I was just reimplementing the select logic from react-table itself. 
https://github.com/TanStack/react-table/blob/76a4a861ee56b782404ef91987c3b5691ecf2ebc/src/plugin-hooks/useRowSelect.js#L161

aruniverse: I might have to fight Tanner :P

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
Guillar1: I looked into again. I feel like this will be a pain since the selectedRowIds aren't arrays. I just remembered that react-table uses a Record with the properties being the row id's.

aruniverse (approved):

PR #197, opened 2022-02-22
update react & react-dom ver for geo-tools

aruniverse:


packages/itwin/geo-tools/src/components/GeoAddressSearch.tsx
```diff
@@ -5,14 +5,15 @@
 
 import "./GeoAddressSearch.scss";
 
-import type { AutoSuggestData, CommonProps} from "@itwin/core-react";
+import type { AutoSuggestData, CommonProps } from "@itwin/core-react";
 import { AutoSuggest, WebFontIcon } from "@itwin/core-react";
-import type { AddressProvider} from "../AddressProvider";
+import type { AddressProvider } from "../AddressProvider";
```
mdastous-bentley, line 10: @aruniverse strange I don't catch those syntax errors, do you have any special VSCode module/configuration?

aruniverse: The lint command should've caught this. But I do have prettier vscode extension enabled so that might be it too. Its annoying to work in core cause of it tho since they have different rules.

jason-crow (approved):

PR #198, opened 2022-02-25
[imodel-react-hooks]: update to 3.0

aruniverse:
Tested the marker hook in a dummy app and that worked.
Not sure if anyone is using the useFeatureOverrides hook; commented out tests for now. Someone will fix it when it becomes an issue.


packages/itwin/imodel-react-hooks/package.json
```diff
@@ -50,14 +54,15 @@
...
+    "@itwin/build-tools": "^3.0.0",
+    "@itwin/core-bentley": "^3.0.0",
+    "@itwin/core-geometry": "^3.0.0",
+    "@itwin/core-common": "^3.0.0",
+    "@itwin/core-frontend": "^3.0.0",
     "@testing-library/react": "^9.1.4",
     "@types/jest": "^24.9.0",
-    "@types/react": "~16.9.35",
-    "@types/react-dom": "^16.8.0",
+    "@types/react": "^17.0.19",
```
jason-crow, line 64: should this be uniform with core's ui packages which are using "^17.0.0" ?

packages/itwin/imodel-react-hooks/src/FeatureOverrides/useFeatureOverrides.tsx
```diff
@@ -113,23 +108,25 @@ export const FeatureOverrideReactProvider = ({
...
 
   useEffect(() => {
-    const attach = () =>
-      IModelApp.viewManager.forEachViewport((vp) => {
+    const attach = () => {
+      for (const vp of IModelApp.viewManager) {
         if (!viewFilter || viewFilter(vp)) {
-          vp.featureOverrideProvider = impl;
+          // to do
+          // vp.featureOverrideProviders = impl;
```
aruniverse, line 115: need to fix this,

jason-crow: think youll need to do this
vp.addFeatureOverrideProvider

aruniverse: thanks!

packages/itwin/imodel-react-hooks/package.json
aruniverse: its just types so should be ok; theyll probably resolve to the same dep because of the ^

packages/itwin/imodel-react-hooks/src/FeatureOverrides/useFeatureOverrides.tsx
```diff
@@ -4,13 +4,8 @@
...
 
 
-import { Logger } from "@bentley/bentleyjs-core";
-import {
-  FeatureOverrideProvider,
-  FeatureSymbology,
-  IModelApp,
-  Viewport,
-} from "@bentley/imodeljs-frontend";
+import { Logger } from "@itwin/core-bentley";
```
aruniverse, line 7: @MichaelBelousov I tested the useMarker hook, could you test useFeatureOverride ?

jason-crow (approved):

packages/itwin/imodel-react-hooks/package.json
```diff
@@ -50,30 +54,31 @@
...
-    "@bentley/geometry-core": "^2.0.0",
-    "@bentley/imodeljs-common": "^2.0.0",
-    "@bentley/imodeljs-frontend": "^2.0.0",
-    "react": "^16.13.0",
-    "react-dom": "^16.13.0"
+    "@itwin/core-bentley": "^3.0.0",
+    "@itwin/core-geometry": "^3.0.0",
+    "@itwin/core-common": "^3.0.0",
+    "@itwin/core-frontend": "^3.0.0",
+    "react": "^17.0.2",
```
MichaelBelousov, line 82: won't this force consumers to need react 17? If this package wasn't changed to break support for react 16, why not: `^16.13.0||^17.0.2`?

MichaelBelousov (changes requested):

packages/itwin/imodel-react-hooks/package.json
```diff
@@ -1,34 +1,38 @@
...
     "hooks",
     "react",
-    "iModel.js"
+    "iModel",
+    "iTwin.js"
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/imodeljs/viewer-components-react/tree/master/packages/imodel-react-hooks"
+    "url": "https://github.com/imodeljs/viewer-components-react/tree/master/packages/itwin/imodel-react-hooks"
```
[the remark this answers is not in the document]
aruniverse: @calebmshafer  @skirby1996 thoughts on pkg name? itwin/imodel-react-hooks?

packages/itwin/imodel-react-hooks/src/FeatureOverrides/useFeatureOverrides.tsx
aruniverse: yea bad practice, but neither of us have the time atm to spend too much time on it. im thinking since we're releasing this pkg as 0.x, if it breaks its not the end of the world

packages/itwin/imodel-react-hooks/package.json
aruniverse: in 3.0 we decided to move all core and related pkgs to 17; no real reason anyone should stick on 16; for the most part its backwards compatible so users shouldnt have any difficulty updgrading

MichaelBelousov: I'm still kind of against forcing an upgrade although i did hear it's backwards compatible. But if core did that then go ahead, resolve as you wish

aruniverse: we (core) didnt want to continue maintaining 2 majors of react; much easier for us to force users to move up with us

packages/itwin/imodel-react-hooks/src/FeatureOverrides/useFeatureOverrides.tsx
aruniverse: i was trying to resolve some test failures, im just going to skip the ones that fail for now. will need anthony to report on the feature override hook later

packages/itwin/imodel-react-hooks/package.json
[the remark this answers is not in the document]
calebmshafer: `imodel-react-hooks` seems fine with me. It'd be a bit awkward to use our standard pattern of putting `-react` as the suffix since the hooks are react specific. Plus it reads weird as `@itwin/imodel-hooks-react`.

As for whether to drop the imodel, I have no problem with that if we are going to expand upon the list of hooks. Do they cover anything beyound Marker and FeatureOverrides at the moment?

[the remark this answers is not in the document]
aruniverse: theyre hook based wrappers over functionality from core-frontend.

we do have the [bentley/react-hooks](https://github.com/iTwin/react-hooks) pkg too for more generic hooks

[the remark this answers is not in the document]
MichaelBelousov: There is definitely room for more hooks and react wrapping. I'd hope someone could contribute if they have or want them. That is the intended scope of this package.

MichaelBelousov (approved):

packages/itwin/imodel-react-hooks/package.json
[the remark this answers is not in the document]
aruniverse: itwin/imodel-react-hooks is what we're going to go with; resolving

PR #203, opened 2022-03-04
Property grid animation

johnnyd710:
[itwin/property-grid-react](https://github.com/iTwin/viewer-components-react/tree/master/packages/itwin/property-grid) pulls in an entire package, namely [react-spring](https://www.npmjs.com/package/react-spring) to do a "simple" animation. The package was removed and replaced by a CSS transition animation.


packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -112,41 +105,34 @@ export const MultiElementPropertyGrid = (props: PropertyGridProps) => {
...
       }}
       key={"SingleElementPropertyGrid"}
     />
   );
 
+  // since css animation requires all react components be rendered - just move non-active components to the side
+  const getAnimationStyle = (idx: MultiElementPropertyContent.PropertyGrid) => {
+    let style: React.CSSProperties;
+
+    if (idx === content) {
```
aruniverse, line 117: Do we need this check?

aruniverse (commented):
Haven't tested but changes look fine. @JoeZman can you review?

kckst8:
> Since you made the pr from the cospace, the VCR repos lockfile didn't get updated. You'll need to open this branch in a copy of VCR not in a cospace, do `rush install` and commit the changes.

We should try to simplify this. I ended up just linking my only copy of all repos in my cospace and I open each individually and use the rush workflow before committing/pushing.

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
johnnyd710: Just tested it, we don't need the check if we initialize style to an empty object. Having an empty object for style has the same visual effect as translate(0,0). I'll push a new commit

aruniverse:
> We should try to simplify this. I ended up just linking my only copy of all repos in my cospace and I open each individually and use the rush workflow before committing/pushing.

Yea this is probably my only complaint with cospaces, and ive been thinking about how to address this but idk if there is a good solution for this.

I was thinking we could add a root level package.json to all the monorepos and just add a dep on rush or yarn or whatever you use to manage your deps in there, then adding that pkg in the cospaces pnpm-workspace.yaml. Then i guess theoretically you could cd into the dir of the repo youve linked and run the commands as you would typically. But the issues with that is you shouldn't mix & match your npm commands because the package managers will get confused, [see](https://rushjs.io/pages/developer/new_developer/#1-avoid-certain-commands-in-a-rush-repo).

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -112,41 +105,32 @@ export const MultiElementPropertyGrid = (props: PropertyGridProps) => {
...
       onBackButton={() => {
         setContent(MultiElementPropertyContent.ElementList);
-        setAnimationForward(false);
       }}
       key={"SingleElementPropertyGrid"}
     />
   );
 
+  // since css animation requires all react components be rendered - just move non-active components to the side
+  const getAnimationStyle = (idx: MultiElementPropertyContent.PropertyGrid) => {
```
aruniverse, line 114: i dont think the type for this should be of `MultiElementPropertyContent.PropertyGrid`. can probably just do number

aruniverse:
> We should try to simplify this. I ended up just linking my only copy of all repos in my cospace and I open each individually and use the rush workflow before committing/pushing.

ANother option, is just running `cospace purge` and then going into the monorepo that you're working in and do your regular workflow. Kinda annoying because youre always deleting the symlinks and need to regen them later. And i believe you could still run `cospace purge` to nuke all the node_modules installed by your other package manager, as long as the pnpm-workspace.yaml file is defined with **all** the pkgs in that monorepo

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.scss
```diff
@@ -22,10 +22,18 @@
...
-  transition: transform 500ms;
+  transition: transform 400ms ease-out;
   position: absolute;
   top: 0;
   left: 0;
   height: 100%;
   width: 100%;
 }
+
+.animate-right {
```
JoeZman, line 33: Can you give animate-left and animate-right a more specific name like property-grid-animate-left/right, or nest it:
.property-grid-react-animated-tab {
  &.animate-right {}
  &.animate-left {}
}

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.scss
```diff
@@ -22,10 +22,18 @@
 }
 
 .property-grid-react-animated-tab {
```
JoeZman, line 24: Is there a need to include "react" in the class name?
property-grid-animated-tab

johnnyd710: I think its because the package is called "property-grid-react" so they may just be following that convention but I'll wait for others to comment!

JoeZman (approved):

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.scss
aruniverse: yea its because the pkg name is property-grid-*react*, and i think we wanted to make sure we didnt have too generic classnames that cuase collisions in other apps. Myabe we need a separate pr to review all the css?

aruniverse (approved):
You just need to merge with master, resolve the conflict by running `rush updat`, commit those changes and the pr should be good to go

PR #211, opened 2022-03-16
[imodel-react-hooks] fix hook deps list

aruniverse:


packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
```diff
@@ -25,19 +25,19 @@ export function useOnMountInRenderOrder(effect: () => void | (() => void)) {
 
 /** run code on component mount */
 export function useOnMount(impl: () => void) {
-  useEffect(() => void impl(), []);
+  useEffect(() => void impl(), [impl]);
```
MichaelBelousov, line 28: this actually should not take a dependency. It's intended to not run on additional effect runs, just the first one (i.e. mount). So this way the user passing a callback will not be surprised that it changes. Otherwise we would need to expose the dependencies to them.

An alternative explicit way of doing it is capturing the first `impl` value with either `useRef` or `useState` and using that as the dependency, since we know it won't change without us changing it

packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
```diff
@@ -25,19 +25,19 @@ export function useOnMountInRenderOrder(effect: () => void | (() => void)) {
...
 }
 
 /** run something on component unmount */
 export function useOnUnmount(impl: () => void) {
-  useEffect(() => () => impl(), []);
+  useEffect(() => () => impl(), [impl]);
```
MichaelBelousov, line 33: same as issue with `useOnMount`

packages/itwin/imodel-react-hooks/src/utils/react-context.ts
```diff
@@ -1,18 +1,15 @@
...
 import React from "react";
 
 /**
  * like React.createContext, but the default state (the one used when no provider is found by a consumer),
  * throws errors on usage, useful for when consuming without a provider should be considered programmer error.
  */
-export const makeContextWithProviderRequired = <T extends any>(
-  name = "UntitledContext"
-) =>
+export const makeContextWithProviderRequired = <T>(name = "UntitledContext") =>
```
MichaelBelousov, line 12: this syntax works now? I thought you couldn't make generic arrow functions without an `extends` clause because of ambiguity with the old cast operator in typescript? `<any>5`

packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
```diff
@@ -25,19 +25,19 @@ export function useOnMountInRenderOrder(effect: () => void | (() => void)) {
...
 }
 
 /** create a stable object reference on intialization */
 export function useStable<T>(make: () => T) {
   // can't use memo since it's only for expensive computations, react will (eventually)
   // selectively recreate the object breaking the contract.
-  const value = useMemo(make, []);
+  const value = useMemo(make, [make]);
```
MichaelBelousov, line 40: probably can replace `useMemo` with another ref and guarantee one call to the expensive function without any dependencies, like so:
```tsx
const valueRef = useRef<T>(undefined as any);
const hasMade = useRef(false);
if (!hasMade) {
  valueRef.current = make();
  hasMade.current = true;
}
return valueRef.current;
```

packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
aruniverse: reverted

packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
aruniverse: reverted

packages/itwin/imodel-react-hooks/src/utils/basic-hooks.ts
MichaelBelousov: oh yeah. Figures considering I didn't actually try my code out

packages/itwin/imodel-react-hooks/src/utils/react-context.ts
MichaelBelousov: I don't think they deprecated the old syntax. I'm not sure how the ambiguity can be resolved.
reference: https://stackoverflow.com/questions/32308370/what-is-the-syntax-for-typescript-arrow-functions-with-generics#45576880
maybe do `T extends object`? idk...

MichaelBelousov (approved):

packages/itwin/imodel-react-hooks/src/utils/react-context.ts
MichaelBelousov: I do remember the linter rules recommend extending `Record<string, any>` (or with `unknown`), but I guess I wouldn't want the string keys restriction. I would prefer just disabling the lint rule for this line and using `any`. The link I posted above recommends using a hanging comma to disambiguate the syntax.
```tsx
const makeContextWithProviderRequired = <T, >(...args) => {};
```
I would probably vote for disabling the linter rule since I haven't seen the hanging-comma hack before

aruniverse: would you rather i just rever the changes to this file?

PR #216, opened 2022-03-24
Add ability for user to supply default location for Tree Widget

bsteinbk:
Design Review would like the ability to set the default location to the left panel.

### DIFF
--- a/packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
+++ b/packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
@@ -3,7 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 /* eslint-disable react/display-name */
-import type {
+import {
   AbstractWidgetProps,
   StagePanelSection,
   UiItemsProvider,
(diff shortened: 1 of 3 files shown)


packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -3,7 +3,7 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 /* eslint-disable react/display-name */
-import type {
+import {
```
aruniverse, line 6: i think youll have to run `lint:fix`

packages/itwin/tree-widget/src/types.ts
```diff
@@ -16,31 +17,33 @@ export interface IModelContentTreeProps
...
 
 export type SpatialTreeProps = Omit<
-SpatialContainmentTreeProps,
-"iModel" | "width" | "height"
+  SpatialContainmentTreeProps,
+  "iModel" | "width" | "height"
 >;
 
 export interface TreeWidgetOptions {
+  defaultPanelLocation?: StagePanelLocation,
```
aruniverse, line 45: im surprised it took this long for a request like this to come through...

should we instead just expose an obj for users to set their preferred AbstractWidgetProps ?

bsteinbk: These locations are not defined in AbstractWidgetProps, but allowing those to be set may also be a good addition.

bsteinbk: @aruniverse - since you own this repo I will let you resolve this if you are ok with response.

aruniverse (approved):

PR #217, opened 2022-03-25
Hide property grid by default

johnnyd710:
Hides property grid by default unless at least one non-transient element is in selection set.

### DIFF
--- a/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
+++ b/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
@@ -15,7 +15,7 @@ import {
 import { UiFramework } from "@itwin/appui-react";
 import * as React from "react";
 
-import { MultiElementPropertyGrid } from "./components/MultiElementPropertyGrid";
+import { MultiElementPropertyGrid, MultiElementPropertyGridId } from "./components/MultiElementPropertyGrid";
 import { PropertyGridManager } from "./PropertyGridManager";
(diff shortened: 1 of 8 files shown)


packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -16,13 +17,22 @@ import { ElementList } from "./ElementList";
...
   ElementList = 1,
   SingleElementPropertyGrid = 2,
 }
 
+function useSpecificWidgetDef(id: string) {
+  const frontstageDef = useActiveFrontstageDef();
+  return frontstageDef?.findWidgetDef(id);
+}
+
+export const MultiElementPropertyGridId = "vcr:PropertyGrid";
```
aruniverse, line 34:
```suggestion
export const MultiElementPropertyGridId = "vcr:MultiElementPropertyGrid";
```

aruniverse (approved):
lgtm

packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -53,10 +53,10 @@ export class PropertyGridUiItemsProvider implements UiItemsProvider {
       )
     ) {
       widgets.push({
```
bsteinbk, line 55: I would like to see  PropertyGridProps expanded to allow the default location to be specified like I am adding in https://github.com/iTwin/viewer-components-react/pull/216.

PR #218, opened 2022-03-25
[Grouping&Mapping] Group by spinner fix and package updates

Guillar1:
Group by property spinners were triggered on rerender. Added extra check. Also did some package updates and got rid of itwinui-css dependency (it's now actually installing 🥴).


bingjiez (approved):

packages/itwin/grouping-mapping-widget/package.json
```diff
@@ -62,26 +62,25 @@
...
-    "@itwin/core-orbitgt": "^3.0.0",
-    "@itwin/core-quantity": "^3.0.0",
-    "@itwin/core-react": "^3.0.0",
-    "@itwin/eslint-plugin": "^3.0.0",
-    "@itwin/itwinui-css": "^0.50.0",
-    "@itwin/presentation-common": "^3.0.0",
-    "@itwin/presentation-components": "^3.0.0",
-    "@itwin/presentation-frontend": "^3.0.0",
-    "@itwin/webgl-compatibility": "^3.0.0",
+    "@itwin/appui-abstract": "^3.1.0",
```
aruniverse, line 65: Did you need to bump all your deps to 3.1?

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.scss
```diff
@@ -10,6 +10,11 @@
   width: 100%;
 }
 
+.iui-modal > .iui-modal-dialog {
```
aruniverse, line 13: This will probably clash for other consumers who are might be using the same class name.

aruniverse (commented):
More concerned about updating the core deps. If it's not required, I'd revert that

packages/itwin/grouping-mapping-widget/package.json
Guillar1: Yes, appui-react 3.1.0 has the latest itwinui-react dependency. Also doing rush update -full is risky and breaks the other widgets in this repo. Unless there is a better way of doing this.

Guillar1: I just noticed your other comment. Is that why the other widgets break when I do rush update -full? I can undo this if that is the case and wait for https://github.com/iTwin/itwinjs-core/pull/3446

aruniverse: Yea thats why other pkgs broke, i'd revert your commit and wait for 3.1.1

aruniverse (approved):

PR #180, opened 2022-02-04
[Carbon Calculation] One Click LCA UI Widget

bingjiez:
Viewer widget to trigger one-click-LCA carbon calculation jobs. The API behind is provided by reporting platform and calculations are based on model insight reports.

UI mainly includes a table of reports, job trigger button, one-click-LCA login UI, job status report and result (link to report/fail message).

PR #219, opened 2022-04-01
rush update, update rush & pnpm

aruniverse:
- Switch pipelines and rush config to use Node >14.9 as 12 reaches EOL at the end of the month
- Tests started to fail in 3.1 due to adding a symbol for core-frontend:
  - we need to list all transitive peers as dev deps so all the pkgs correctly reoslve, and theres only ever one copy of each module


calebmshafer (approved):

packages/itwin/breakdown-trees/package.json
```diff
@@ -41,81 +41,86 @@
...
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.6.3",
     "chai": "^4.2.0",
-    "nyc": "^15.1.0",
     "chai-as-promised": "^7.1.1",
     "chai-jest-snapshot": "^2.0.0",
     "chai-spies": "1.0.0",
-    "sinon": "^7.5.0",
-    "sinon-chai": "^3.2.0",
-    "faker": "^4.1.0",
+    "cpx": "^1.5.0",
```
a-gagnon, line 85: Should not use cpx.
Replace by "cpx2": "^3.0.0",

aruniverse: snuck that into https://github.com/iTwin/viewer-components-react/pull/180

PR #223, opened 2022-04-12
Scss fix Tree Search Bar Width when opened

WillGrayMSU:
The search bars for the breakdown, model, and categories trees were only taking up half the width of the widget in Design Review.  This fix sets the 'right' CSS property of search-bar-search-container to 0 so that the search bar spans the width of the widget when opened.


aruniverse (approved):

aruniverse:
please have @JoeZman review

JoeZman (approved):

PR #226, opened 2022-04-15
[breakdown-trees] Create section fix

DanishMehmood-bit:


packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -10,100 +10,86 @@ export class DataLink {
   public static async queryAllBuildings(iModel: IModelConnection) {
```
JuliusArmalis, line 10: > public static async queryAllBuildings

If you don't await anything, than you don't need to write async. Great job with the refactoring!

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -10,100 +10,86 @@ export class DataLink {
...
-        return new Range3d(rows[0].minX, rows[0].minY, rows[0].minZ, rows[0].maxX, rows[0].maxY, rows[0].maxZ);
-    } else {
-      queryGetStoreyProps = queryGetStoryWithElementsProps;
-      const rows = await this.executeQuery(iModel, queryGetStoreyProps);
-      if (rows.length > 0 && rows[0].minX)
-        return new Range3d(rows[0].minX, rows[0].minY, rows[0].minZ, rows[0].maxX, rows[0].maxY, rows[0].maxZ);
-    }
-    return undefined;
+
+    return clipAtSpaces ? runQuery(queryGetStoreyProps) ?? runQuery(queryGetStoryWithElementsProps) : runQuery(queryGetStoryWithElementsProps);
```
JuliusArmalis, line 57: > return clipAtSpaces ? runQuery(queryGetStoreyProps) ?? runQuery(queryGetStoryWithElementsProps) : runQuery(queryGetStoryWithElementsProps);

The last 2 method calls are identical, we should be able to combine them.
Maybe something like `return clipAtSpaces && runQuery(queryGetStoreyProps) || runQuery(queryGetStoryWithElementsProps);` would be even better?

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -10,100 +10,86 @@ export class DataLink {
...
   }
 
   public static async queryRooms(iModel: IModelConnection, storyId?: string) {
-    let query = "";
-    query = `SELECT room.ECInstanceId as id, room.UserLabel as label, room.CodeValue as code, room.BBoxHigh as bBoxHigh, room.BBoxLow as bBoxLow, story.ECInstanceId as composingId
+    const queryWithoutStoryId = `SELECT room.ECInstanceId as id, room.UserLabel as label, room.CodeValue as code, room.BBoxHigh as bBoxHigh, room.BBoxLow as bBoxLow, story.ECInstanceId as composingId
     FROM BuildingSpatial:Space room
     JOIN Spatial.Story story ON story.ECInstanceId = room.composingElement.id`;
+    const queryWithStoryId = `SELECT room.ECInstanceId as id, room.UserLabel as label, room.CodeValue as code, room.BBoxHigh as bBoxHigh, room.BBoxLow as bBoxLow, story.ECInstanceId as composingId
+    FROM BuildingSpatial:Space room
```
JuliusArmalis, line 65: As I understand only the JOIN part is different between these queries. Maybe you can DRY the other parts?

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
DanishMehmood-bit: Lint is giving error if I don't async

DanishMehmood-bit: You can see in my first commit I didn't added async but was getting lint errors so I added them back

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
DanishMehmood-bit: fixed

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
DanishMehmood-bit: fixed

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -10,100 +10,79 @@ export class DataLink {
...
+    const constructQuery = (queryClass: string) => {
+      return `select MIN(spel.minX) as minX, MIN(spel.minY) as minY, MIN(spel.minZ) as minZ, MAX(spel.maxX) as maxX, MAX(spel.maxY) as maxY, MAX(spel.maxZ) as maxZ
+        from ${queryClass} ovr join biscore.spatialIndex spel on spel.ECinstanceId = ovr.targetecinstanceid
+        where sourceECInstanceId in
+        (select sp.ecinstanceid from buildingspatial.story s join buildingspatial.space sp on s.ecinstanceid=sp.composingelement.id where s.ecinstanceid=${storyId}
+        union
+        select s.ecinstanceid from buildingspatial.story s where s.ecinstanceid=${storyId})`;
+    };
+
+    return clipAtSpaces && runQuery(constructQuery("spatialcomposition.CompositeComposesSubComposites")) || runQuery(constructQuery("spatialcomposition.CompositeOverlapsSpatialElements"));
```
JValiunas, line 52: this may be more compact, but has poor readability

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
```diff
@@ -10,100 +10,79 @@ export class DataLink {
...
   public static async queryStoriesWithRooms(iModel: IModelConnection) {
     const query = `SELECT DISTINCT story.ECInstanceId as id, story.UserLabel as label, story.CodeValue as code
     FROM BuildingSpatial:Story story, BuildingSpatial:Space room WHERE room.composingElement.id = story.ECInstanceId`;
-    const rows = await this.executeQuery(iModel, query);
-    return rows;
+
+    return this.executeQuery(iModel, query);
   }
 
   public static async queryStoryRange(iModel: IModelConnection, storyId: string, clipAtSpaces: boolean): Promise<Range3d | undefined> {
```
JValiunas, line 34: there are significant changes in this function, but we seem to be only testing that "it calls query() and it returns whatever it gets"

could you add a test for "storyid" and "clipatspaces" parameters too?

JValiunas (changes requested):

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
DanishMehmood-bit: Yes I think we talked about this as team and concluded that we'll try to DRY as much as possible. @JuliusArmalis any thoughts?

JValiunas: DRY goes for "Don't repeat yourself" not squeeze everything into a single line.

i'd make it 2 lines:

const query = clipAtSpaces ? constructQuery("spatialcomposition.CompositeComposesSubComposites") : constructQuery("spatialcomposition.CompositeOverlapsSpatialElements");

return runQuery(query);

//or something similar

DanishMehmood-bit: This wouldn't work since if clipAtSpaces is true then we need to run the query with CompositeComposesSubComposites but if that query returns undefined then we need to run the query again with CompositeOverlapsSpatialElements. I could change it as following

const compositeComposesQuery = constructQuery("spatialcomposition.CompositeComposesSubComposites");
const compositeOverlapsQuery = constructQuery("spatialcomposition.CompositeOverlapsSpatialElements");

return clipAtSpaces && runQuery(compositeComposesQuery) || runQuery(compositeOverlapsQuery);

JuliusArmalis: I think that using named variables to increase readability is a good option here.

JValiunas: i think that statement "if that query returns undefined then we need to run the query again with CompositeOverlapsSpatialElements" is why this needs to be organized better. For me I missed that detail as I was reviewing, and good code should not need documentation.

I think something like this would be more readable? :

let result = undefined;

if (clipAtSpaces)
	{
	result = runQuery(constructQuery("spatialcomposition.CompositeComposesSubComposites"));
	}

return result ?? runQuery(constructQuery("spatialcomposition.CompositeOverlapsSpatialElements"));

DanishMehmood-bit: I would totally agree on this one because we should prioritize readability

DanishMehmood-bit: Fixed

packages/itwin/breakdown-trees/src/Views/visibility/DataLink.ts
DanishMehmood-bit: Added tests

JValiunas: can we also have a test to capture the other point? "if that query returns undefined then we need to run the query again with CompositeOverlapsSpatialElements"

also, i think it would be better not to have a constructquery function in the test, but just hardcode the expected string.

DanishMehmood-bit: Added new test to capture the mentioned scenario + refractor

JValiunas (approved):

PR #230, opened 2022-04-21
[Tree-Widget] Fix AutoSizer to render children even if width or height is 0

saskliutas:
Changed AutoSizer behavior to avoid unmounting children when width or height is 0. This was causing trees from `@itwin/appui-react` to lose their state.


JoeZman (approved):

aruniverse:
I need to test this out, I think the reason I added this check was because the props were required by the trees and couldn't be falsey

aruniverse:
tested this out locally and its fine; im not sure why i had the falsey check earlier anymore

aruniverse (approved):

PR #232, opened 2022-04-25
Bugfix/measure tools radius aggregate

a-gagnon:
#### Bumped the version number to 0.8.0.


packages/itwin/measure-tools/public/locales/en/MeasureTools.json
```diff
@@ -82,11 +82,13 @@
       "unhandledGeoLocationError": "Failed to convert coordinates to GeoLocation"
     },
     "MeasureRadius": {
-      "measureRadius": "Radius",
+      "keyin": "Radius",
+      "flyover": "Radius",
+      "description": "Create radius measurements.",
+      "longLabel": "Measure Radius",
```
a-gagnon, line 88: The `longLabel` is only used externally (for now) and is meant to replace the tool name whenever it is not obvious we are measuring something.
Most apps will probably have all these tools under a common `Measure` group but that's not always the case.

I'm thinking we could refactor the various `MeasureToolDefinitions` getters into functions and have the caller decide whether he wants the 'short' name (eg. `Distance`) or the 'long' name (eg. `Measure Distance`) rather than having the consuming package request a localized string from the measure-tools package.

bsy-nicholasw (approved):

PR #235, opened 2022-04-29
[Grouping&Mapping] Add custom calculation formula validation

lzaromskis:
Adds custom calculation formula validation in the UI. Now the formula is being validated client side, so the user should be getting specific error messages describing what is wrong with the formula


JValiunas:
because you're adding the validation into the package itself, I have to ask.. is the validation possible via REST API?
assuming we have an API available for other parties to use, wouldn't they need a similar check? and then if they do, do we really need to have it client side too (as opposed to making an API call)?

lzaromskis:
> because you're adding the validation into the package itself, I have to ask.. is the validation possible via REST API? assuming we have an API available for other parties to use, wouldn't they need a similar check? and then if they do, do we really need to have it client side too (as opposed to making an API call)?

The API returns a generic error that the formula is wrong and does not provide any explanation. This is the main reason why it needs to be checked client side.

packages/itwin/grouping-mapping-widget/src/widget/components/PropertyMenu.tsx
```diff
@@ -74,6 +107,10 @@ export const PropertyMenu = ({
   hideCalculatedProps = false,
   hideCustomCalculationProps = false,
 }: PropertyModifyProps) => {
+  const groupId = group.id ?? "";
```
Guillar1, line 110: While we are at it, we should replace the rest with this as well in this file.

packages/itwin/grouping-mapping-widget/src/widget/utils.ts
```diff
@@ -5,3 +5,11 @@
 export type CreateTypeFromInterface<Interface> = {
   [Property in keyof Interface]: Interface[Property];
 };
+
+export function debounce<F extends (...args: any[]) => void>(f: F, delay: number) {
+  let timer: NodeJS.Timeout;
```
Guillar1, line 10: Lets steer away from using Node.js types, this will cause issues in the browser due to the Node API not being available. Eventually, we are switching to Webpack 5, which gets rid of the automatic pollyfilling of the Node.js APIs. Try using the window object.

packages/itwin/grouping-mapping-widget/src/widget/components/PropertyMenu.tsx
```diff
@@ -144,11 +213,22 @@ export const PropertyMenu = ({
     [],
   );
 
-  const calculatedPropertyReturn = useCallback(async () => {
+  const groupPropertyReturn = useCallback(async (modified: boolean) => {
+    setPropertyMenuView(PropertyMenuView.DEFAULT);
+    modified && refreshGroupProperties();
```
Guillar1, line 218: Should probably await these promises.

Guillar1 (approved):

PR #237, opened 2022-05-05
[property-grid-react] fix ui items provider so the property grid appears when loaded into appui v1.0

jason-crow:
The current implementation of the property grid when using a ui items provider is to rely on the useEffect in the MultiElementGrid to handle hiding / showing the property grid on selection events. This won't work in appui v1.0 because the component will be unmounted and cannot have its state changed from that useEffect. 

The workaround is to implement a mechanism similar to what Bill did here: https://github.com/iTwin/itwinjs-core/blob/ce886ecde5caec43972f7b6a4dd4cf3cd13eba06/test-apps/ui-items-providers-test/src/ui/providers/NetworkTracingUiProvider.tsx#L69

If the PropertyGridUiItemsProvider is created in appui v1.0, then a listener will be added (that is cleaned up during the onUnregister callback) that will handle the toggling of the property grid's widget state in place of the useEffect, which will not run if in appui v1.0

### DIFF
--- a/common/changes/@itwin/property-grid-react/fix-property-gruid-provider-ui-1.0_2022-05-05-15-05.json
+++ b/common/changes/@itwin/property-grid-react/fix-property-gruid-provider-ui-1.0_2022-05-05-15-05.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/property-grid-react",
+      "comment": "fix property grid ui items provider when loading into appui v1.0",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@itwin/property-grid-react"
+}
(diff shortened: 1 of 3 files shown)


aruniverse (approved):
Don't love needing to support ui1.0 and having additional code for it. But I guess since it's temporary it's fine

JoeZman (approved):

jason-crow:
> Don't love needing to support ui1.0 and having additional code for it. But I guess since it's temporary it's fine

yeah I agree, my take is a little bit of code to bridge the gap now is fine to keep up from having to do something more extreme like fork this version. I don't think any large code changes pandering to 1.0 should be done at this point, but this seems succinct enough to be worth doing and we can remove later on once 1.0 is abandoned for good

PR #242, opened 2022-05-09
[property-grid-react] Add ability to persist shared favorites via callbacks from consuming app

cortneysmithbentley:
These changes enable an application using the property-grid package to manage sharing and unsharing favorite properites, and persisting these shared properties with its own storage method.

### Added PropertyGridProps Properties
- **customOnDataChanged** (IPresentationPropertyDataProvider): Used as a callback to make changes to the data provider, in this case to make sure that shared favorites are shown for all users in the model.

- **actionButtonRenderer**: Used to display the shared icon on shared properties, similar implementation to previous version just decoupled from the PropertyGrid. 

### Added ContextMenuItemInfo Properties
In order to pass in Share Favorite and Unshare Favorite via the additionalContextMenuOptions, had to make these changes to the context menu items.

- **PropertyGridDefaultContextMenuKey**: new string enum to easily identify the existing default context menu items for the property grid by replacing the hard-coded string keys.

-  **hideDefaultContextMenuItem** (PropertyGridDefaultContextMenuKey): Pass in the key for a default context menu item that you want to hide when displaying yours, essentially replacing it. Needed this functionality so that if a favorite is shared, the context menu hides "Remove Favorite" and shows "Unshare Favorite". The user must unshare before removing a favorite.

- **isValid** (PropertyRecord, Field): A callback to determine if the context menu option is valid (aka should be visible). Needed for displaying "Share Favorite" and "Unshare Favorite" when appropriate. Pass both PropertyRecord and Field because checking if a property is already shared needs the name from the PropertyRecord, and checking if a property is a favorite but is unshared needs the Field to pass to Presentation.


cortneysmithbentley:
Need to do the rush change

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -178,6 +165,9 @@ export const PropertyGrid = ({
       if (propertyData) {
         setTitle(propertyData?.label);
         setClassName(propertyData?.description ?? "");
+        if (dataProvider && customOnDataChanged) {
+          customOnDataChanged(dataProvider);
```
jason-crow, line 169: this should be async

packages/itwin/property-grid/src/types.ts
```diff
@@ -30,6 +38,10 @@ export interface PropertyGridProps {
   isOrientationFixed?: boolean;
   enableFavoriteProperties?: boolean;
   favoritePropertiesScope?: FavoritePropertiesScope;
+  customOnDataChanged?: (
+    dataProvider: IPresentationPropertyDataProvider
+  ) => void;
```
jason-crow, line 43: Promise<void>

packages/itwin/property-grid/src/types.ts
```diff
@@ -8,16 +8,24 @@ import type {
...
+  PropertyRecord,
   StagePanelLocation,
   StagePanelSection,
 } from "@itwin/appui-abstract";
 
 export type ContextMenuItemInfo = ContextMenuItemProps &
-React.Attributes & { label: string };
+  React.Attributes & {
+    label: string;
+    hideDefaultContextMenuItem?: PropertyGridDefaultContextMenuKey;
```
jason-crow, line 26: I think a better way of handling this hideDefaultContextMenuItem would be to have a way to override one or more props of any default context menu, which can be mapped by the enum you created, so that you could for instance pass an isValid handler for the favorite options

Open to suggestions, but we would want at least to ensure that each individual context menu option doesnt have its own prop to hide another, it should be a sibling of the additionContextMenuOptions

cortneysmithbentley: What would need to be overridden if not visibility? Otherwise we're just creating a more complicated way to hide the menu item by overriding its isValid call to hide it when the additional context menu option is valid. I could rename it to something like replaceDefaultContextMenuItem. Otherwise, if I'm understanding correctly, it doesn't sound like that level of abstraction is needed here, since I'm not sure what other use case there is (or what other user there would be, even).

jason-crow: The point is that props of a context menu item shouldn't be coupled to the unique behavior of any one context menu item. i.e. we use a `onSelect` instead of a `onShareFavorite`. Having `hideDefaultContextMenuItem` as a prop makes some specific assumptions about a particular context menu's behavior, such as that it may need to hide exactly one other context menu presumably when this context menu's isValid is true. This would need to be amended if needed to hide more than one context menu item or a context menu item that isn't a default option. It would be preferable therefore to find a way to avoid adding an additional prop that caters only to a specific use case. IMO the solution is to have a way to override default context menu item attributes as this is generally useful and provides an intuitive way to express 'this context menu cant show for the same reason this other one should show' without adding any new concept. 

It also stands to reason that a consumer may wish to modify something about a default component without having to make that change here as well, so it solves your problem and potentially a future problem

cortneysmith: This is pushed if you want to give it a look @jason-crow

cortneysmithbentley: this is pushed if you want to take a look @jason-crow

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -303,6 +294,24 @@ export const PropertyGrid = ({
           }
         }
 
+        // Do any overrides on default menu options
+        if(defaultContextMenuOptions?.size && defaultContextMenuOptions.size > 0 ) {
```
jason-crow, line 298: after the `if` or `for` there should be a space (look for all applicable examples)

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -303,6 +294,24 @@ export const PropertyGrid = ({
...
+          for(const key of Object.values(PropertyGridDefaultContextMenuKey)) {          
+            const overrides = defaultContextMenuOptions?.get(key);
+            const itemIndex = items.map((item) => item.key).indexOf(key);
```
jason-crow, line 301: you should gate the next 2 lines after getting overrides so that they only run if overrides exist

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -303,6 +294,24 @@ export const PropertyGrid = ({
...
+            items[itemIndex] = { ...items[itemIndex], ...overrides};
+          }
+        }
+
+        // Verify all existing options are valid, and if not remove them
+        for(let i = items.length - 1; i >= 0; --i) {
+          const item = items[i];
+          if(item.isValid && field && !item.isValid(args.propertyRecord, field)) {
```
jason-crow, line 309: althought its correct syntax, is confusing to use item.isValid in a condition this way as the fact that it is a function isn't obvious, it reads as though it could be a boolean and that makes the logic of this condition confusing to read. 

So doing `item.isValid !== undefined` is more clear

jason-crow, line 309: Grigas informed me that there are cases where a field could be undefined without implying an error state, he gave the example of if the property is a struct or array. I think it makes sense then that `isValid`'s definition allows for the `field` argument to be conditional

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -303,6 +294,24 @@ export const PropertyGrid = ({
...
+            items.splice(i, 1);
+
```
jason-crow, line 311: unnecessary newline here

jason-crow (approved):

packages/itwin/property-grid/src/types.ts
```diff
@@ -30,13 +37,19 @@ export interface PropertyGridProps {
   isOrientationFixed?: boolean;
   enableFavoriteProperties?: boolean;
   favoritePropertiesScope?: FavoritePropertiesScope;
+  customOnDataChanged?: (
+    dataProvider: IPresentationPropertyDataProvider
+  ) => Promise<void>;
+  actionButtonRenderer?: ActionButtonRenderer;
```
aruniverse, line 43: Shouldn't this have been an array?

cortneysmithbentley: I'm only passing in the one renderer, but I can make it an array just in case for future uses (plus I know the virtualized property grid takes an array).

PR #244, opened 2022-05-10
[property-grid-react] manually bump to 0.4.2

jason-crow:
manually bump to 0.4.2 to allow previous commits to be published


aruniverse (approved):
I think this will publish as 0.4.3

PR #249, opened 2022-05-12
[property-grid-react] Element name and class name should be aligned with property groups

cortneysmithbentley:
Removed left margin on the property grid panel header that was causing the element name and class name to be misaligned.


aruniverse (commented):
please see comment from other pr: https://github.com/iTwin/viewer-components-react/pull/242#discussion_r871605651

JoeZman (approved):

PR #253, opened 2022-05-16
[Grouping&Mapping] Add an extraction enabled toggle

lzaromskis:
This PR adds an extraction toggle in the mapping add/modify view and also in the three dot menu in the mappings table.

When the extraction is toggled from the three dot menu then an overlay with a loading circle is displayed to block user clicks

Toggle in the add/modify view:
add_modify

Toggle in the three dot menu:
threedotmenu

The blocking overlay:
overlay


bingjiez:
Could we also add an icon for enable extraction? It kills my OCD lol

Guillar1:
> Could we also add an icon for enable extraction? It kills my OCD lol

Let's use the process SVG. 
image

lzaromskis:
Added an icon
image

Guillar1 (approved):

PR #254, opened 2022-05-17
[property-grid-react] support overriding the auto child category expansion behavior

jason-crow:
allow for overriding the child category expansion
default behavior:
image

if `autoEpandChildCategories` is set to false this will be the result:
image


aruniverse (approved):

grigasp:
@jason-crow you might want to have a look at how we let users choose the behavior in our [rules editor](https://qa-presentationruleseditor.bentley.com/open-imodel?itwinId=656dd74d-8798-4aa4-8d13-6e6458789639&imodelId=881c9ca0-34ff-4875-ae63-2dd8ac015c27):

PR #256, opened 2022-05-18
[property-grid-react] if the property grid tab is minimized do not force it open

jason-crow:
when the property grid is a minimized tab like so:
image
or merged into a zone with another active like so:
image
in those cases we do not want to force the property grid to open when a selection event occurs, instead we only want to do this when the property grid is hidden

Note: since this affects only tab behavior it only affects the property grid behavior in app ui v1

### DIFF
--- a/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
+++ b/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
@@ -40,7 +40,10 @@ const onPresentationSelectionChanged = async (evt: SelectionChangeEventArgs, sel
       }
     );
     if (instanceKeys.some((key) => !Id64.isTransient(key.id))) {
-      widgetDef.setWidgetState(WidgetState.Open);
+      // if the widget is minimized, do not force it open
+      if (widgetDef.activeState === WidgetState.Hidden) {
+        widgetDef.setWidgetState(WidgetState.Open);
+      }
     } else {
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -40,7 +40,10 @@ const onPresentationSelectionChanged = async (evt: SelectionChangeEventArgs, sel
       }
     );
     if (instanceKeys.some((key) => !Id64.isTransient(key.id))) {
-      widgetDef.setWidgetState(WidgetState.Open);
+      // if the widget is minimized, do not force it open
```
aruniverse, line 43: could you comment this is for appui 1.0?

jason-crow: if you look below this whole routine is in an if block for that
```typescript
if (UiFramework.uiVersion === "1") {
      this._removeListenerFunc = Presentation.selection.selectionChange.addListener(onPresentationSelectionChanged);
    }

aruniverse: sorry didnt look at the larger diff. resolving

jason-crow: i added a comment to the routine as well though so its more clear

PR #260, opened 2022-05-25
Allow default widget priority

JoeZman:
Allow passing default widget priority to both Tree & Property Grid item providers.

### DIFF
--- a/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
+++ b/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
@@ -101,6 +101,7 @@ export class PropertyGridUiItemsProvider implements UiItemsProvider {
         getWidgetContent: () => <MultiElementPropertyGrid {...this._props} />,
         defaultState: WidgetState.Hidden,
         icon: "icon-info",
+        priority: this._props?.defaultPanelWidgetPriority,
       });
     }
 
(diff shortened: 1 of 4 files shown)


aruniverse (commented):
You need to run `rush change`, `patch` should be ok. Just to confirm, you don't need to override the measure-tools widget?

common/changes/@itwin/property-grid-react/Allow-default-priority_2022-05-25-21-03.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/property-grid-react",
+      "comment": "Allow a caller to specify a default priority for Tree / Property item providers.",
```
aruniverse, line 5:
```suggestion
      "comment": "Allow a caller to specify a default priority for PropertyGrid item provider.",
```

common/changes/@itwin/tree-widget-react/Allow-default-priority_2022-05-25-21-03.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "Allow a caller to specify a default priority for Tree / Property item providers.",
```
aruniverse, line 5:
```suggestion
      "comment": "Allow a caller to specify a default priority for Tree Widget item provider.",
```

aruniverse (approved):

JoeZman:
For now, these are the only two (really only the Tree provider is necessary).  This is for UI 1.0.  For 2.0, we will probably have to revisit the remaining providers, some of them, Design Review is currently not using.

PR #263, opened 2022-05-26
[One Click LCA] Update error message of job creation for unauthorized users

bingjiez:
Before, if an unauthorized user tries to trigger a carbon calculation job, the export modal will show a spinner hanging forever and the error message is showed in console. To make it more user friendly, we provide a toaster notification with detailed information and show the job failed in the modal.

Also updated error message and notification for other types of failures like signing in, getting job status etc.

image


aruniverse (approved):
lol nice catch,

sorry didnt catch this earlier
you are missing rush change

PR #264, opened 2022-05-27
[Grouping & Mapping] Formula validation fixes

lzaromskis:
This PR fixes some formula validation issues.

- Fixed not being allowed to reference other custom calculations.
- Fixed not being allowed to reference properties which have capital letters in their name.


Guillar1:
Upgrading insights-client introduces some breaking changes because the newest insight-client version deals with custom URL's differently. It no longer takes a prefix but rather an entirely new base URL. Everywhere a new ReportsClient is created will have to be changed to support this new logic with some function to generate the custom base URL.

A quick solution that can be used but don't have to.
```
export const prefixUrl = (baseUrl?: string, prefix?: string) => {
  if (prefix && baseUrl) {
    return baseUrl.replace("api.bentley.com", `${prefix}-api.bentley.com`);
  }
  return baseUrl;
};

new ReportingClient(prefixUrl(REPORTING_BASE_PATH, apiContext.prefix))
```

packages/itwin/grouping-mapping-widget/src/widget/components/utils.tsx
```diff
@@ -152,3 +153,15 @@ export const LoadingSpinner = () => {
     </div>
   );
 };
+
+const prefixUrl = (baseUrl?: string, prefix?: string) => {
+  if (prefix && baseUrl) {
+    return baseUrl.replace("api.bentley.com", `${prefix}api.bentley.com`);
```
Guillar1, line 159: `${prefix}-api.bentley.com` missing the dash after prefix.

Guillar1 (approved):

Guillar1:
@aruniverse We need approval due to package.json changes.

packages/itwin/grouping-mapping-widget/src/widget/components/utils.tsx
aruniverse: This will be fine. In prod it's just API.bentley.com not prod-api.bentley.com

aruniverse: Nvm I guess it depends one what you guys use as prefix. Everywhere else we include the dash in it

aruniverse (approved):
Lock file changes look ok to me

packages/itwin/grouping-mapping-widget/src/widget/components/utils.tsx
Guillar1: I prefer the whole URL to future proof. Either we get rid of imodel-browser-react or add support for custom URL's. As of now it only lets you change the prefix and they also don't include the dash.

packages/itwin/grouping-mapping-widget/src/widget/components/utils.tsx
```diff
@@ -152,3 +153,15 @@ export const LoadingSpinner = () => {
...
+  }
+  return baseUrl;
+};
+
+export const getReportingClient = (prefix?: "" | "qa" | "dev") => {
+  const url = prefixUrl(REPORTING_BASE_PATH, prefix ?? process.env.IMJS_URL_PREFIX);
```
Guillar1, line 165: Remove IMJS_URL_PREFIX as we are not planning on supporting that.

lzaromskis: Removed

packages/itwin/grouping-mapping-widget/src/widget/components/utils.tsx
lzaromskis: Added a dash.

PR #271, opened 2022-06-06
[Property Grid] Added persistNullValueToggle to property grid props, and the code to persist it to UserPreferences

cortneysmithbentley:
These changes enable the consuming application to choose for the hide / show empty fields option to be persisted in UserPreferences. 

- Added `persistNullValueToggle` in `PropertyGridProps`.
- Added `getShowNullValues()` to the `onDataChanged()` use effect to pull and set the toggle value.
- Removed a lot of code for the filter, as it was only used for hiding and showing empty values. Kept the default placeholder filter as it's needed to pass to the virtualized property grid to render it, and might be used later for actual filtering I suppose.
- Instead of using filter, now using the built-in option in the data provider (`AutoExpandingPropertyDataProvider.includeFieldsWithNoValues`). 
- Changed `onHideNull()` and `onShowNull()` to a single function `updateShowNullValues()` which expects a boolean argument that is the new value.


packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -171,8 +172,37 @@ export const PropertyGrid = ({
           await customOnDataChanged(dataProvider);
         }
       }
+      if(persistNullValueToggle) {
+        let res = await getShowNullValues();
+        if(res !== undefined && dataProvider)
+        {
+          dataProvider.includeFieldsWithNoValues = res;
+          setShowNullValues(res);
+        }
```
jason-crow, line 181: could you run prettier against this, there's formatting issues in several places like unexpected trailing whitespace and no space in between the `if` and `(`

cortneysmithbentley: Ah gotcha, yeah I always miss that stuff since it's not part of the build or commit lol. I'll run it and push fixes now.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -200,19 +258,7 @@ export const PropertyGrid = ({
     },
 
     [iModelConnection, favoritePropertiesScope]
-  );
-
-  const onHideNull = useCallback(() => {
-    setFilterer(new NonEmptyValuesPropertyDataFilterer());
```
jason-crow, line 206: why is the NonEmptyValuesPropertyDataFilterer no longer used?

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -171,8 +172,37 @@ export const PropertyGrid = ({
...
     };
 
+    const getShowNullValues = async () => {
+      const userPrefs = IModelApp.userPreferences;
+      if(userPrefs) {
+        const accessToken = await IModelApp.getAccessToken();
+        const iModel = UiFramework.getIModelConnection();
+        const iTwinId = iModel?.iTwinId;
+        const iModelId = iModel?.iModelId;
+        const showNullValuesRes = await userPrefs.get({
```
jason-crow, line 192: it would be much cleaner to move all the code related to getting and setting the null value toggle in the user preferences to another file so you can simply invoke it here, it always needs to have try / catch handling and logging in place for the error cases

packages/itwin/property-grid/src/components/PropertyGrid.tsx
cortneysmithbentley: As I mentioned in the PR description, I updated it to use this property from the data provider itself: `AutoExpandingPropertyDataProvider.includeFieldsWithNoValues`. The filter was only used for this purpose and seemed like overkill when there's a built-in way to handle it already.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
cortneysmithbentley: I can add the try/catch and use the logger from core. Based on the rest of the file, it seems weird to only move out the two functions I wrote to their own file. This same file already handles adding/removing favorites, building it's own context menu, etc. If we're gonna refactor it then that's fine, but it just seems weird to move this one thing out. I can if I need to though.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -171,6 +172,34 @@ export const PropertyGrid = ({
           await customOnDataChanged(dataProvider);
         }
       }
+      if(persistNullValueToggle) {
+        const res = await getShowNullValues();
```
jason-crow, line 176: it seems like running this in the onDataChanged will cause it run more than it needs to be, it basically needs to be called just once then it only changes on user action

cortneysmithbentley: I'm not sure where else to run it. It's async so I can't just run it by itself after the variable declarations. I could put it in a use callback of its own, but it depends on propertydata so it'd still run whenever the data is changed.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
jason-crow: was it not working correctly? why do this now?

packages/itwin/property-grid/src/components/PropertyGrid.tsx
jason-crow: it depends on the dataProvider, but does it need to run in ever onDataChanged event? This will lead to for example it running this code when you modify favorites etc

cortneysmithbentley: True. I will move it to a callback dependent on dataProvider.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
cortneysmithbentley: I mean I can switch it back if you really want, it just seemed like an overly complicated way to do it when there's already an option in the data provider and nothing else in property grid uses the filter. And if there's a simple way to do it, then when what better time to change than when I'm already updating the code?

jason-crow: couple thoughts...

- if you make this change then you have to consider the following
   - If you are removing the usage of NonEmptyValuesPropertyDataFilterer should you go ahead an also remove the PlaceholderPropertyDataFilterer (does that one only exist to counter the NonEmptyValues one and have no purpose when by itelf)
   - if you are removing these and they are no longer used should you remove the definitions for these filters from the source code
   - are you positive this change introduces no regression

wrt to 'what better time to change than when I'm already updating the code', i think that the scope of a PR is important. That is to say that just because you're modifying a file does not mean it's prudent to modify anything in that file worth changing. Here a couple reasons why:
- those changes may get conflated with the purpose of the PR they were made in. i.e. a developer may assume these changes explicitly relate to persisting null values versus just a more efficient way to toggle null values
- code change has inherent risk and if you include changes in a PR that are unnecessary (strictly speaking), if they cause an issue they will impede the resolution of the issue the PR intends to fix whereas if those changes were isolated they could also be resolved separately
- changes unrelated to the express purpose of a PR will obfuscate the changes that are relevant and make it more challenging to identify them for improvements/fixes later on (this is also why moving code more than a few lines into a separate file that has a single responsibility is a good thing to do as it makes it quicker and less risky to modify code pertaining to that responsibility)

packages/itwin/property-grid/src/components/PropertyGrid.tsx
cortneysmithbentley: Moved this to a useMemo that's just dependent on persistNullValueToggle.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
cortneysmithbentley: All current changes have been checked for lint errors and fixes for those are pushed.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -160,6 +162,21 @@ export const PropertyGrid = ({
     };
   }, []);
 
+  // If persisting hide/show empty values, get the preference
+  useMemo(() => {
```
jason-crow, line 166: useMemo if for setting a value that needs to be changed when the deps change, useCallback is for setting a function, and useEffect is for running some code, which in this case is the right one to use

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -160,6 +162,21 @@ export const PropertyGrid = ({
...
+    const getPreferences = async () => {
+      if(persistNullValueToggle) {
+        const res = await getShowNullValuesPreference();
+        if(res !== undefined) {
```
jason-crow, line 170: format changes please

cortneysmithbentley: I'm not sure what you mean here. Everything passes the linter.

packages/itwin/property-grid/src/components/PropertyGrid.tsx
```diff
@@ -322,11 +327,25 @@ export const PropertyGrid = ({
 
         setContextMenuItemInfos(items.length > 0 ? items : undefined);
       }
+
+      // Fcn for updating toggle for Hide / Show Empty Fields menu options
+      const updateShowNullValues = async (value: boolean) => {
```
jason-crow, line 332: move this out of the buildContextMenu routine and have it use the `useCallback`, basically it should be idential setup to the `onHideNull`/`onShowNull` except that it's combing their logic and additionally persisting the value

packages/itwin/property-grid/src/components/PropertyGrid.tsx
jason-crow: look at the `if(` it should be `if (` and this should happen automatically when you save the file if you have the settings found in `.vscode/settings.json` here running:
```json
"editor.formatOnSave": true,
  ```
  perhaps something is overriding that for you

jason-crow (approved):

Issue #268, opened 2022-06-02
[measure-tools] add some configurability to the provider


aruniverse:
should be able to override the provider to set default priority of the group

_Originally posted by @aruniverse in https://github.com/iTwin/viewer/pull/177#discussion_r888178338_

aruniverse:
@johnnyd710 can you help @achrysaetos  with setting up a [CoSpace](https://github.com/aruniverse/cospace) to test his changes out for this next week?

PR #273, opened 2022-06-13
[measure-tools] Expose some customizability in provider

achrysaetos:
Issue https://github.com/iTwin/viewer-components-react/issues/268. Replace core's measure-tools with provider's measure-tools:
image


packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -57,7 +57,7 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
         return [
           ToolbarItemUtilities.createGroupButton(
             "measure-tools-toolbar",
-            500,
+            1000000,
```
achrysaetos, line 60: Does this look right? If I remove a 0, the button moves back up to the top, but 1000000 seems like an unusually large number.

aruniverse: @bsteinbk this doesnt seem right. Any idea?

Also, that debug tool you showed off a few months ago to show the item priority, is that available anywhere?

bsteinbk: @aruniverse I see that constructor takes a prop for itemPriority that seem like a  good idea, to future-proof the api you may also allow a groupPriority to be specified. 

As for the tool I demoed - it is only in ui-test-app. I will try to separate it out into a debugToolItemsProvider.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -24,6 +24,11 @@ import { UiFramework } from "@itwin/appui-react";
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
+  private _props?: { itemPriority: number };
```
aruniverse, line 27: could you pull this out and type this, and just reuse it?

aruniverse: could you also update it to add groupPriority like Bill mentioned?

achrysaetos: > could you pull this out and type this, and just reuse it?

You mean create a new interface for it in somewhere like MeasureTools.ts?

aruniverse: yes create a new interface/type, in this file would be ok i think

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -22,12 +22,21 @@ import { MeasurementPropertyWidget, MeasurementPropertyWidgetId } from "./Measur
...
 
+interface Priority {
+  itemPriority?: number;
+  groupPriority?: number;
+}
+
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
-  private _props?: { itemPriority: number };
+  private _props;
```
aruniverse, line 32: you should type this.

also you should export the interface. 
also it just happens to be about priority right now, but it might expand later on to add other options. Something like MeasureToolsUiProviderOptions might be a better fit

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -22,12 +22,21 @@ import { MeasurementPropertyWidget, MeasurementPropertyWidgetId } from "./Measur
 import { AbstractZoneLocation } from "@itwin/appui-abstract";
 import { UiFramework } from "@itwin/appui-react";
 
+interface Priority {
+  itemPriority?: number;
+  groupPriority?: number;
```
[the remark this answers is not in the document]
achrysaetos: Do you know where might be a good place to use this? None of the toolbar classes have methods that take groupPriority as a parameter.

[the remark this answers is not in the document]
aruniverse: ok awesome, i was wondering if that played a role

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -57,12 +71,13 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
         return [
           ToolbarItemUtilities.createGroupButton(
             "measure-tools-toolbar",
-            500,
+            this._props.itemPriority!,
```
bsteinbk, line 74: I am not a fan of adding the `!`

achrysaetos: But itemPriority is a required field of createGroupButton, so if I remove the exclamation mark TypeScript complains of `Argument of type 'number | undefined' is not assignable to parameter of type 'number'.` even though this._props.itemPriority is never undefined because it is always defaulted to either 500 or the user's custom input. Do you think there's a better way to do this?

aruniverse: instead of providing defaults in the consturctor you could just add the defaults here

aruniverse (approved):

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -71,13 +67,13 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
...
             "measure-tools-toolbar",
-            this._props.itemPriority!,
+            this._props?.itemPriority ?? 20,
             "icon-measure",
             MeasureTools.localization.getLocalizedString(
               "MeasureTools:MeasurementGroupButton.tooltip",
             ),
             ToolbarHelper.constructChildToolbarItems(tools),
-            { groupPriority: this._props.groupPriority },
+            { groupPriority: this._props?.groupPriority ?? 10 },
```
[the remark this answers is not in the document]
aruniverse: @bsteinbk we're hoping for this provider to replace the measure tools provided by the standard content tools provider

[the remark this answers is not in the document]
bsteinbk: @achrysaetos @aruniverse - The separator line in the image above could only occur if the group priorty is being set on the other two buttons. Where is the definition of the other buttons in this toolbar coming from? Also what is the frontstage definition look like that is being used in this example, because it may be possible that the group priority is being set in fronstage appdata..  My only point is that most apps do not set a group priority so setting one by default may cause a separator to appear where one is not wanted.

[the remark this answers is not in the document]
bsteinbk: @aruniverse  @achrysaetos  Applogies...I am blaming it on old age.... I took time to actually look at me standard tool provider and see that it defaults the groupPriority to 10 just like you had it.  Sorry I should have looked before commenting initially.

bsteinbk (approved):

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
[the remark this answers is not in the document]
aruniverse: No problem, i was unaware of the group priority being a thing until now, good to know :)

PR #279, opened 2022-06-15
[Grouping&Mapping] Added extension point for providing a custom MappingClient

MindaugasButkus:
Added extension point for providing a custom IMappingClient implementation. This is not a breaking change, since the added parameter is optional. Added parameter defaults to ReportingClient which implicitly implements IMappingClient interface.

IMappingClient contains all the methods that are used by the widget for storing the Mapping configuration.

Having this will allow to decorate MappingClient requests with logging, caching, requests to other services, etc.

I also moved some generic types to a single line instead of having them split into multiple lines. It was not possible to save files in VSCode without getting linter errors. It conflicts with @typescript-eslint/indent rule ([known issue, it will not get fixed](https://github.com/typescript-eslint/typescript-eslint/issues/1824)). It is most noticeable for `Dispatch<SetActionState<...>>` types. I don't think we should be passing functions of such types to child components anyway, but changing this is out of scope for this PR.


packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
```diff
@@ -7,21 +7,51 @@ import { Mappings } from "./Mapping";
...
+   */
+  client?: IMappingClient;
 }
 
-const GroupingMapping = ({ accessToken, prefix }: GroupingMappingProps) => {
+const GroupingMapping = (props: GroupingMappingProps) => {
   const [currentAccessToken, setCurrentAccessToken] = useState<string>("");
+  const [mappingClient, setMappingClient] = useState<IMappingClient>(createDefaultMappingClient());
+
+  const { accessToken, prefix } = props;
```
Guillar1, line 53: Why not destructure this in the parameters?

packages/itwin/grouping-mapping-widget/src/widget/components/CalculatedPropertyTable.tsx
```diff
@@ -31,11 +30,7 @@ interface CalculatedPropertyTableProps {
   iModelId: string;
   mappingId: string;
   groupId: string;
-  setSelectedCalculatedProperty: React.Dispatch<
-  React.SetStateAction<
-  CreateTypeFromInterface<CalculatedPropertyType> | undefined
-  >
-  >;
+  setSelectedCalculatedProperty: React.Dispatch<React.SetStateAction<CreateTypeFromInterface<CalculatedPropertyType> | undefined>>;
```
Guillar1, line 33: Not for this PR. I just noticed that this is wrapped twice.

MindaugasButkus: What do you mean by "wrapped twice"? All I did here was put the whole generic type into a single line, because VSCode kept indenting the generic types and the linter kept failing because of that.

Guillar1: Its unrelated to this PR. I just noticed that CreateTypeFromInterface is being used on CalculatedPropertyType which does nothing. This change made me notice it.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
MindaugasButkus: Destructured.

Guillar1 (approved):

PR #285, opened 2022-06-21
[Grouping&Mapping] accessToken parameter now takes a callback fn

Guillar1:
accessToken now takes a callback function.


packages/itwin/grouping-mapping-widget/src/widget/GroupingMappingWidget.tsx
```diff
@@ -14,19 +14,19 @@ import {
...
   public readonly id = "GroupingMappingProvider";
 
-  private readonly _accessToken?: AccessToken;
+  private readonly _getAccessToken?: () => Promise<AccessToken>;
   private readonly _prefix?: ClientPrefix;
   private readonly _client?: IMappingClient;
 
-  constructor(accessToken?: AccessToken, prefix?: "" | "dev" | "qa", client?: IMappingClient) {
-    this._accessToken = accessToken;
+  constructor(getAccessToken?: () => Promise<AccessToken>, prefix?: "" | "dev" | "qa", client?: IMappingClient) {
```
MindaugasButkus, line 28: Since you are making a breaking change here, maybe change all these parameters to a single optional params object? It's pretty annoying when I want to provide only the `client` - need to pass 2 `undefined` values before I can pass in the client.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
```diff
@@ -350,9 +350,9 @@ const GroupPropertyAction = ({
       let newEcProperties: ECProperty[];
       // Fetch already existing ec properties then add all classes from presentation
       if (groupPropertyId) {
-        // TODO Error handling
```
MindaugasButkus, line 353: Where are the errors handled?

packages/itwin/grouping-mapping-widget/src/widget/GroupingMappingWidget.tsx
MindaugasButkus: Maybe `GroupingMappingProps` could be reused here.

Guillar1: Good idea

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
Guillar1: Oh, forgot to go back to this.

packages/itwin/grouping-mapping-widget/src/widget/GroupingMappingWidget.tsx
Guillar1: Converted to a single object params.

packages/itwin/grouping-mapping-widget/src/widget/GroupingMappingWidget.tsx
```diff
@@ -14,19 +14,19 @@ import {
...
   public readonly id = "GroupingMappingProvider";
 
-  private readonly _accessToken?: AccessToken;
+  private readonly _getAccessToken?: () => Promise<AccessToken>;
   private readonly _prefix?: ClientPrefix;
   private readonly _client?: IMappingClient;
 
-  constructor(accessToken?: AccessToken, prefix?: "" | "dev" | "qa", client?: IMappingClient) {
-    this._accessToken = accessToken;
+  constructor({ getAccessToken, prefix, client }: GroupingMappingProps = {}) {
```
MindaugasButkus, line 28: I don't think it's worth to deconstruct props here since later the same props object is rebuilt. If this was not reconstructed then it could be used like this: `<GroupingMapping {...this._props} />`.

Guillar1: I agree. Changed it to pass the object directly.

MindaugasButkus: It could be simplified even further - the separate field declaration is not needed, there is a shorthand for constructors:
`constructor(private readonly _props: GroupingMappingProps = {}) {}`

Guillar1: Oh right completely forgot about that shorthand. Thanks!

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
```diff
@@ -69,13 +41,21 @@ const GroupingMapping = ({ accessToken, prefix, client }: GroupingMappingProps)
   }, [clientProp]);
 
   return (
-    currentAccessToken ? <ApiContext.Provider value={{ accessToken: currentAccessToken, prefix }}>
+    <GroupingMappingApiConfigContext.Provider
+      value={{
+        getAccessToken:
+          getAccessToken ??
+          (async () =>
```
MindaugasButkus, line 48: I think this might cause excessive re-renders when no `getAccessToken` is provided because each time a new function (lambda expression) will be created. Do you know if multiple instances of this lambda will have reference equality? If not, maybe a non-lambda function would have reference equality (non exported function could be defined in the same file).

packages/itwin/grouping-mapping-widget/src/widget/components/SelectProject.tsx
```diff
@@ -54,6 +56,14 @@ const SelectProject = ({ onSelect, onCancel }: SelectProjectProps) => {
     [apiContext.prefix],
   );
 
+  useState(() => {
```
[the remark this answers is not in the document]
Guillar1: Yup.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
Guillar1: You are correct. The lambda will not have reference equality per render. I wrapped it in a useCallback. I also realized that the array dependencies could be more specific.

packages/itwin/grouping-mapping-widget/src/widget/components/SelectIModel.tsx
```diff
@@ -26,20 +27,29 @@ const SelectIModel = ({
   onCancel,
   backFn,
 }: SelectIModelProps) => {
-  const apiContext = useContext(ApiContext);
+  const { getAccessToken, prefix } = useGroupingMappingApiConfig();
+  const [accessToken, setAccessToken] = useState<AccessToken>();
 
   const apiOverrides = useMemo<ApiOverrides<IModelFull[]>>(
```
MindaugasButkus, line 33: Just a note about `useMemo`: https://reactjs.org/docs/hooks-reference.html#usememo
`You may rely on useMemo as a performance optimization, not as a semantic guarantee.`

This place looks like you are relying on `useMemo` as a semantic guarantee to prevent re-renderings.

packages/itwin/grouping-mapping-widget/src/widget/components/SelectIModel.tsx
```diff
@@ -26,20 +27,29 @@ const SelectIModel = ({
...
+  }, [getAccessToken]);
+
   return (
     <div className='imodel-grid-container'>
       <div className='imodel-grid'>
         <IModelGrid
           projectId={projectId}
           onThumbnailClick={onSelect}
-          accessToken={apiContext.accessToken}
+          accessToken={accessToken}
```
MindaugasButkus, line 52: There is a chance that an expired `accessToken` will be passed into `IModelGrid`. New `accessToken` will only be fetched if `getAccessToken` function changes, not when the token expires. It would be best if the `IModelGrid` component would accept a function to get the token instead of the token itself. Chance of the "expired" `accessToken` issue occuring is relatively low, so we can leave this one for another PR.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
Guillar1: I'll put the object in the value in a useState. Earlier, nothing was using the object as a dependency within the widget. Only the values within. Even if the object was recreated the properties would be unchanged and no rerender would happen. But I don't mind being safe about it.

packages/itwin/grouping-mapping-widget/src/widget/components/SelectIModel.tsx
Guillar1: Removed useMemo. I vaguely remember people screaming about this.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.tsx
Guillar1: I put the object in the value in a useState. Nevermind my earlier comment.

Guillar1:
@aruniverse I've added @MindaugasButkus to codeowners. Needs approval.

packages/itwin/grouping-mapping-widget/src/widget/components/SelectIModel.tsx
Guillar1: Haha. It wasn't directed to me. There was some discussion on the documentation being a bit contradictory. 
https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
`As a last resort, you can add a function to effect dependencies but wrap its definition into the` [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback)` Hook. This ensures it doesn’t change on every render unless its own dependencies also change:`

In the `useCallback` docs it says `useCallback(fn, deps) is equivalent to useMemo(() => fn, deps).` So inherently its also subject to future cache busting. So it is a little confusing to suggest this pattern.

aruniverse (approved):

aruniverse:
please have @MindaugasButkus approve before merging

MindaugasButkus (approved):
Approved.

PR #286, opened 2022-06-21
[Grouping&Mapping] exposed formula datatype resolver

MindaugasButkus:
Changed formula validator to not only validate, but to also return the resolved data type.


MindaugasButkus:
Need to update `DataType` values to match what is returned from the API.

packages/itwin/grouping-mapping-widget/src/formula/FormulaDataTypeResolver.ts
```diff
@@ -0,0 +1,42 @@
...
+    return { errorMessage: "Opened but not closed parenthesis found." };
+  } else if (ParenthesisState.NotOpened === parenthesisState) {
+    return { errorMessage: "Closed but not opened parenthesis found." };
+  }
+
+  let infixFormulaTokens;
+  try {
+    infixFormulaTokens = splitFormula(formula);
+  } catch (ex) {
+    return { errorMessage: (ex as Error).message };
```
Guillar1, line 32: Would it be possible to type guard this with `instanceOf`?

MindaugasButkus: Added type guard with fallback to `console.error` what was received.

PR #287, opened 2022-06-23
[property-grid-react] refresh property grid when frontstage changes

jason-crow:
- Currently if the  frontstage changes while a selection set is active the property grid will be in an improper state
- This occurs in both UI 1.0 and UI 2.0 but unfortunately a unique solution is required for both
  - in UI 1.0 the widget state must be refreshed from an onFrontstageReady listener that is setup from the ui items provider constructor
  - in UI 2.0 the state must be refreshed from an additional listener for onFrontstageReady that is setup in a useEffect inside the MultiElementPropertyGrid
  - The reason for the difference in solutions is that UI 2.0 the property grid stays mounted so the useEffect listeners can continue syncing its state with ongoing events in the app, while in UI 1.0 it is unmounted and will have to be remounted in the proper state based on these events

### DIFF
--- a/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
+++ b/packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
@@ -12,9 +12,10 @@ import {
   StageUsage,
   WidgetState,
 } from "@itwin/appui-abstract";
+import type { FrontstageDef, FrontstageReadyEventArgs } from "@itwin/appui-react";
 import { FrontstageManager, UiFramework } from "@itwin/appui-react";
 import { Id64 } from "@itwin/core-bentley";
-import type { InstanceKey } from "@itwin/presentation-common";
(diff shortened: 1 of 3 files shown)


packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -50,23 +54,47 @@ const onPresentationSelectionChanged = async (evt: SelectionChangeEventArgs, sel
...
+        removePresentationListener();
+        removeFronstageReadyListener();
+      };
     }
   }
 
   // When the provider is unloaded also remove the handler
   public onUnregister = () => {
-    this._removeListenerFunc && this._removeListenerFunc();
+    this._removeListeners && this._removeListeners();
```
aruniverse, line 97: nit, you could do

```suggestion
    this._removeListeners?.();
```

PR #289, opened 2022-06-30
Dynamic Location Measurement Tool

HughMCWJR:
Adds measurement that follows cursor so users can see current position of mouse.


common/changes/@itwin/property-grid-react/fix-property-grid-frontstage-refresh_2022-06-23-15-15.json
```diff
@@ -0,0 +1,10 @@
+{
```
aruniverse, line 1: why do you have this file?

aruniverse (commented):
@a-gagnon can you take a pass?

@HughMCWJR you probably need to merge with master, not sure why you have rush change log files for pkgs you didnt touch

HughMCWJR:
> @a-gagnon can you take a pass?
> 
> @HughMCWJR you probably need to merge with master, not sure why you have rush change log files for pkgs you didnt touch

I made some mistakes with git that causes those I believe.

a-gagnon:
- Added a ToolSettings to turn on/off the dynamic measurement in case some users don't want it.
- Reworked the code to obtain geolocation/slope off the locate.
- Added tests for the dynamic measurement.

It took a while to address this PR due to other priorities and that I had to fix AccuSnap in the iTwin Viewer to make sure our snap requests don't interfere with AccuSnap's. It turns out we were overwriting the default snap behavior using `SnapMode.Nearest`.

a-gagnon (approved):

PR #294, opened 2022-07-13
Group ordering by name

Simonas-A:
Groups are now ordered by name.

Only the 459th line has changed. The whole code was tab-incremented.


Simonas-A:
Should the group ordering be like in the mappings list (order when column header is clicked) or default ordering by group name is enough?
@arnobmallickbsw @Guillar1 

image
(ordering on column click)


image
(default ordering by group name)

MindaugasButkus:
> Should the group ordering be like in the mappings list (order when column header is clicked) or default ordering by group name is enough? @arnobmallickbsw @Guillar1
> 
> image (ordering on column click)
> 
> image (default ordering by group name)

I think the default ordering is fine for now. If it will not be enough, then we will look into this a bit closer.

Guillar1 (approved):

PR #298, opened 2022-07-22
Persist the show/hide null values toggle per user instead of per iModel or iTwin.

cortneysmithbentley:
Removed iTwinId and iModelId from calls to save/get null value toggle so that it is persisted for the user and not as a shared setting.

### DIFF
--- a/common/changes/@itwin/property-grid-react/persist-show-hide-null-per-user_2022-07-22-16-49.json
+++ b/common/changes/@itwin/property-grid-react/persist-show-hide-null-per-user_2022-07-22-16-49.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@itwin/property-grid-react",
       "comment": "Persist show/hide null value toggle per user and not per iTwin/iModel",
-      "type": "patch"
+      "type": "minor"
     }
   ],
   "packageName": "@itwin/property-grid-react"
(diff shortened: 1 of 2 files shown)


common/changes/@itwin/property-grid-react/persist-show-hide-null-per-user_2022-07-22-16-49.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/property-grid-react",
+      "comment": "Persist show/hide null value toggle per user and not per iTwin/iModel",
+      "type": "patch"
```
jason-crow, line 6: the is a major version bump change since its a breaking change

cortneysmithbentley: I didn't have the option for major change. It was minor or patch.

cortneysmithbentley: I can go back and do minor if you want

cortneysmithbentley: okay it's minor now, everybody calm down 😜

jason-crow (approved):

PR #299, opened 2022-07-26
[ReportingPlatform] ComboBoxes Fixes

Guillar1:
- Fixed ComboBox in Reports Config Widget which includes https://github.com/iTwin/iTwinUI-react/issues/697
- Waiting for https://github.com/iTwin/iTwinUI-react/issues/709 to be published since this is another ongoing bug. 
- In Reports Config, API Config is now being stored in a state to stop unnecessary rerenders.
- CSS Class names were not matching.
- Replaced the ComboBox in Group Properties to a simple Select as a bandaid fix due to the ComboBox getting trapped in an infinite loop. Not going to fix this since this component is getting a full refactor anyways.


bingjiez (approved):

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
```diff
@@ -28,11 +28,12 @@ import {
   SvgRemove,
 } from "@itwin/itwinui-icons-react";
 import type {
```
aruniverse, line 30: whats up with the import being on multiple lines for no reason?

aruniverse (approved):

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
aruniverse: comment applies for other files where imports were changed

Guillar1: Oh, I see what happened. 
I accidently had prettier split it into multiple lines. 
```
import {
  ReportsApiConfig,
  ReportsApiConfigContext,
} from "../context/ReportsApiConfigContext";
```
ESLint then pulled the type out from the import verbatim I am assuming. 

```
import type {
  ReportsApiConfig} from "../context/ReportsApiConfigContext";
import {
  ReportsApiConfigContext,
} from "../context/ReportsApiConfigContext";
```

PR #301, opened 2022-08-03
Bugfix/distance measurement ray3d

a-gagnon:
Temporarily fix a problem in core-geometry where Ray3d.createStartEnd captures the origin point instead of copying it. This solves an issue where the DistanceMeasurement's start point was being modified when it shouldn't have.
(This will eventually be fixed with [PR #4012 on itwinjs-core.](https://github.com/iTwin/itwinjs-core/pull/4012))

### DIFF
--- a/common/changes/@itwin/measure-tools-react/bugfix-distanceMeasurementRay3d_2022-08-03-03-43.json
+++ b/common/changes/@itwin/measure-tools-react/bugfix-distanceMeasurementRay3d_2022-08-03-03-43.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


aruniverse (approved):

aruniverse:
@a-gagnon you might want to backport that core fix pr to the release/3.3.x branch if you need it now, unless you want to wait til 3.4

a-gagnon:
> @a-gagnon you might want to backport that core fix pr to the release/3.3.x branch if you need it now, unless you want to wait til 3.4

All good, I just cloned the point as a temporary workaround for the measure-tools package. We'll pick up on this change eventually.

PR #302, opened 2022-08-03
[ReportsConfig] Missing dep

Guillar1:
Adding a missing dev dep.


aruniverse (commented):
Should this be a dev dep or a real one?

Guillar1:
> Should this be a dev dep or a real one?

Dev. It is used to mock it out in tests.

aruniverse (approved):

PR #303, opened 2022-08-03
property-grid-react: Fix broken show/hide empty values toggle when multiple elements selected.

cortneysmithbentley:
Selecting multiple elements for the property grid, navigating to the element list, then choosing a specific element would show a property grid where the show/hide empty values context menu item wouldn't work. 

Updated the code to use the disableUnifiedSelection toggle to... disable unified selection... within the FilteringPropertyGrid. This means it won't overwrite the property grid (created for the specific element from the multiple element list) with the elements that are current selected in the view (which is still multiple elements).

The isOverLimit value is hardcoded to false in this case, which should be fine. Disabling unified selection only occurs through the multielement property grid when a user has selected an element from the element list, so it will always be displaying a single element which isn't over the limit.


packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -130,7 +131,7 @@ export const FilteringPropertyGridWithUnifiedSelection = (
     ),
   }), []);
 
-  const { isOverLimit } = usePropertyDataProviderWithUnifiedSelection({ dataProvider: props.dataProvider as IPresentationPropertyDataProvider });
+  const { isOverLimit } = props.disableUnifiedSelection ? { isOverLimit: false } : usePropertyDataProviderWithUnifiedSelection({ dataProvider: props.dataProvider as IPresentationPropertyDataProvider });
```
jason-crow, line 134: what is the relationship of 'isOverLimit' with 'disableUnifiedSelection'

cortneysmithbentley: From debugging the code, it looks like unified selection in this case will automatically see which elements are selected in the view and show them in the property grid. The isOverLimit is making sure that the number of elements selected is within the limit to show in the property grid and is what is returned from this hook.

The only place we use disableUnifiedSelection is in the multi-element property grid (I searched GitHub and bentleycs on Azure). It only uses the property grid when displaying a single element that has been selected from the ElementList or when showing the multiple elements (where it shows property values as "varies" etc). 

I can try and figure out a way to make sure it's not over the limit without calling the unified selection hook if needed. But if this hook is called with the multi-element property grid, it breaks it (selecting an element from the ElementList would take you back to the "multiple elements selected" view instead of to the property grid for that specific element, since unified selection will see the multiple elements selected in the view).

cortneysmithbentley: When disableUnifiedSelection is true, the original property grid code just called VirtualizedPropertyGridWithDataProvider directly (the same thing that happens here if isOverLimit is false). So I think this implementation should be fine.

jason-crow: Thanks, if I'm following, we need the filtering in `FilteringPropertyGridUnifiedSelection` that is absent in `VirtualizedPropertyGridWithDataProvider` in order for the null value toggle to work, but currently using `FilteringPropertyGridUnifiedSelection` will trigger the `isOverlimit` clause and not properly show the property grid?

We can't conditionally use a useEffect like `usePropertyDataProviderWithUnifiedSelection` and to me it makes no sense to pass a prop called `disableUnifiedSelection` into a component that has `WithUnifiedSelection` in its name.

If we need a version without the unified selection we need a component called `FilterPropertyGrid` that is just that and this `FilterPropertyGridWithUnifiedSelection` can wrap that with `usePropertyDataProviderWithUnifiedSelection`, then in the PropertyGrid.tsx we can use the component corresponding to whether or not we want unified selection

cortneysmithbentley: Wouldn't it make more sense to just rename FilteringPropertyGridWithUnifiedSelection to FilteringPropertyGrid and keep the rest as is (passing in disableUnifiedSelection)?

cortneysmithbentley: Just pushed the change I suggested.

cortneysmithbentley: This is pushed now

jason-crow (approved):

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -112,26 +118,53 @@ implements IPropertyDataProvider, IDisposable {
...
+ */
 export const FilteringPropertyGridWithUnifiedSelection = (
-  props: FilteringPropertyGrid
+  props: FilteringPropertyGridProps
 ) => {
   const localizations = useMemo(() => ({
     tooManySelected: PropertyGridManager.translate(
       "context-menu.selection.too-many-elements-selected"
     ),
-    noneSelected: PropertyGridManager.translate(
```
aruniverse, line 128: why was this removed? isnt this needed for ui-1?

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -71,11 +77,11 @@ export class NonEmptyValuesPropertyDataFilterer extends PropertyRecordDataFilter
 }
 
 type CustomPropertyDataProvider<TPropertyData> = IDisposable &
-Omit<IPropertyDataProvider, "getData"> & {
-  getData: () => Promise<TPropertyData>;
-};
+  Omit<IPropertyDataProvider, "getData"> & {
```
aruniverse, line 80: these indentations are probably gonna fail the linter

common/changes/@itwin/property-grid-react/fix-empty-props-toggle-broken-multiselect_2022-08-03-20-00.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/property-grid-react",
+      "comment": "Using disableUnifiedSelection property to get FilteringPropertyGrid to work with multiple elements",
```
aruniverse, line 5: is this comment relevant now?

aruniverse: might also want to update the pr description as necessary

cortneysmithbentley: good catch, i'll fix that

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
cortneysmithbentley: I couldn't find any reference to it in the code. If it's referenced in some other way where it wouldn't show up in a search I can add it back.

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
cortneysmithbentley: Ran the linter and fixed issues.

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
cortneysmithbentley: Verified with the other design review devs about removing this string. They said it shouldn't matter for UI 1.0 and since it's not referenced in any of the files it should be removed.

PR #304, opened 2022-08-05
[Grouping&Mapping] Hilite optimization

Guillar1:
Optimized the hiliting by querying for the class name rather then using the expensive getProps.


packages/itwin/grouping-mapping-widget/src/widget/components/viewerUtils.tsx
```diff
@@ -371,49 +372,30 @@ export const getHiliteIds = async (
...
+  }
+  const queryWithIdAndECClassName = `SELECT q.ECInstanceId, ec_classname(e.ECClassId) FROM (${query}) q JOIN BisCore.Element e on e.ECInstanceId = q.ECInstanceId`;
+
+  const rowIterator = iModelConnection.query(queryWithIdAndECClassName, undefined, {
+    rowFormat: QueryRowFormat.UseECSqlPropertyIndexes,
+  });
+
+  const keys: InstanceKey[] = [];
+
+  while (true) {
```
MindaugasButkus, line 391: Rewrite this using the `for await ... of ...` syntax.

Should be something like:
```
for await (const value of rowIterator) {
    keys.push({ id: value[0], className: value[1] });
}
```

MindaugasButkus (changes requested):

packages/itwin/grouping-mapping-widget/src/widget/components/viewerUtils.tsx
Guillar1: switched to ```for await``` syntax.

MindaugasButkus (approved):

Guillar1:
@aruniverse Can you review?

aruniverse (approved):

PR #313, opened 2022-08-22
[Grouping&Mapping] Minor edit to group coloring wording

arinsasmal:


packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -411,7 +411,7 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
               </Button>
               <ButtonGroup className="gmw-toolbar-buttons">
                 <ToggleSwitch
-                  label="Color by Group"
+                  label="Group by Color"
```
MindaugasButkus, line 414: I don't understand this change. Could you provide the reasoning. I thought that this widget was coloring elements by the group that they fall into. Is this creating a Group based on the element's color?

arinsasmal: The specific change was a requested by Vikte in my Azure DevOps, so I just implemented the change as per the instructions.

MindaugasButkus: We had a discussion with Vikte. It turns out that the current wording was misunderstood. This PR should be abandoned.

PR #322, opened 2022-09-13
[G&M] Context Custom UI

bingjiez:
Context UI provider with callback capability.
image


packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -550,58 +560,85 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
+                                      ))
+                                  }
+                                >
+                                  <SvgEdit
+                                    style={{
+                                      width: "16px",
+                                      height: "16px",
+                                      margin: "0 8px 0 0",
+                                    }}
+                                  />
```
Guillar1, line 592: This should be passed in through the icon prop.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -16,20 +26,50 @@ export interface GroupingMappingCustomUI {
...
+ */
+export interface ContextCustomUI extends IGroupingMappingCustomUI {
+  /**
+   * UI type.
+   */
+  type: GroupingMappingCustomUIType.CONTEXT;
+  /**
+   * User defined component for UI interaction with grouping mapping widget.
+   */
+  uiComponent?: () => JSX.Element;
```
Guillar1, line 46: It would be nice if you can also allows props here as well.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -16,20 +26,50 @@ export interface GroupingMappingCustomUI {
...
+  /**
+   * Callback function for non UI interaction with grouping mapping widget.
+   */
+  callback?: (groupId: string, mappingId: string, iModelId: string) => void;
```
Guillar1, line 50: The name `onClick` for this prop seems more appropriate then `callback` because this fires when you click on the custom context menu item.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+const groupsFactory = (): GroupCollection => ({
+  groups: Array.from(
+    { length: faker.datatype.number({ min: 3, max: 5 }) },
+    (_, index) => ({
+      id: `${faker.datatype.uuid()}`,
+      groupName: `mOcKgRoUp${index}`,
+      description: `mOcKgRoUpDeScRiPtIoN${index}`,
+      query: `mOcKgRoUpQuErY${index}`,
+      deleted: false,
```
MindaugasButkus, line 47: Groups do not have `deleted` flags.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+jest.mock("@itwin/appui-react", () => ({
+  ...jest.requireActual("@itwin/appui-react"),
+  useActiveIModelConnection: () => connectionMock.object,
+}));
+
+const server = setupServer();
+
+beforeAll(async () => {
+  // This is required by the i18n module within iTwin.js
+  (global as any).XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest; // eslint-disable-line @typescript-eslint/no-var-requires
```
MindaugasButkus, line 79: I think the localization module can be replaced with a mock implementation. I think it would be better than this hacky approach.

MindaugasButkus: `NoRenderApp.startup()` accepts an object with this property:
```
/** If present, supplies the [[Localization]] for this session. Defaults to [ITwinLocalization]($i18n). */
  localization?: Localization;
```

(https://github.com/iTwin/itwinjs-core/blob/34654961639e0c6d20537c1c411ddf1caf136f7b/core/frontend/src/IModelApp.ts#L109)

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+const selectionScopesManagerMock = moq.Mock.ofType<SelectionScopesManager>();
+
+jest.mock("@itwin/appui-react", () => ({
+  ...jest.requireActual("@itwin/appui-react"),
+  useActiveIModelConnection: () => connectionMock.object,
+}));
+
+const server = setupServer();
+
+beforeAll(async () => {
```
MindaugasButkus, line 77: This `beforeAll` will probably run before ALL tests in this test suite. I'd say it should be scoped to the specific component being tested. Use `describe` to reduce the scope. Also, mocks should be set up in that scope as well.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+afterEach(() => {
+  server.resetHandlers();
+});
+
+describe("Groupings View with default UIs", () => {
+  it("List all groups and click on add group and edit group buttons", async () => {
+    // Arange
+    const mockGroups = groupsFactory();
+    server.use(
```
MindaugasButkus, line 123: You should be mocking the insights client and not checking each request. What if the internals of insights client would change? Then you would need to update all of the tests.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    // Assert
+    const addButton = screen.getAllByTestId("gmw-add-group-button");
+    expect(addButton).toHaveLength(1);
+
+    // Click on 'Add Group' button
+    await user.click(addButton[0]);
+
+    // Should have three menu items
+    const AddMenuItems = screen.getAllByTestId("gmw-add-group-menu-item");
```
MindaugasButkus, line 148: Local variables or consts should start with a lowercase letter.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -4,10 +4,20 @@
 *--------------------------------------------------------------------------------------------*/
 import type { ReactElement } from "react";
 
-/**
- * Customized UI type definition for grouping mapping widget.
- */
-export interface GroupingMappingCustomUI {
+export enum GroupingMappingCustomUIType {
+  GROUP = "group",
```
MindaugasButkus, line 8: Use PascalCase for enum values.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -16,20 +26,68 @@ export interface GroupingMappingCustomUI {
...
+ */
+export interface GroupingCustomUI extends IGroupingMappingCustomUI {
+  /**
+   * UI type.
+   */
+  type: GroupingMappingCustomUIType.GROUP;
+  /**
+   * Custom UI Component to build query interactively. Refer to SearchGroupingCustomUI/ManualGroupingCustomUI.
+   */
+  uiComponent: (props: GroupingCustomUIProps ) => JSX.Element;
```
MindaugasButkus, line 64: Formatting - trailing whitespace after the prop type.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -16,20 +26,68 @@ export interface GroupingMappingCustomUI {
...
+}
+
+/** Custom UI Component props definitions  */
```
MindaugasButkus, line 67: Missing period at the end of sentence.

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -550,65 +560,81 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
+                                  disabled={isLoadingQuery}
+                                  data-testid="gmw-context-menu-item"
+                                  subMenuItems={
+                                    (groupUIs.length > 0
+                                      ? groupUIs
+                                      : defaultUIMetadata)
+                                      .map((p) => (
+                                        <MenuItem
+                                          className='gmw-menu-item'
+                                          data-testid="gmw-edit-menu-item"
```
MindaugasButkus, line 577: Non unique test ID if multiple groupUIs are provided.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
bingjiez: This is not a UI component prop. The uiComponent is a peer property in the same class so I think keep it as call back would be nice to match the requirement? But yes, for such external props, naming should be discussed. How do you think?@arnobmallickbsw

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
bingjiez: I used an empty localization. Not sure this will work for other widgets @Guillar1

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
bingjiez: Wouldn't it be better to be kept in beforeAll so that it only runs once for entire suite? Each test will need to render the components within widget and will have interfaction with viewer

MindaugasButkus: Having global setups can make it harder to debug other tests. I think you can run into some weird race conditions if you keep it this way. If you really want to have a global setup for all tests - move it into a separate file. The require it for the test script in package.json (e.g. https://github.com/iTwin/insights-client/blob/main/package.json, see `--require` for `test:integration`).

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
MindaugasButkus: I think `callback` is fine. It's more generic than `onClick`, this would still give us flexibility to change the actual event when this is fired. Maybe we will need to change it to some different event, e.g. `onChange` for some dropdown selection. Then we would need to rename this public property from `onClick` to `onChange`. I'd say keep the `callback` so that we do not need to change it later.

common/changes/@itwin/grouping-mapping-widget/gm-context-uiProvider_2022-09-21-20-18.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "add context custom UI support ",
```
MindaugasButkus, line 5: Change log comments should start with a capital letter. Also, drop the trailing space.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -16,20 +26,68 @@ export interface GroupingMappingCustomUI {
...
+ */
+export interface ContextCustomUI extends IGroupingMappingCustomUI {
+  /**
+   * UI type.
+   */
+  type: GroupingMappingCustomUIType.CONTEXT;
+  /**
+   * User defined component for UI interaction with grouping mapping widget.
+   */
+  uiComponent?: (props: ContextCustomUIProps) => JSX.Element;
```
MindaugasButkus, line 46: I wonder if it would be better to move `uiComponent` to the base interface with a generic type for `props`. Also, why is this optional?

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
MindaugasButkus: I don't see a reason why it shouldn't work.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+afterEach(() => {
+  server.resetHandlers();
+});
+
+describe("Groupings View with default UIs", () => {
+  it("List all groups and click on add group and edit group buttons", async () => {
+    // Arange
+    const mockGroups = groupsFactory();
+    mockMappingClient
+      .setup(async (x) => x.getGroups(moq.It.isAnyString(), moq.It.isAnyString(), moq.It.isAnyString()))
```
MindaugasButkus, line 121: I think you can be a bit more specific with parameters on this setup.

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+const connectionMock = moq.Mock.ofType<IModelConnection>();
+const selectionManagerMock = moq.Mock.ofType<SelectionManager>();
+const selectionScopesManagerMock = moq.Mock.ofType<SelectionScopesManager>();
+
+jest.mock("@itwin/appui-react", () => ({
+  ...jest.requireActual("@itwin/appui-react"),
+  useActiveIModelConnection: () => connectionMock.object,
+}));
+
+const server = setupServer();
```
MindaugasButkus, line 74: Do you still need this `server`, when the `insights-client` is mocked?

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
```diff
@@ -4,10 +4,20 @@
...
+  CONTEXT = "Context"
+}
+
+export type GroupingMappingCustomUI = GroupingCustomUI | ContextCustomUI;
+
+/** Custom UI Definitions */
+
+export interface IGroupingMappingCustomUI {
+  /**
+   * Custom UI type: grouping or context
```
MindaugasButkus, line 18: Documentation / comment sentences should end with periods.

MindaugasButkus: What if new type is added to the enum? Will the developer remember to update the documentation? It would be better to link to the enum itself with `see GroupingMappingCustomUIType` or something.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
MindaugasButkus, line 46: It would be great to have an example of such `ContextCustomUI`. Maybe you could try and think of something?

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -84,6 +87,7 @@ enum GroupsView {
   MODIFYING = "modifying",
   ADD = "ADD",
   PROPERTIES = "properties",
+  CUSTOM = "custom",
```
MindaugasButkus, line 90: This enum is inconsistent. String values should be formatted similarly (`ADD` is all uppercase while everything else is lowercase). Also, Change enum value names to PascalCasing.

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -140,6 +147,7 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
 
   const [queryGenerationType, setQueryGenerationType] =
     useState<string>("Selection");
+  const [selectedContextCustomUI, setSelectedContextCustomUI] = useState<ContextCustomUI>();
```
MindaugasButkus, line 150: What is the default value for this state?

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -425,25 +433,27 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
+                  (groupUIs.length > 0
+                    ? groupUIs
                     : defaultUIMetadata)
                     .map((p) => (
                       <MenuItem
                         key={p.name}
                         onClick={() => addGroup(p.name)}
                         icon={p.icon}
                         className='gmw-menu-item'
+                        data-testid={`gmw-add-${p.name}`}
```
MindaugasButkus, line 448: Are there any restrictions of what can be in the `data-testid` attribute? Can it contain spaces? Can the `name` contain spaces? Are all characters allowed in both of them?

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -548,67 +558,83 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
-                              </MenuItem>,
-                            ]}
+                                  icon={<SvgDelete />}
+                                  data-testid="gmw-context-menu-item"
+                                >
+                                  Remove
+                                </MenuItem>,
+                              ].concat(
+                                contextUIs.map((p) => {
+                                  return <MenuItem
```
MindaugasButkus, line 611: Reduce `<MenuItem />` duplication. Create an array with all the required values and use `map` to create all the `MenuItem` elements.

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -548,67 +558,83 @@ export const Groupings = ({ mapping, goBack }: GroupsTreeProps) => {
...
                               styleType='borderless'
+                              data-testid="gmw-more-button"
                             >
-                              <SvgMore
-                                style={{
-                                  width: "16px",
-                                  height: "16px",
-                                }}
-                              />
+                              <SvgMore />
```
MindaugasButkus, line 637: Why was the style override removed? Does it no longer require it? How did this change visually?

packages/itwin/grouping-mapping-widget/src/test/test-utils.tsx
```diff
@@ -0,0 +1,77 @@
...
+    );
+  }
+
+  public static async initializeUiFramework(_?: IModelConnection) {
+    const store = this.createAppStore();
+
+    await UiFramework.initialize(store);
+
+    // Set the iModelConnection in the Redux store
+    // Can not get it to function as intended. Store does not update with the iModel connection
```
MindaugasButkus, line 70: What do you mean "you cannot get it to function as intended"? How was it supposed to work, and how did it actually work?

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    const contextMenuItems = screen.getAllByTestId("gmw-context-menu-item");
+    expect(contextMenuItems).toHaveLength(3);
+    expect(contextMenuItems[0]).toHaveTextContent("Edit");
+    expect(contextMenuItems[1]).toHaveTextContent("Properties");
+    expect(contextMenuItems[2]).toHaveTextContent("Remove");
+
+    // Hover on 'Edit'
+    await user.hover(contextMenuItems[0]);
+
+    // Should have 3 sub menu items
```
MindaugasButkus, line 177: Do I understand correctly that this is a temporary workaround until we have the group query creation metadata stored as a comment on the query itself?

packages/itwin/grouping-mapping-widget/README.md
```diff
@@ -55,11 +60,48 @@ You can define your UI provider as follows in the GroupingMappingProvider:
...
     },
   ],
 });
 ```
+
+### Context Custom UI
+
+You can also add custom UI in context menu for each group. This could be UI driven interaction or simply a callback function to execute your own code.
+
+You can pass in `iModelId`, `mappingId` and `groupId` in `ContextCustomUIProps` to assist customized user experience.
```
MindaugasButkus, line 77: The user has to pass these IDs into the custom ui component, or does the widget pass them into the component that user created?

common/config/rush/pnpm-lock.yaml
```diff
@@ -295,14 +302,21 @@ importers:
...
       react-dom: ^17.0.2
       react-redux: ^7.2.0
       react-table: ^7.7.0
       redux: ^4.1.0
       rimraf: ^3.0.2
       simple-react-validator: ^1.6.1
+      ts-jest: ^27.1.3
+      typemoq: ^2.1.0
       typescript: ~4.3.0
+      xmlhttprequest: ^1.8.0
```
MindaugasButkus, line 319: `xmlhttprequest` is probably no longer used, since you switched to a different localization override.

packages/itwin/grouping-mapping-widget/README.md
bingjiez: the widget has passed in. Bad description, I will change it to 'You can use ...

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
bingjiez: yes

packages/itwin/grouping-mapping-widget/src/test/test-utils.tsx
bingjiez: Copy pasted the test utils. Will dig into it later

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
bingjiez: the name has a validator not allowing spaces in the UI

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
bingjiez: I looked into jest configuration. The beforeAll is scoped to each test file so there won't be a global race condition. There is a jest configuration 'globalSetup' which will be called before all test suites but it is run before test-environment is setup. I tried it but a lot of setups here involves jsdom and it doesn't work. I submitted an issue in Jest repository and wait for anybody to give me a solution. Right now, I don't think it worths the time to dig deeper into this problem. We will have more bandwidth in Q4 for GM testing

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
bingjiez: user can just pass in a callback without uiComponent

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
bingjiez: Yeah the style is defined in the IconButton. The visual won't be affected and it's more consistent

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
bingjiez: The onClick callback for each MenuItem is quite different and the first one has sub menu items. I think it is more clear to keep it like this now

packages/itwin/grouping-mapping-widget/src/test/GroupingMappingCustomUI.test.tsx
Guillar1: It should work. When I wrote the tests I didn't realize that you can supply an empty localization through startup.

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupingMappingCustomUI.ts
Guillar1: IMO, I feel that it's too generic. It's not obvious right away when this actually gets called. If the behavior changes I believe it should be obvious and not keep the developer wondering why the behavior just randomly changed.

MindaugasButkus (approved):

aruniverse (approved):

PR #326, opened 2022-09-23
Datasets can be updated in the Reports page. Reports tiles are selectable

Simonas-A:
Search button moved to the right


packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.scss
```diff
@@ -0,0 +1,29 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+@import "~@itwin/itwinui-css/scss/variables";
+
+.iui-message {
+  margin: auto;
+}
+
+.extraction-status {
```
MindaugasButkus, line 11: Prefix classes with the `rcw-` prefix.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.scss
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+@import "~@itwin/itwinui-css/scss/variables";
+
+.iui-message {
```
MindaugasButkus, line 7: Do not override global styles. What if another widgets in the same applications use the `iui-message` class?

packages/itwin/reports-config-widget/src/widget/components/SearchBar.scss
```diff
@@ -20,8 +21,13 @@
...
     width: 20%;
   }
+
   0% {
     opacity: 1;
     width: 100%;
   }
 }
+
+.search-button {
```
MindaugasButkus, line 31: Prefix classes with `rcw-`.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -181,45 +232,66 @@ export const Reports = () => {
...
+                            >
+                              {ReportsConfigWidget.localization.getLocalizedString(
+                                "ReportsConfigWidget:Remove"
+                              )}
+                            </MenuItem>,
+                          ]}
+                        >
+                          <IconButton styleType="borderless">
+                            <SvgMore
+                              style={{
```
MindaugasButkus, line 287: Does this style really override anything? In other PR this kind of style was removed for one icon.

Simonas-A: With this styleType the rendered button receives an additional css class `iui-borderless`. Therefore, the difference is visible.
image
image

Guillar1: This icon should be passed through the `icon` prop in this component without the style override.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,188 @@
...
+        continue;
+      }
+
+      const states: ExtractorState[] = [];
+      for (const runId of runs) {
+        const state = stateByRunId.get(runId);
+        if (state) {
+          states.push(state);
+        } else {
+          const state = await this.getSingleState(runId, extractionClientApi, accessToken);
```
Guillar1, line 53: We should avoid shadowing variables.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -107,6 +114,28 @@ export const Reports = () => {
...
+          setJobStarted(false);
+          bulkExtractor.clearJobs();
+        }
+        setExtractionStates(states);
+      }, 5000)
+      return () => clearInterval(interval);
+    }
+    setTimeout(() => {
+      setExtractionStates(new Map<string, ExtractionStates>());
+    }, 5000);
```
Guillar1, line 135: This setTimeout can be moved into its own useEffect with its own clearTimeout cleanup.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -107,6 +114,28 @@ export const Reports = () => {
     [reports, searchValue]
   );
 
+  useEffect(() => {
+    if (jobStarted) {
+      const interval = setInterval(async () => {
```
Guillar1, line 119: Please use the window object to call these timer functions.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
MindaugasButkus: My comment was for the `style` not `styleType`.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
MindaugasButkus: Any reason why they should be called through the `window` object?

Guillar1: It returns Node.JS types if you don't use the window object. Trying to stay clear of Node types is usually less risky due to some bundlers like Webpack 5 not supporting Node types by default any longer.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.scss
Simonas-A: Prefix has been added

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.scss
Simonas-A: `.iui-message` style override has been deleted

packages/itwin/reports-config-widget/src/widget/components/SearchBar.scss
Simonas-A: Prefix has been added

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: style for SVG size has been removed

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Variable has been renamed

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `setTimeout` has been moved into new useEffect

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `window` object is now used to call timer functions

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -181,45 +240,62 @@ export const Reports = () => {
...
+                              icon={<SvgDelete />}
+                            >
+                              {ReportsConfigWidget.localization.getLocalizedString(
+                                "ReportsConfigWidget:Remove"
+                              )}
+                            </MenuItem>,
+                          ]}
+                        >
+                          <IconButton styleType="borderless">
+                            <SvgMore
```
MindaugasButkus, line 294: Close the element on the same line (move `/>` up).

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,188 @@
...
+export type ReportMappingType = CreateTypeFromInterface<ReportMapping>;
+
+export type ReportMappingAndMapping = ReportMappingType & {
+  mappingName: string;
+  mappingDescription: string;
+  iModelName: string;
+};
+
+export default class BulkExtractor {
+  reportStates = new Map<string, ExtractionStates>();
```
MindaugasButkus, line 27: Mark all fields as either `private` or `public`. I think these should be `private`. Then, all private fields should be prefixed with an underscore `_`.

Simonas-A: Prefixes for fields has been added

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `/>` has been moved up

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+      case ExtractionStates.Failed:
+        timer = window.setTimeout(() => setFadeOut(true), 5000);
+    }
+    return () => clearTimeout(timer);
+  }, [state, setExtractionState]);
+
+  switch (state) {
+    case ExtractionStates.Starting:
+      return (
+        <div
```
MindaugasButkus, line 59: Maybe the contents of each `case` could be moved into separate files? That would make the `ExtractionStatus` component smaller and more readable.

Move all of these components into a subdirectory.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  useEffect(() => {
+    let timer: number;
+    switch (state) {
+      case ExtractionStates.Succeeded:
+      case ExtractionStates.Failed:
+        timer = window.setTimeout(() => setFadeOut(true), 5000);
+    }
+    return () => clearTimeout(timer);
```
MindaugasButkus, line 53: I think you need to use `clearTimeout` from the windows object here as well.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,190 @@
...
+        reportId
+      );
+      const iModelClientOptions: IModelsClientOptions = {
+        api: { baseUrl: generateUrl(Constants.api.baseUrl, apiContext.baseUrl) },
+      };
+
+      const iModelsClient: IModelsClient = new IModelsClient(iModelClientOptions);
+      const authorization =
+        AccessTokenAdapter.toAuthorizationCallback(accessToken);
+      const iModelNames = new Map<string, string>();
```
MindaugasButkus, line 153: Move the `iModelNames` cache to a private field on BulkExtractor - this will save some calls to iModels API when multiple fetches are made for report mappings. iModelNames do not change that often.

MindaugasButkus: And then write a wrapper function for that cache which would get the value from cache (Map) if it exists, and if not - get ir from iModels API and store it in the cache.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+    return undefined;
+  }
+
+  private async fetchReportMappings(
+    reportId: string,
+    apiContext: ReportsApiConfig
+  ): Promise<ReportMappingAndMapping[]> {
+    try {
+      const reportsClientApi = new ReportsClient(
```
MindaugasButkus, line 135: Clients should be stored in a private field and reused.

Whenever baseUrl changes the BulkExtractor object should be recreated, because `_reportRunIds` becomes invalid.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,190 @@
...
+  public async startJobs(reportIds: string[]) {
+    const extractionClientApi = new ExtractionClient(
+      generateUrl(REPORTING_BASE_PATH, this._apiConfig.baseUrl)
+    );
+    const accessToken = await this._apiConfig.getAccessToken();
+
+    const extractionByIModel = new Map<string, string>();
+    for (const reportId of reportIds) {
+      const IModels = (await this.fetchReportMappings(reportId, this._apiConfig)).map((m) => m.imodelId);
+      const uniqueIModels = Array.from(new Set(IModels));
```
MindaugasButkus, line 98: `[...new Set(iModels)]`.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  public async startJobs(reportIds: string[]) {
+    const extractionClientApi = new ExtractionClient(
+      generateUrl(REPORTING_BASE_PATH, this._apiConfig.baseUrl)
+    );
+    const accessToken = await this._apiConfig.getAccessToken();
+
+    const extractionByIModel = new Map<string, string>();
+    for (const reportId of reportIds) {
+      const IModels = (await this.fetchReportMappings(reportId, this._apiConfig)).map((m) => m.imodelId);
```
MindaugasButkus, line 97: Local variables should start with lowercase letters.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+    return stateByReportId;
+  }
+
+  public clearJobs() {
+    this._reportRunIds = new Map<string, string[]>();
+  }
+
+  private getFinalState(states: ExtractorState[]) {
+    const responseState = states.includes(ExtractorState.Failed) ? ExtractionStates.Failed :
```
MindaugasButkus, line 72: There is no benefit of having `responseState` variable.

Replace it with `if`s that return immediately if the condition is true.

If no conditions are met - return `Failed`.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+import "./ExtractionStatus.scss";
+import {
+  SvgStatusError,
+  SvgStatusPending,
+  SvgStatusPendingHollow,
+  SvgStatusSuccess,
+} from "@itwin/itwinui-icons-color-react";
+import { ReportsConfigWidget } from "../../ReportsConfigWidget";
+
+export enum ExtractionStates {
```
MindaugasButkus, line 17: You might want to extend extraction status enum that Extraction API returns. Take a look at this: https://blog.logrocket.com/extend-enums-typescript/

This might make the conversions easier.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -181,45 +239,61 @@ export const Reports = () => {
...
-                            style={{
-                              width: "16px",
-                              height: "16px",
-                            }}
-                          />
-                        </IconButton>
-                      </DropdownMenu>
+                    selected={selectedReports.some((r) => report.id === r.id)}
+                    onClick={(e) => {
+                      if (e?.target?.className?.toString().startsWith("rcw-horizontal-tile-container")) {
```
MindaugasButkus, line 244: You should not rely on the order of class names.

MindaugasButkus, line 244: This is a big event handler - move it to a separate function.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -107,6 +114,35 @@ export const Reports = () => {
...
+        if (Array.from(states.values())
+          .filter((x) => x === ExtractionStates.Succeeded ||
+            x === ExtractionStates.Failed ||
+            x === ExtractionStates.None)
+          .length === states.size) {
+          setJobStarted(false);
+          bulkExtractor.clearJobs();
+        }
+        setExtractionStates(states);
+      }, 5000);
```
MindaugasButkus, line 130: Try to use less magic numbers. Move it to some constants file.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setFadeOut(false);
+    }
+  };
+
+  useEffect(() => {
+    let timer: number;
+    switch (state) {
+      case ExtractionStates.Succeeded:
+      case ExtractionStates.Failed:
+        timer = window.setTimeout(() => setFadeOut(true), 5000);
```
MindaugasButkus, line 51: Same magic number here. You will definitely benefit from the constants file.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: Each `ExtractionState` has been moved into a separate file in `ExtractionStates` directory

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: `clearTimeout` is now called from the window object

packages/itwin/reports-config-widget/src/widget/components/HorizontalTile.tsx
```diff
@@ -6,51 +6,31 @@ import type { ReactNode } from "react";
...
-            title={subtextToolTip}
-            variant="small"
-          >
-            {subText}
-          </Text>
-        )}
+    <div className={classNames("rcw-horizontal-tile-container", { "rcw-horizontal-tile-selected": props.selected })} onClick={props.onClick} data-testid="rcw-horizontal-tile">
+      <div className="rcw-body-container">
+        <div className="rcw-body">
+          <Text className={classNames("rcw-body-text", { "iui-anchor": props.onClickTitle })} onClick={props.onClickTitle} variant="body" title={props.titleTooltip}>{props.title}</Text>
```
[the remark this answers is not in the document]
Simonas-A: Code in `HorizontalTile` has been reformatted. The lines are now shorter

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `iModelNames` map has been moved to a private field in `BulkExtractor`

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: All clients were moved to a private field and are initialized in constructor

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Unnecessary conversion removed

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `IModels` variable renamed to `iModels`

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `responseState` variable has been removed. State is returned immediately

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `split(" ").includes` has been added to check every class in element

Simonas-A: Event handler has been moved into function `onClickTile`

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `Constants.ts` file has been added with constant variable `STATUS_CHECK_INTERVAL`

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: `STATUS_CHECK_INTERVAL` is now used from constants file

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: `ExtractionStates` enum has not been extended as conversion from `ExtractorState` is being done in `getFinalState` in `BulkExtractor`

packages/itwin/reports-config-widget/src/widget/components/ExtractionStates/QueuedExtractionState.tsx
```diff
@@ -0,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import React from "react";
+import { SvgStatusPending } from "@itwin/itwinui-icons-color-react";
+import { ReportsConfigWidget } from "../../../ReportsConfigWidget";
+
+export const QueuedExtractionState = () => {
+  return (
```
Guillar1, line 10: Can be made into an implicit return.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -107,6 +115,49 @@ export const Reports = () => {
...
+    if (!jobStarted) {
+      const timeout = window.setTimeout(() => {
+        setExtractionStates(new Map<string, ExtractionStates>());
+      }, STATUS_CHECK_INTERVAL);
+      return () => window.clearTimeout(timeout);
+    }
+    return;
+  }, [jobStarted]);
+
+  function onClickTile(e: any, report: Report) {
```
Guillar1, line 146: Make this into an arrow function.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+  GetSingleIModelParams,
+} from "@itwin/imodels-client-management";
+import { Constants, IModelsClient } from "@itwin/imodels-client-management";
+import { AccessTokenAdapter } from "@itwin/imodels-access-frontend";
+import type { ReportsApiConfig } from "../context/ReportsApiConfigContext";
+import { ExtractionStates } from "./ExtractionStatus";
+
+export type ReportMappingType = CreateTypeFromInterface<ReportMapping>;
+
+export type ReportMappingAndMapping = ReportMappingType & {
```
MindaugasButkus, line 19: Why does this need to be a type? Why can't it be a simple interface?

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+  iModelName: string;
+};
+
+export default class BulkExtractor {
+  private _reportRunIds = new Map<string, string[]>();
+  private _reportsClientApi: ReportsClient;
+  private _mappingsClientApi: MappingsClient;
+  private _iModelsClient: IModelsClient;
+  private _extractionClientApi: ExtractionClient;
+  private _accessToken: Promise<string>;
```
MindaugasButkus, line 31: This should not be a promise. This should be a function which returns a promise.
```
private _accessToken: () => Promise<string>;
```

MindaugasButkus: Your implementation would be fine if AccessTokens would never expire, but they do expire and it would break this component.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+      stateByReportId.set(reportId, finalState);
+    }
+    return stateByReportId;
+  }
+
+  public clearJob(reportId: string) {
+    this._reportRunIds.delete(reportId);
+  }
+
+  private getFinalState(states: ExtractorState[]) {
```
MindaugasButkus, line 77: Make `getFinalState` static since it does not use any of the fields of the object.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
```diff
@@ -0,0 +1,70 @@
...
+      case ExtractionStates.Succeeded:
+      case ExtractionStates.Failed:
+        timer = window.setTimeout(() => setFadeOut(true), STATUS_CHECK_INTERVAL);
+    }
+    return () => window.clearTimeout(timer);
+  }, [state]);
+
+  switch (state) {
+    case ExtractionStates.Starting:
+      return StartingExtractionState();
```
MindaugasButkus, line 56: These are react components - switch them to jsx syntax.
```
<StartingExtractionState />
<SucceededExtractionState fadeOut={fadeOut} />
... etc
```

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -181,45 +254,52 @@ export const Reports = () => {
...
+                    actionGroup={
+                      <div className="rcw-button-container">
+                        <ExtractionStatus
+                          state={extractionStates.get(report.id) ?? ExtractionStates.None}
+                          clearExtractionState={() => {
+                            extractionStates.delete(report.id);
+                            bulkExtractor.clearJob(report.id);
+                          }}
+                        >
+                          <DropdownMenu
```
MindaugasButkus, line 268: `DropdownMenu` should not be a child of `ExtractionStatus`. `actionGroup` should container `ExtractionStatus` OR `DropdownMenu`.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStates/SucceededExtractionState.tsx
```diff
@@ -0,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import React from "react";
+import { SvgStatusSuccess } from "@itwin/itwinui-icons-color-react";
+import { ReportsConfigWidget } from "../../../ReportsConfigWidget";
+import { ANIMATION_DELAY, ANIMATION_DURATION } from "../Constants";
+
+export const SucceededExtractionState = (animation: string, onAnimationEnd: () => void) => {
```
MindaugasButkus, line 10: React component should accept a single props argument.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+        reportId
+      );
+      const authorization =
+        AccessTokenAdapter.toAuthorizationCallback(await this._accessToken);
+
+      const reportMappingsAndMapping = await Promise.all(
+        reportMappings.map(async (reportMapping) => {
+          const iModelId = reportMapping.imodelId;
+          let iModelName = "";
+          const mapping = await this._mappingsClientApi.getMapping(
```
MindaugasButkus, line 154: A single `startJobs` call for multiple reports might result in multiple `getMapping` requests for the same mapping.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+    try {
+      const response = await this._extractionClientApi.getExtractionStatus(accessToken, runId);
+      return response.state;
+    } catch (error: any) {
+      handleError(error.status);
+    }
+    return ExtractorState.Failed;
+  }
+
+  public async startJobs(reportIds: string[]) {
```
MindaugasButkus, line 103: Create a separate method that would start extraction for a single reportId and would return an array of runIds if that report has data from multiple iModels.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
MindaugasButkus: The wrapper function for working with the iModelNames cache was not added yet.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,186 @@
...
+        reportMappings.map(async (reportMapping) => {
+          const iModelId = reportMapping.imodelId;
+          let iModelName = "";
+          const mapping = await this._mappingsClientApi.getMapping(
+            await this._accessToken,
+            iModelId,
+            reportMapping.mappingId
+          );
+          if (this._iModelNames.has(iModelId)) {
+            iModelName = this._iModelNames.get(iModelId) ?? "";
```
MindaugasButkus, line 160: If the Map has iModelId, then it will have a valid value. Change to `!` instead of null coalescing.

packages/itwin/reports-config-widget/src/widget/components/ReportMappings.tsx
```diff
@@ -59,9 +59,7 @@ enum ReportMappingsView {
 }
 
 const fetchReportMappings = async (
-  setReportMappings: React.Dispatch<
-  React.SetStateAction<ReportMappingAndMapping[]>
-  >,
+  setReportMappings: React.Dispatch<React.SetStateAction<ReportMappingAndMapping[]>>,
```
MindaugasButkus, line 62: There is no point of having such a long type. Change to `(x: ReportMappingAndMapping[]) => void`

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Function was refactored and iModelNames is no longer used.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: Typescript does not support simple enum extensions. Extending them will cause more problems.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStates/QueuedExtractionState.tsx
Simonas-A: All `ExtractionState` components now uses implicit return

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Interface is now being used instead of type

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `_accesstToken` type changed to `() => Promise<string>;`

packages/itwin/reports-config-widget/src/widget/components/ExtractionStatus.tsx
Simonas-A: Components syntax switched to jsx

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `ExtractionStatus` moved to `ReportHorizontalTile`. `DropdownMenu` is now in `actionGroup` instead of child.

packages/itwin/reports-config-widget/src/widget/components/ExtractionStates/SucceededExtractionState.tsx
Simonas-A: `ExtractionStateProps` interface added

packages/itwin/reports-config-widget/src/widget/components/ExtractionStates/FailedExtractionState.tsx
```diff
@@ -0,0 +1,32 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import React from "react";
+import { SvgStatusError } from "@itwin/itwinui-icons-color-react";
+import { ReportsConfigWidget } from "../../../ReportsConfigWidget";
+import { ANIMATION_DELAY, ANIMATION_DURATION } from "../Constants";
+
+export const FailedExtractionState = (animation: string, onAnimationEnd: () => void) => {
+
```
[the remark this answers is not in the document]
Simonas-A: Empty lines removed

packages/itwin/reports-config-widget/src/widget/components/HorizontalTile.tsx
```diff
@@ -6,51 +6,48 @@ import type { ReactNode } from "react";
...
-          <Text
-            className="body-text"
-            isMuted={true}
-            title={subtextToolTip}
-            variant="small"
-          >
-            {subText}
+      <div className="rcw-body-container">
+        <div className="rcw-body">
+          <Text className={classNames("rcw-body-text", { "iui-anchor": props.onClickTitle })}
```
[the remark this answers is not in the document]
Simonas-A: `props.onClickTitle` is now converted to boolean

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `startJobs` has been refactored. `getMapping` will not be called. Only unique `iModels` will be used

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Separate method is no longer needed as `startJobs` ensures that multiple jobs for the same iModel would not be started.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: `iModelNames` is no longer used

packages/itwin/reports-config-widget/src/widget/components/ReportMappings.tsx
Simonas-A: Type has been simplified

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: function made into arrow function. Is `useCallback` really necessary here?

Guillar1: Not necessary, disregard that. Nothing is depending on that function.

packages/itwin/reports-config-widget/src/widget/components/HorizontalTile.tsx
```diff
@@ -6,51 +6,47 @@ import type { ReactNode } from "react";
 import React from "react";
 import { Text } from "@itwin/itwinui-react";
 import "./HorizontalTile.scss";
+import classNames from "classnames";
 
-interface HorizontalTileProps {
+export interface HorizontalTileProps {
```
Guillar1, line 11: I'm not liking the any's. Can we extend this with ```React.ComponentPropsWithoutRef<"div">``` and in the container spread the props ```{...props}``` and remove the onClick. Also ```onClickTitle``` type can be ```React.MouseEventHandler```

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,155 @@
...
+      reportIModels.forEach((iModelId) => {
+        const runId = extractionByIModel.get(iModelId);
+        if (runId)
+          runs.push(runId);
+      });
+      this._reportRunIds.set(reportId, runs);
+    });
+  }
+
+  private async runExtraction(iModelId: string) {
```
MindaugasButkus, line 135: Add return types to all methods. That should make it easier to understand what is being returned. I think returning the `id` (`string | undefined`) from `runExtraction` would be enough.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -81,6 +80,9 @@ export const Reports = () => {
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [searchValue, setSearchValue] = useState<string>("");
   const [reports, setReports] = useState<Report[]>([]);
+  const bulkExtractor = useMemo(() =>
+    new BulkExtractor(apiConfig, reports.map((r) => r.id)), [apiConfig, reports]);
```
MindaugasButkus, line 84: This should be a bit easier to read. Moving dependencies to a separate line makes it easier to find them.
```
const bulkExtractor = useMemo(
  new BulkExtractor(apiConfig, reports.map((r) => r.id)),
  [apiConfig, reports]
);
```

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -81,6 +80,9 @@ export const Reports = () => {
...
+  const jobStartEvent = new BeEvent<(reportId: string) => void>();
```
MindaugasButkus, line 85: This event should be memo'ized. Currently a new event will be created for each re-render.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
```diff
@@ -143,12 +145,27 @@ export const Reports = () => {
...
-                <SearchBar
-                  searchValue={searchValue}
-                  setSearchValue={setSearchValue}
-                  disabled={isLoading}
-                />
+              <IconButton
+                title={ReportsConfigWidget.localization.getLocalizedString(
+                  "ReportsConfigWidget:UpdateDatasets"
+                )}
+                onClick={async () => {
```
MindaugasButkus, line 152: Move out into `useCallback` and make it depend on `selectedReports`, `jobStartEvent`, and `bulkExtractor`.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
```diff
@@ -0,0 +1,130 @@
...
+  report: Report;
+  onClickDelete: () => void;
+  onClickModify: () => void;
+}
+
+export const ReportHorizontalTile = (props: ReportHorizontalTileProps) => {
+  const [jobStarted, setJobStarted] = useState<boolean>(false);
+  const [extractionState, setExtractionState] = useState<ExtractionStates>(ExtractionStates.None);
+
+  if (props.jobStartEvent)
```
MindaugasButkus, line 40: This should be done in `useEffect`. The `useEffect` function should return a function that would remove the listener.

MindaugasButkus, line 40: `jobStartEvent` will never be falsy.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
```diff
@@ -0,0 +1,130 @@
...
+  SvgDelete,
+  SvgEdit,
+  SvgMore,
+} from "@itwin/itwinui-icons-react";
+import { HorizontalTile } from "./HorizontalTile";
+
+export interface ReportHorizontalTileProps {
+  onClickTitle: (e: any) => void;
+  selected: boolean;
+  setSelectedReports: (value: React.SetStateAction<Report[]>) => void;
```
MindaugasButkus, line 28: Selection should be handled by the parent component. Change this to `onSelectionChange`.
First parameter - reportId.
Second - boolean for selected or deselected.
Third - optional boolean, which would be set when second parameter is true and means that the selected report should be appended or replace the whole list.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
```diff
@@ -0,0 +1,130 @@
...
+        const state = props.bulkExtractor.getState(props.report.id);
+        if (state) {
+          setExtractionState(state);
+          if (state === ExtractionStates.Failed || state === ExtractionStates.Succeeded) {
+            setJobStarted(false);
+          }
+        }
+      }, STATUS_CHECK_INTERVAL);
+      return () => window.clearInterval(interval);
+    }
```
MindaugasButkus, line 60: If `jobStarted` is false and `interval` is valid - clearInterval.

packages/itwin/reports-config-widget/src/widget/components/HorizontalTile.tsx
MindaugasButkus: I don't think it would be as simple as spreading the props in that case. Caller could override the classNames, which would break the ui.

I agree that `any`s could be changed to some types.

MindaugasButkus: Also, do the `onClick` and `onClickTitle` even need parameters? Maybe we could get rid of them?

Guillar1: className={classNames(
       'rcw-horizontal-tile-container',
        { "rcw-horizontal-tile-selected": props.selected },
        props.className
      )}`
This would stop it from being overridden and would just extend it.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: Closing parenthesis has been moved to a new line. Dependencies left on the same line due to lint bug regarding indentation bug.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: Return types added to all methods.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `jobStartEvent` now uses `useMemo`. Should there be any dependencies?

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Simonas-A: `onClick` function moved into `updateDatasets` callback.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Simonas-A: Listener is now added inside `useEffect`. Return function for listener removal added.

Simonas-A: `if` statement removed

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Simonas-A: Interval is cleared when jobStarted is false and everytime before starting a new interval.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Simonas-A: Added `onSelectionChange` instead of `setSelectedReports`. `onSelectedChange` accepts two parameters: `reportId` and `controlPressed` - boolean. The event is then handled at the parent component. Is that correct?

packages/itwin/reports-config-widget/src/widget/components/HorizontalTile.tsx
Simonas-A: `onClick` event is now extended with `React.ComponentPropsWithoutRef<"div">`. Props are now extended.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
MindaugasButkus: Still not all methods have return types. You missed some private methods.

Simonas-A: Missing types added

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  useEffect(() => {
+    props.jobStartEvent.addListener((startedReportId: string) => {
+      if (startedReportId === props.report.id) {
+        setExtractionState(ExtractionStates.Starting);
+        setJobStarted(true);
+      }
+    });
+
+    return function cleanup() {
```
Guillar1, line 49: Turn it into an arrow function

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Guillar1: Looks good

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Guillar1: Resolved

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Guillar1: Having one return statement outside of the `if` statement should be enough.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Guillar1: No this should be enough.

packages/itwin/reports-config-widget/src/widget/components/Reports.tsx
Guillar1: Looks good.

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
```diff
@@ -0,0 +1,155 @@
...
+      this._reportStates.set(reportId, ExtractionStates.Starting);
+    }
+    const iModels = new Set(Array.from(reportIModelIds.values()).flat());
+    const extractionMap =
+      Array.from(iModels).map(async (iModel): Promise<[string, string | undefined]> => {
+        const run = await this.runExtraction(iModel);
+        return [iModel, run?.id];
+      });
+
+    const extractionByIModel = new Map<string, string | undefined>(await Promise.all(extractionMap));
```
Guillar1, line 122: I would create a new variable for the results of that `Promise.all`

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Guillar1: Confirmed.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Simonas-A: return statement has been moved outside of the if statement

packages/itwin/reports-config-widget/src/widget/components/BulkExtractor.ts
Simonas-A: New `const` variable is created before creating map

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Simonas-A: function changed to an arrow function. Should I even simplify it to: `return props.jobStartEvent.clear;` ?

Guillar1: Yeah the simplification works.

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
Guillar1: resolved

packages/itwin/reports-config-widget/src/widget/components/ReportHorizontalTile.tsx
MindaugasButkus: Doesn't `clear` remove all listeners? You need to remove only that one listener for this specific component instead of all the listeners.

Guillar1: @MindaugasButkus is right. It does clear all listeners. I am not too familiar with `BeEvent`. I am assuming there is something like `removeListener()`?

MindaugasButkus: Yes, there should be something like that and you need to have the exact same function reference that you passed into the `addListener`.

Simonas-A: Listener is created before adding or removing it from `jobStartEvent`.

Simonas-A:
@aruniverse Hello, can you please review this pull request?

MindaugasButkus (approved):

Guillar1 (approved):

aruniverse (approved):
Approving lockfile changes

PR #327, opened 2022-09-26
[ReportingWidgets] Fixed text overflow

Guillar1:
Fixed text overflow ellipses not triggering when shrinking the widgets.


MindaugasButkus:
Please provide before and after screenshots.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMapping.scss
```diff
@@ -5,6 +5,9 @@
 @import "~@itwin/itwinui-css/scss/variables";
 
 .gmw-group-mapping-container {
+  position: absolute;
```
MindaugasButkus, line 8: This looks like a hack. How does this solve a "text overflow" issue? And what was the text overflow issue?

Guillar1: The viewers core widget container has no set width so it doesn't constrain the widget to trigger `text-overflow: ellipsis`. The hackier method is to define a `min-width: 0;` in the `nz-content` class that is a part of the viewer which is what we don't want to do. Or do this which is to position the widget itself relative to the container `nz-widget-contentContainer` and have it resize accordingly. 

Before: 
image
After:
image

bingjiez (approved):

MindaugasButkus (approved):

PR #328, opened 2022-09-26
[measure-tools] add formatter tests

a-gagnon:
- Add launch configuration to be able to properly debug tests in the measure-tools package.
- Use the same symbology as IModelApp.quantityFormatter when formatting degrees in 'DMS' format. Instead of the expected prime (\u2032) and double prime ('\u2033'), they're using the apostrophe **'** and the quotation mark **"**.
- Fix minor inconsistency when formatting slope with/without ratio appended at the end.


aruniverse (approved):
only approving on behalf of the launch.json & webpack config changes; the actual measure-tools code changes should be reviewed by @bsy-nicholasw or team

PR #331, opened 2022-10-05
Infrastructure Iot - Improved view zoom behavior

AntonAI:


.editorconfig
```diff
@@ -0,0 +1,16 @@
+# Editor configuration, see https://editorconfig.org
```
aruniverse, line 1: Is this repo wide?

aruniverse (approved):

.editorconfig
AntonAI: Yes, thought there should be one, but I can remove if you don't want it there

PR #333, opened 2022-10-12
[G&M] Fixed key issue

Guillar1:
Key was not unique enough in some instances.

### DIFF
--- a/packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyUtils.ts
+++ b/packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyUtils.ts
@@ -82,7 +82,7 @@ const extractPrimitives = (
       primitiveNavigationClass,
       actualECClassName,
       parentPropertyClassName,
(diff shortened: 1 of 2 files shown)


packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyUtils.ts
```diff
@@ -82,7 +82,7 @@ const extractPrimitives = (
       primitiveNavigationClass,
       actualECClassName,
       parentPropertyClassName,
-      key: `${actualECClassName}__${newPropertyTraversal.join("_")}`,
+      key: `${parentPropertyClassName}_${actualECClassName}_${newPropertyTraversal.join("_")}`,
```
MindaugasButkus, line 85: Use a separator character that would not appear in ECName (https://www.itwinjs.org/bis/ec/ec-name/). I think a good candidate would be `|`.

Guillar1: Changed it to `|`.

MindaugasButkus (approved):

PR #336, opened 2022-10-21
[measurement-tools] Hide Measurement Tools from toolbars when view is a sheet view.

cortneysmithbentley:
Added some checks to the UI Items Provider so that the toolbar button doesn't show on a sheet view.


aruniverse (dismissed):
shouldnt this be done at the app level? 

arent there measurement tools tht make sense in sheet views like distance or angles or whatever?

cortneysmithbentley:
> 



> 

I thought so too but @JoeZman and @jason-crow believed it should be here, so I'll let them defend the decision. 

I don't know if those measurements types make sense for sheets.

jason-crow:
> 

There are measurement tools that make sense for drawings but for sheets they won't work

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -64,6 +67,10 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
       }
 
       if (toolbarOrientation === ToolbarOrientation.Vertical) {
+        // don't display measurement tools on a sheet view
+        const vp = IModelApp.viewManager.selectedView;
```
jason-crow, line 71: Don't add handling here. The tool can be added regardless of the current view its just the tool will hide itself if the current view is a sheet, which could change without this code here rerunning

cortneysmithbentley: where would I add it then?

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -64,6 +67,10 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
...
+        if (vp && ViewHelper.isSheetView(vp))
+          return [];
         return [
           ToolbarItemUtilities.createGroupButton(
```
jason-crow, line 75: the last parameter for createGroupButton is overrides, here you could define a isHidden value that you set with a ConditionalBooleanValue that syncs with the ViewStateChange and checks if the view is a sheet

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -80,10 +87,14 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
 
       if (tools.length > 0 && toolbarOrientation === ToolbarOrientation.Horizontal) {
         const isHidden = new ConditionalBooleanValue(
-          () => !MeasurementUIEvents.isClearMeasurementButtonVisible,
+          () => {
+            const vp = IModelApp.viewManager.selectedView;
+            return (vp ? !MeasurementUIEvents.isClearMeasurementButtonVisible || ViewHelper.isSheetView(vp) : !MeasurementUIEvents.isClearMeasurementButtonVisible);
```
jason-crow, line 92: why not
!MeasurementUIEvents.isClearMeasurementButtonVisible && ViewHelper.isSheetView(vp)

jason-crow (approved):

aruniverse (approved):

PR #338, opened 2022-11-03
[measure-tools-react] Hide Perpendicular distance tool for drawing view

cortneysmithbentley:
Don't show the perpendicular distance tool (under measure tools) when viewing a drawing.


packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -70,7 +71,7 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
       if (!featureFlags?.hideAngleTool) {
         tools.push(MeasureToolDefinitions.measureAngleToolCommand);
       }
-      if (!featureFlags?.hidePerpendicularTool) {
+      if (!featureFlags?.hidePerpendicularTool && !isDrawingViewActive()) {
```
jason-crow, line 74: This code isn't reactive to view change events. You need to make that tool hide itself when a 2d view is active

cortneysmithbentley: Pushed the fix for this.

jason-crow (approved):

PR #345, opened 2022-12-12
Move core visibility trees from itwinjs-core

saskliutas:
Moved core visibility trees used by the tree widget to this package. Updated itwin packages to 3.5.0 because some of tests are using new API release in only in 3.5.0


grigasp (changes requested):
- Let's omit the `JoesHouse.bim` dataset. There're only a few `ModelsTree` integration tests that use it and they're pretty much worthless anyway. We can add proper integration tests targeted towards specific issues when they come up.

- I'm not sure I like the current directory structure. I suggest this:
  ```
  ./components
    - trees
      - ...common stuff
      - category-tree
        - ...newly added category tree related stuff
        - CategoriesTree.tsx
      - models-tree
        - ...newly added models tree related stuff
        - ModelsTree.tsx
      - spatial-tree
        - ...newly added spatial tree related stuff
        - SpatialTree.tsx
  ```

packages/itwin/tree-widget/package.json
```diff
@@ -26,62 +26,84 @@
...
+    "xmlhttprequest": "^1.8.0"
   },
   "peerDependencies": {
-    "@itwin/appui-abstract": "^3.0.0",
-    "@itwin/appui-react": "^3.0.0",
-    "@itwin/components-react": "^3.0.0",
-    "@itwin/core-frontend": "^3.0.0",
-    "@itwin/core-react": "^3.0.0",
-    "@itwin/presentation-components": "^3.0.0",
+    "@itwin/appui-abstract": "~3.5.0",
```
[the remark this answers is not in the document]
saskliutas: Fixed. Forgot to revert while doing so testing

saskliutas:
> * Let's omit the `JoesHouse.bim` dataset. There're only a few `ModelsTree` integration tests that use it and they're pretty much worthless anyway. We can add proper integration tests targeted towards specific issues when they come up.
> * I'm not sure I like the current directory structure. I suggest this:
>   ```
>   ./components
>     - trees
>       - ...common stuff
>       - category-tree
>         - ...newly added category tree related stuff
>         - CategoriesTree.tsx
>       - models-tree
>         - ...newly added models tree related stuff
>         - ModelsTree.tsx
>       - spatial-tree
>         - ...newly added spatial tree related stuff
>         - SpatialTree.tsx
>   ```

Updated structure and removed `JoeHouse` imodel and tests based on it.

packages/itwin/tree-widget/public/locales/en/TreeWidget.json
```diff
@@ -12,5 +12,44 @@
   "models": "Models",
   "categories": "Categories",
   "containment": "Spatial Containment",
-  "treeview": "Tree View"
+  "treeview": "Tree View",
+  "coreTrees": {
```
aruniverse, line 16: do we need this top level, "coreTrees"?

packages/itwin/tree-widget/src/components/trees/category-tree/Categories.json
```diff
@@ -0,0 +1,102 @@
+{
+  "$schema": "../../../../node_modules/@itwin/presentation-common/Ruleset.schema.json",
+  "id": "ui-framework/CategoriesTree",
```
aruniverse, line 3: should this id now be updated to reflect where this tree is coming from? tree-widget/CategoriesTree?

saskliutas: Yes it makes sense to update id to match package. 

I also was thinking to update css classes to use `tree-widget` prefix instead of `ui-fw` but I am worried that it might be bigger breaking change than moving these trees from one package to another (apps that override styles of those tree will need to update css instead of just updating import). What do you think?

aruniverse: yea i was going to comment about the use of ui-fw, but i assumed this was mostly a direct port of the code from the core repo. i know the appui team is planning to kill off all usage of uifw and buic scss vars/mixins for appui 4.0 wasnt sure if we should tackled the ui thing then. thx for bringing this up

packages/itwin/tree-widget/src/components/core-trees/spatial-tree/SpatialContainmentTree.tsx
```diff
@@ -0,0 +1,69 @@
...
+   * @beta
+   */
+  enableElementsClassGrouping?: ClassGroupingOption;
+}
+
+/**
+ * Tree which displays and manages models or categories contained in an iModel.
+ * @public
+ */
+export function SpatialContainmentTree(props: SpatialContainmentTreeProps) {
```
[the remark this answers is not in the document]
grigasp: @JValiunas just confirmed that [SpatialContainmentTree](https://github.com/iTwin/viewer-components-react/blob/fbafc4c2cbf72a5ef414000cbee5afa9b4edab37/packages/itwin/breakdown-trees/src/Views/SpatialContainmentTree.tsx#L64) is the spatial containment tree everyone should be using. 

Let's remove it from this PR and allow it to peacefully die when UI code is moved from core.

packages/itwin/tree-widget/src/components/trees/category-tree/Categories.json
grigasp: Technically the components in this repo are different components than the ones currently in core. That, and the fact that people overriding these styles is unlikely, IMO, suggests we should go ahead and rename the classes to match the new package.

saskliutas: Renamed classes that contained `ui-fw` prefix in their name

grigasp (approved):

aruniverse (approved):

PR #370, opened 2023-02-24
Removed redundant callback

Guillar1:
Removing a redundant callback that was not necessary. useEffect was already making a call to visualizeGroupColorsWrapper.

### DIFF
--- a/packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
+++ b/packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
@@ -207,7 +207,6 @@ export const Groupings = ({
       await showGroup(group);
       setHiddenGroupsIds(hiddenGroupsIds.filter((id) => id !== group.id));
     }
-
   };
 
   const refresh = useCallback(async () => {
@@ -242,19 +241,6 @@ export const Groupings = ({
     const allIds = await getHiliteIdsFromGroupsWrapper(groups);
     await zoomToElements(allIds);
   }, [setHiddenGroupsIds, groups, hideGroupsWrapper, getHiliteIdsFromGroupsWrapper]);
-
-  const toggleGroupColor = useCallback(
-    async (e: any) => {
-      if (e.target.checked) {
-        await visualizeGroupColorsWrapper(groups);
-        setShowGroupColor(true);
-      } else {
-        clearEmphasizedOverriddenElements();
-        setShowGroupColor(false);
-      }
-    },
-    [groups, visualizeGroupColorsWrapper, setShowGroupColor],
-  );
 
   return (
     <>
@@ -302,7 +288,7 @@ export const Groupings = ({
                 className='gmw-toggle'
                 disabled={isLoadingQuery}
                 checked={showGroupColor}
-                onChange={toggleGroupColor}
+                onChange={() => setShowGroupColor((b) => !b)}
               ></ToggleSwitch>
               <IconButton
                 title='Show All'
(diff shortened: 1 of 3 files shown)


packages/itwin/grouping-mapping-widget/src/widget/components/context/GroupHilitedElementsContext.ts
```diff
@@ -13,7 +13,7 @@ export interface GroupHilitedElements {
   groups: Group[];
   setGroups: (groups: Group[]) => void;
   setHiddenGroupsIds: (hiddenGroupIds: string[]) => void;
-  setShowGroupColor: (showGroupColor: boolean) => void;
+  setShowGroupColor: (showGroupColor: boolean | ((showGroupColor: boolean) => boolean)) => void;
```
MindaugasButkus, line 16: Setter should only accept a boolean as an argument, not a function, which accepts a boolean argument.

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -302,7 +288,7 @@ export const Groupings = ({
                 className='gmw-toggle'
                 disabled={isLoadingQuery}
                 checked={showGroupColor}
-                onChange={toggleGroupColor}
+                onChange={() => setShowGroupColor((b) => !b)}
```
MindaugasButkus, line 291:
```suggestion
                onChange={() => setShowGroupColor(!showGroupColor)}
```

MindaugasButkus (changes requested):

packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
Guillar1: In React, it is better to do it the first way. Whenever the next state requires the previous state, it is good practice to use the functional state update. There is a risk of the state being updated based on a stale value if there are other updates happening simultaneously. The previous state is guaranteed to be most up to date.

MindaugasButkus: But is it applicable in this case? Is it applicable to checkboxes?

If I click on a checked checkbox I would want it to become unchecked, regardless of what is happening at the moment. I do not want it to be toggled. I want it unchecked.

Guillar1: But if you click on an unchecked checkbox you want it to be checked. It's still a toggle.
It's not strictly necessary in this case, but it's good practice to follow when working with states in React. 
Here is a stack overflow discussion on it. 
[https://stackoverflow.com/questions/70566016/better-way-to-use-usestate-hook-for-setting-the-boolean-state-in-react](url)

MindaugasButkus: You just reversed my example. It doesn't matter which way you go. Yes, it is a toggle. If that toggle would be changing its state by some random event without human interaction - sure, using the function passed to the state setter would be better. In this case the state can only be changed by a human pressing on a checkbox.

Guillar1: But we don't know when this could change. This is on a context level. It could possibly change outside of this component in another fashion. I'm just saying it doesn't hurt by preparing for every possibility in this case and by also applying good practice.

MindaugasButkus (approved):

PR #383, opened 2023-03-06
Fix "show all" and "hide all" buttons in model tree

### LINKED ISSUE #374: Models tree toolbar buttons don't work as expected
The "show all", "hide all", "invert" buttons in the Models Tree should make sure they really show/hide **ALL**. At the moment it seems they only work at the models' level and making sure all models are displayed doesn't guarantee that all iModel content is going to be visible.

simasjar:
Right now "show all" and "hide all" buttons affect only models.
With this change "show all" and "hide all" buttons will affect all iModel content

fixes #374


packages/itwin/tree-widget/src/components/trees/CategoriesVisibilityUtils.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import type { IModelConnection, ViewManager, Viewport } from "@itwin/core-frontend";
+import { CategoryVisibilityHandler, loadCategoriesFromViewport } from "./category-tree/CategoryVisibilityHandler";
```
saskliutas, line 6: I think functions that are used by `toggleAllCategories` should be defined here and not imported from `CategoryVisibilityHandler`. 

Also maybe it would be possible to decouple `toggleAllCategories` function from `IPresentationTreeDataProvider`? Maybe instead of taking `filteredProvider` it could take an optional list of `categoryIds`. In that case category ids lookup from `filteredProvider` can be moved to `CategoriesTreeComponent`. That way this function would be more reusable between visibility trees as it would not have `CategoryTree` specific implementation details.

saskliutas (changes requested):

common/changes/@itwin/tree-widget-react/Fix_show_all_button_in_model_tree_2023-03-06-09-14.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "tree-widget: ensure that the \"show all\" and \"hide all\" buttons affect all iModel content",
```
grigasp, line 5: The message will show up in the CHANGELOG of `tree-widget-react` package - no need to prefix this with `tree-widget`. Instead, I think we should prefix with the name of component that gets affected by the change.

packages/itwin/tree-widget/src/components/trees/CategoriesVisibilityUtils.ts
```diff
@@ -0,0 +1,131 @@
...
+  const ids = await getCategories(imodel, activeView);
+
+  // istanbul ignore else
+  if (ids.length > 0) {
+    enableCategory(viewManager, imodel, ids, display, forAllViewports ?? false);
+  }
+}
+
+/**
+ * Gets ids of all categories or categories from filtered data provider.
```
saskliutas, line 33: Comment needs updating after code restructuring.

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeComponent.tsx
```diff
@@ -84,15 +86,37 @@ export const ModelsTreeComponent = (props: ModelTreeProps) => {
...
-  const hideAll = useCallback(() => {
+  const hideAll = useCallback(async () => {
     viewport?.changeModelDisplay(availableModels, false);
+    viewport?.clearAlwaysDrawn();
+    if (iModel) {
+      await toggleAllCategories(
+        IModelApp.viewManager,
+        iModel,
+        false,
+        undefined,
```
saskliutas, line 97: I think `viewport` should be passes to this function. And probably we do not want to toggle categories in all viewports.

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/trees/CategoriesVisibilityUtils.ts
simasjar: fixed

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeComponent.tsx
simasjar: fixed

common/changes/@itwin/tree-widget-react/Fix_show_all_button_in_model_tree_2023-03-06-09-14.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "models-tree: ensure that the \"show all\" and \"hide all\" buttons affect all iModel content",
```
saskliutas, line 5:
```suggestion
      "comment": "`ModelsTreeComponent`: ensure that the \"show all\" and \"hide all\" buttons affect all iModel content",
```

PR #397, opened 2023-03-15
Migrated UI items providers to itwin 4.0.

mdastous-bentley:


eringram (approved):

aruniverse:
> @aruniverse Are good with this changes. I considered removing the 'automatic' ui component registration, but since already have an option for skipping it, I decided to not break the API for our consumers and keep things unchanged.

I think we should discuss with the AppUi team what to do here and what best practice is. It should be ok to make breaking changes in here, but i guess we should deprecate the older way of auto registering in 3.7 if we wanted to make that change

mdastous-bentley:
> > @aruniverse Are good with this changes. I considered removing the 'automatic' ui component registration, but since already have an option for skipping it, I decided to not break the API for our consumers and keep things unchanged.
> 
> I think we should discuss with the AppUi team what to do here and what best practice is. It should be ok to make breaking changes in here, but i guess we should deprecate the older way of auto registering in 3.7 if we wanted to make that change

Already discussed this with @NancyMcCallB and she suggested to remove it..  I am just trying to simplify the migration to 4.0 by minimizing breaking changes.  If you strongly suggest to remove it, then lets do it.  Would you say it's already too late to deprecate something in ?

aruniverse:
@mdastous-bentley  I'm in agreement with Nancy and do think it needs to be removed. I dont think the breaking change will be that bad for users to react to. 

@calebmshafer let us know if you think we should hold off

calebmshafer:
> @mdastous-bentley I'm in agreement with Nancy and do think it needs to be removed. I dont think the breaking change will be that bad for users to react to.
> 
> @calebmshafer let us know if you think we should hold off

@mdastous-bentley @aruniverse I'm good with removing it. We've heard some teams bring this up; they don't want the automatic registration due to the wanting to control it via feature flags which the registration doesn't allow.

mdastous-bentley:
@aruniverse Removed auto registration of UiItemsProviders, please have another look.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -4,53 +4,50 @@
...
+      ];
+    }
+
+    return [];
+  }
+
+  public provideWidgets(_stageId: string, stageUsage: string, location: StagePanelLocation, section?: StagePanelSection) {
+    const widgets = [];
+
+    if ((undefined === section && stageUsage === StageUsage.General && location === StagePanelLocation.Right ) ||
```
aruniverse, line 43: i think this can be condensed. previously had multiple conditions to support ui1.0 which is killed off in appui4.0

aruniverse (approved):

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
mdastous-bentley: should be better now.

PR #399, opened 2023-03-20
Ensure that the "Show all" button shows all models when there are models with an "exclusive" flag

### LINKED ISSUE #398: Tree Widget: "Show all" doesn't take effect all the time
The current implementation makes sure we display all models, all categories and there's nothing in the "never drawn" list. 

However, there may be some elements in "always drawn" list with "exclusive" flag, which means we're going to draw only those elements. Which means we're not going to "show all".

simasjar:
closes #398

### DIFF
--- a/common/changes/@itwin/tree-widget-react/Clear_elements_with_exclusive_on_show_all_2023-03-20-05-55.json
+++ b/common/changes/@itwin/tree-widget-react/Clear_elements_with_exclusive_on_show_all_2023-03-20-05-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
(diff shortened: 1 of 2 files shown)


common/changes/@itwin/tree-widget-react/Clear_elements_with_exclusive_on_show_all_2023-03-20-05-55.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "ModelsTree: ensure that the \"Show all\" button is going to show all models when there are models with an \"exclusive\" flag",
```
saskliutas, line 5:
```suggestion
      "comment": "ModelsTree: ensure that the \"Show all\" button is going to show all elements when there are elements with an \"exclusive\" flag",
```

saskliutas (approved):

PR #400, opened 2023-03-21
[G&M] Replaced Selection PropertyGrid

Guillar1:


packages/itwin/grouping-mapping-widget/src/widget/components/GroupAction.tsx
```diff
@@ -91,6 +91,7 @@ export const GroupAction = (props: GroupActionProps) => {
     if (groups.length > 0) {
       if (showGroupColor) {
         await visualizeGroupColors(iModelConnection, groups, groups, hiddenGroupsIds, hilitedElementsQueryCache);
+        return;
```
MindaugasButkus, line 94: How is this related?

Guillar1: Found this reset bug, it would `visualizeGroupColors` then ` clearEmphasizedElements` which was counterproductive. Not related, just forgot to make a change entry.

Guillar1: Actually scratch that. I misremembered. The bug was fixed when I rewrote the entire checkbox logic. This was here because the old logic was breaking this reset call.

MindaugasButkus:
I am not sure if this is related, but when I am creating a Group and I click on an element, I end up with this view:
image

As you can see I get property list on top as a separate widget and the properties to select for Group query creation is not reachable (I scrolled to the very bottom).

Guillar1:
> I am not sure if this is related, but when I am creating a Group and I click on an element, I end up with this view: image
> 
> As you can see I get property list on top as a separate widget and the properties to select for Group query creation is not reachable (I scrolled to the very bottom).

The top widget property grid is unrelated. I'm not sure how to reproduce this. What iTwinJs version are you running?

MindaugasButkus:
> > I am not sure if this is related, but when I am creating a Group and I click on an element, I end up with this view: image
> > As you can see I get property list on top as a separate widget and the properties to select for Group query creation is not reachable (I scrolled to the very bottom).
> 
> The top widget property grid is unrelated. I'm not sure how to reproduce this. What iTwinJs version are you running?

It's a fresh app created with `npx create-react-app your-app-name --template @itwin/web-viewer --scripts-version @bentley/react-scripts` . The app was created with iTwinJs 3.6.2.

Guillar1:
> > The top widget property grid is unrelated. I'm not sure how to reproduce this. What iTwinJs version are you running?
> 
> It's a fresh app created with `npx create-react-app your-app-name --template @itwin/web-viewer --scripts-version @bentley/react-scripts` . The app was created with iTwinJs 3.6.2.

I think I've figured it out. The issue is that the container's size is just too small, causing the `PropertyGrid` to disappear when its height reaches 0. The Group By container's height stays the same, but its children (the buttons) can't shrink anymore, so they overextend and make the entire `GroupAction` container overflow. The best solution would be for the contents of the `PropertyGrid` to cause the `GroupAction` container to overflow.

But the current implementation of the `PropertyGrid` makes this tricky. It needs fixed `width` and `height` properties and also has internal `overflow` handling. This makes it difficult for the content inside to overflow anything outside of its container due to its fixed `height` and `width`. So, the solution was to keep observing the size changes of the container of the `PropertyGrid` but this time opted for a 'fullscreen' effect within the widget, providing the necessary space for the `PropertyGrid`. 
image
image

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupQueryBuilderCustomUI.tsx
```diff
@@ -127,20 +129,35 @@ export const GroupQueryBuilderCustomUI = ({
...
-              size="small"
-              onClick={onClickResetButton}
-            >
-              Reset
-            </Button>
-          </div>
         </>
-      }
+      )}
+      <div className="gmw-select-footer">
```
MindaugasButkus, line 136: It looks like this could be a component called `SelectFooter`.

MindaugasButkus:
> > > The top widget property grid is unrelated. I'm not sure how to reproduce this. What iTwinJs version are you running?
> > 
> > 
> > It's a fresh app created with `npx create-react-app your-app-name --template @itwin/web-viewer --scripts-version @bentley/react-scripts` . The app was created with iTwinJs 3.6.2.
> 
> I think I've figured it out. The issue is that the container's size is just too small, causing the `PropertyGrid` to disappear when its height reaches 0. The Group By container's height stays the same, but its children (the buttons) can't shrink anymore, so they overextend and make the entire `GroupAction` container overflow. The best solution would be for the contents of the `PropertyGrid` to cause the `GroupAction` container to overflow.
> 
> But the current implementation of the `PropertyGrid` makes this tricky. It needs fixed `width` and `height` properties and also has internal `overflow` handling. This makes it difficult for the content inside to overflow anything outside of its container due to its fixed `height` and `width`. So, the solution was to keep observing the size changes of the container of the `PropertyGrid` but this time opted for a 'fullscreen' effect within the widget, providing the necessary space for the `PropertyGrid`.

I don't think the solution of introducing a button to open properties list is good... It makes the group creation more complex and requires user to click on additional button. Why would I want to open properties when I want to create a group?

Also, the padding of the footer buttons looks off when you click on the "Open Properties" button. It's too small. And some long properties overflow outside of the property table:
image

packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupQueryBuilderCustomUI.tsx
MindaugasButkus: No longer relevant.

MindaugasButkus:
> Does it happen on both the PropertyGrids? (The included one as well)

Does not happen anymore. Must have been something with my local build.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupAction.tsx
Guillar1: Resolving as it's no longer relevant.

MindaugasButkus:
Looks like the UI changes quite a bit. From this:
image

To this:
image

Notice the difference in the scope of vertical scrollbar.

I'd say this needs to be looked at by QA and UX before we can merge this PR.

See https://github.com/iTwin/viewer-components-react/pull/406 which addresses the same issue as this PR but with less drastic changes (UI stays the same).

packages/itwin/grouping-mapping-widget/src/widget/components/GroupDetails.tsx
MindaugasButkus, line 1: Missing copyright header

packages/itwin/grouping-mapping-widget/src/widget/components/GroupAction.tsx
```diff
@@ -238,95 +211,116 @@ export const GroupAction = (props: GroupActionProps) => {
...
-          props.onClickCancel && props.onClickCancel();
-        } : undefined}
-        isSavingDisabled={isBlockingActions}
-        isLoading={isLoading}
-      />
+      <div className='gmw-action-panel'>
+        {isLoading &&
+          <LoadingSpinner />
+        }
+        {currentStep !== GroupActionStep.QueryBuilder ?
```
MindaugasButkus, line 283: It would be best to have separate components for each step instead of relying on a bunch of ternary operators to display the right buttons / titles / classes, etc.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupingMappingHeader.tsx
```diff
@@ -4,22 +4,26 @@
...
 export const GroupingMappingHeader = ({
-  routingHistory,
   goBack,
   currentRoute,
 }: {
-  routingHistory: Route[];
   goBack: () => void;
   currentRoute: Route;
 }) => {
+  const shouldDisableReturnFn =
```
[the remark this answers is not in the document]
Guillar1: None of the shareable components should have routing logic within them. This solution is specific to the widget and should not be exposed. I believe this does not go against componentization.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupAction.tsx
Guillar1: Separated components.

PR #411, opened 2023-03-30
Upgrade `property-grid-react` to AppUI 4.0

saskliutas:
Upgraded dependencies: itwinjs-core to 3.7, appui to 4.0, presentation-components to 4.0 and itwinui to 2.0.
Fixed deprecations.


grigasp (approved):

packages/itwin/property-grid/package.json
```diff
@@ -97,9 +99,9 @@
     ],
     "extends": "plugin:@itwin/ui",
     "rules": {
-      "@typescript-eslint/no-shadow": "off",
+      "@itwin/no-internal": "off",
       "no-duplicate-imports": "off",
```
aruniverse, line 103: i think you can nuke this line if youre going to add the one on 104

i had initially added this so we can use type imports and help devs quickly determine if a dep needs to be a peer or not

saskliutas: Removing `"no-duplicate-imports": "off"` lie results in eslint complaining about duplicate imports when using type imports. 

I don't mind using type imports and for that to work `import/no-duplicates` rule should be used instead of `no-duplicate-imports` which is used by [`itwinjs` eslint config](https://github.com/iTwin/eslint-plugin/blob/754e2c6a1fdf50277f3f25086e0df75e4226ef46/dist/configs/itwinjs-recommended.js#L290)

aruniverse: Up to you guys what you want to use as the eslint config as you guys are owning this pkg moving fwd and I dont want to force anything on you guys

saskliutas: Well this package was using type imports before we took ownership so I just wanted to resolve lint errors without disabling duplicate imports rule completely. 😄 

Do you think we should stick to using type imports or stick to recommended `itwinjs` eslint config and avoid type imports? Because from what I found `no-duplicate-imports` rule does not work with type imports.

aruniverse: I wonder if we should look into using https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues and use "error". I think this would just force us to manually use type imports as needed and then we wouldnt need any additional linter rules or changes to make. i should note, i havent tried this yet

grigasp:
@aruniverse I think this is waiting for you approval

aruniverse (approved):

PR #420, opened 2023-04-07
Add test viewer for testing widgets locally

saskliutas:
- Added test-viewer created using web-viewer templace https://github.com/iTwin/viewer/tree/master/packages/modules/cra-template-web-viewer
- Added tree-widget and property-grid-widget to the viewer. Widget can be enabled/disabled through environment variable `IMJS_ENABLED_WIDGETS`
- Added script to support hot reloading after changes in local packages.


grigasp (commented):
- I don't think this repo is going to have any other "app", so I'd vote for placing the app to `app`, `test-app` or `test-viewer` directory rather than `apps/test-viewer`.
- I believe we need to clean up and provide more relevant information in the app's `README.md`
- I'd like to hear @aruniverse opinion about this.

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      if (!process.env.IMJS_ITWIN_ID) {
+        throw new Error(
+          "Please add a valid iTwin ID in the .env file and restart the application or add it to the iTwinId query parameter in the url and refresh the page. See the README for more information."
+        );
+      }
+    }
+
+    if (urlParams.has("iModelId")) {
+      setIModelId(urlParams.get("iModelId") as string);
```
aruniverse, line 104: dont need to check/throw error if imodel id is missing in .env?

aruniverse (approved):
> I don't think this repo is going to have any other "app", so I'd vote for placing the app to app, test-app or test-viewer directory rather than apps/test-viewer.
I'd like to hear @aruniverse opinion about this.

I'm ok with the addition of this test app viewer. I was going to look into possibly creating a standalone pkg for debugging uiProviders later in the quarter. but this is a good solution for now. I dont think there will be any other tests apps. I'l leave naming and where the app should live to you guys.

Pipelines have also been fixed, after conflict is resolved should be good to go

grigasp (approved):

apps/test-viewer/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Running app
+
+`npm start`
+
+### Running app with workspace packages hot reloading
+
+`npm run start:dev`
+
+By default `pnpm` symlinks dependencies to the modules directory. But this causes problems if workspace package `devDependencies` and app `dependencies` have different versions of the same package. To avoid this problem workspace packages should be [injected](https://pnpm.io/package_json#dependenciesmetainjected) into app. However, this breaks `webpack` hot reloading feature.
```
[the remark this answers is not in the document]
saskliutas: I don't think it is possible to avoid this without forcing all pkgs to use same versions. It shouldn't be a problem when all pkgs used by test-viewer and test-viewer are updated to 4.0.0 but it will comeback as soon as one of those pkgs moves to new minor or patch versions. As that will require to update test-viewer deps and as a result other pkgs used in test-viewer.

apps/test-viewer/src/components/Viewer.tsx
saskliutas: This code is from cra-web-viewer-template I only moved it around to make component smaller. I can add check if you think it's necessary.

aruniverse: It's fine. Resolving...
I'll make sure we update the cra template

Issue #377, opened 2023-02-28
Tree Widget: Rethink the way different trees are supplied to the widget


grigasp:
At the moment the widget has a default set of trees, but provides a pretty inconvenient way to hide them and include others. We should rethink the API to make it more convenient to use.

PR #428, opened 2023-04-14
Tree-Widget: Rework how trees are added to widget

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/377

Refactored `TreeWidgetUiItemsProvider` to have single config property for controlling trees shown in widget instead of having multiple properties hiding/adding trees.

Fixed eslint.

### DIFF
--- a/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
+++ b/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/tree-widget-react",
-      "comment": "Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
+      "comment": "**BREAKING:** Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
       "type": "none"
     }
   ],
--- a/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
+++ b/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/tree-widget-react",
-      "comment": "**BREAKING:** Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
+      "comment": "**BREAKING:** Reworked how trees are provided to the widget. Instead of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
       "type": "none"
     }
   ],
--- a/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
+++ b/common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/tree-widget-react"
+}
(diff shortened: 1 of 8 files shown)


packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -2,30 +2,19 @@
...
 
 export interface TreeWidgetOptions {
   defaultPanelLocation?: StagePanelLocation;
   defaultPanelSection?: StagePanelSection;
   defaultTreeWidgetPriority?: number;
-  enableElementsClassGrouping?: boolean;
-  additionalTrees?: SelectableContentDefinition[];
-  modelsTreeProps?: ModelTreeComponentProps;
-  categoriesTreeProps?: CategoriesTreeComponentProps;
-  defaultTreeId?: string;
```
aruniverse, line 24: is the recommended approach moving fwd going to be, if you want to set a specific tree as the default, or add an additional tree, or change any of the defaults for our default trees here, you need to specify all that yourself?

common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
```
aruniverse, line 5: might want to mention this is breaking

packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
saskliutas: Yes. I think this way provides the most flexibility for consumers.

common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
saskliutas: Updated change comment

common/changes/@itwin/tree-widget-react/rework_additional_trees_handling_2023-04-14-14-58.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "**BREAKING:** Reworked how trees are provided to the widget. Instaed of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
```
grigasp, line 5:
```suggestion
      "comment": "**BREAKING:** Reworked how trees are provided to the widget. Instead of having separate configuration properties for hiding default trees and adding additional trees now there is only one property for supplying a list of trees to show.",
```

grigasp (approved):

packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -40,64 +29,32 @@ export class TreeWidgetUiItemsProvider implements UiItemsProvider {
...
-        content: <TreeWidgetComponent trees={trees} />,
-        icon: "icon-hierarchy-tree",
-        priority: this._treeWidgetOptions?.defaultTreeWidgetPriority,
-      });
+    if (location !== preferredLocation || section !== preferredPanelSection || stageUsage !== StageUsage.General) {
+      return [];
     }
 
-    return widgets;
+    const trees: TreeDefinition[] = this._treeWidgetOptions?.trees ?? [
```
jason-crow, line 39: So we have no way to add additional trees? its either just these 2 or we provide everything?

saskliutas: These 2 trees are shown by default only for the simplest case when you want to see default `ModelsTree` and `CategoriesTree`. If you want to customize content shown in `TreeWidget` (hide, add, reorder, modify trees) you will need to explicitly specify what you want to see. I know this adds some additional code when you want to have default `ModelsTree` and `CategoriesTree` with some additional trees (about 6 lines) but it makes API more flexible and cleaner.

grigasp: @jason-crow, the goal here is to simplify the API and make it more flexible at the same time.

- The way available trees was controlled was really unintuitive - that was controlled through `hideTrees`, `additionalTrees` and `defaultTreeId` props. To change the order of the trees you would've had to hide all trees and provide them in correct order through `additionalTrees`.

- Configuring default trees - the props were provided through `enableElementsClassGrouping`, `modelsTreeProps` and `categoriesTreeProps`.
  - What's the meaning of `{ modelsTreeProps: { ...customProps }, hideTrees: { modelsTree } }`?
  - What's the meaning of `{ enableElementsClassGrouping: true, modelsTreeProps: { enableElementsClassGrouping: ClassGroupingOption.No }`?

The current approach solves all those inconsistencies - you pass in all the trees you want to see in the widget, in the order you want, with the configuration you want. 

Please let us know if you have any concerns - all 1.0 changes are being pushed to a separate branch and we can make improvements before we release, if necessary.

PR #430, opened 2023-04-15
[G&M] Visualization race condition fix

Guillar1:
These sorts were sorting in place causing very rare race condition issues (list mutating in the middle of for loops...) due to the state becoming uncontrolled. 
Found this issue in another branch in the middle of refactoring.

### DIFF
--- a/packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
+++ b/packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
@@ -75,7 +75,7 @@ const fetchGroups = async (
       iModelId,
       mappingId,
     );
-    setGroups(groups);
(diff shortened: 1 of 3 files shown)


packages/itwin/grouping-mapping-widget/src/widget/components/Grouping.tsx
```diff
@@ -75,7 +75,7 @@ const fetchGroups = async (
       iModelId,
       mappingId,
     );
-    setGroups(groups);
+    setGroups(groups.sort((a, b) => a.groupName?.localeCompare(b.groupName ?? "") ?? 1));
```
MindaugasButkus, line 78: `Group.groupName` from `@itwin/insights-client` should never be null or undefined.

MindaugasButkus (approved):

PR #431, opened 2023-04-17
Make property grid column resizer more movable

simasjar:
closes iTwin/presentation#116

### DIFF
--- a/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
+++ b/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
@@ -179,7 +179,7 @@ export const FilteringPropertyGrid = (
         {...props}
         minLabelWidth={10}
         minValueWidth={10}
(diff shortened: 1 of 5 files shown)


saskliutas (approved):

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -177,6 +177,9 @@ export const FilteringPropertyGrid = (
     <>
       <VirtualizedPropertyGridWithDataProvider
         {...props}
+        minLabelWidth={10}
+        minValueWidth={10}
+        actionButtonWidth={props.actionButtonWidth ?? props.actionButtonRenderers ? 90 : 0}
```
saskliutas, line 182: Maybe instead of 90 we can pass `undefined` so that `VirtualizedPropertyGrid` default would be used?

simasjar: fixed

PR #433, opened 2023-04-18
[G&M] Group Visual Split

Guillar1:


packages/itwin/grouping-mapping-widget/src/widget/components/GroupsShowHideButtons.tsx
```diff
@@ -0,0 +1,48 @@
...
+import type { Group } from "@itwin/insights-client";
+import { SvgVisibilityHide, SvgVisibilityShow } from "@itwin/itwinui-icons-react";
+import { IconButton } from "@itwin/itwinui-react";
+import React, { useCallback, useMemo } from "react";
+
+interface GroupsShowHideButtonsProps {
+  group: Group;
+  hiddenGroupsIds: string[];
+  isLoadingQuery: boolean;
+  setHiddenGroupsIds: (ids: string[]) => void;
```
MindaugasButkus, line 14: It might be better to use a `Set` for this to ensure uniqueness. `includes` operations should work a bit faster with sets.

MindaugasButkus (approved):

packages/itwin/grouping-mapping-widget/src/widget/components/GroupsShowHideButtons.tsx
Guillar1: I agree, changed.

PR #437, opened 2023-04-19
Replace removed variable $uicore-button-focus-boxshadow in MapLayer's scss files

jason-crow:
Webpack is throwing errors compiling @itwin/map-layers due to the usage of $uicore-button-focus-boxshadow which no longer exists.

I've swapped those references with `rgba(var(--iui-color-border-accent), var(--iui-opacity-4)) 0px 0px 0px 2px;`, which seems to be the easiest replacement given the following:
- In 3.7.x version of appui I found: `$uicore-button-focus-boxshadow: $buic-focus-boxshadow 0px 0px 0px 2px;` and also from there I found `--buic-focus-boxshadow: rgba(var(--iui-color-foreground-primary-rgb), var(--iui-opacity-4));` 
- From the iTwinUI migration [doc](https://github.com/iTwin/iTwinUI/wiki/iTwinUI-variables-v1-migration-guide) I found `--iui-color-foreground-primary-rgb` is now equivalent to `--iui-color-border-accent`

I'm sure there is a simpler replacement, but this appears to be close to 1 to 1 with now non existent original variable. I'm thinking we can get this in so the build is viable then perhaps someone can determine if a better alternative can be used

### DIFF
--- a/common/changes/@itwin/map-layers/fix-map-layers-scss_2023-04-18-21-52.json
+++ b/common/changes/@itwin/map-layers/fix-map-layers-scss_2023-04-18-21-52.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/map-layers",
+      "comment": "remove refs of $uicore-button-focus-boxshadow in maplayers scss files",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/map-layers"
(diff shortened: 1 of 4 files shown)


aruniverse (approved):
Test app worked fine?
No other changes needed for the other vcr pkgs you pull in?

PR #440, opened 2023-04-20
[G&M] Group Action Optional Visualization

Guillar1:


packages/itwin/grouping-mapping-widget/src/widget/hooks/useVisualization.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (showGroupColor) {
+      await visualizeGroupColors(iModelConnection, groups, hiddenGroupsIds, hilitedElementsQueryCache);
+    } else {
+      clearOverriddenElements();
+    }
+    clearEmphasizedElements();
+  }, [groups, hiddenGroupsIds, hilitedElementsQueryCache, iModelConnection, showGroupColor, shouldVisualize]);
+
+  useEffect(() => {
+    if (!shouldVisualize) return;
```
MindaugasButkus, line 30: Probably the listener should be removed if `shouldVisualize` toggles from `true` to `false`.

Guillar1: It does get removed, the cleanup function of the previous effect will run before the new effect. https://react.dev/reference/react/useEffect#parameters

MindaugasButkus (approved):

PR #446, opened 2023-04-25
Added geo-tools-react package to viewer and enabled background map

mdastous-bentley:


apps/test-viewer/package.json
```diff
@@ -20,6 +20,7 @@
     "@itwin/core-quantity": "^3.7.1",
     "@itwin/core-react": "^4.0.0-dev.15",
     "@itwin/core-telemetry": "^3.7.1",
+    "@itwin/geo-tools-react": "workspace:*",
```
aruniverse, line 23: You also need to add the dependencies meta

aruniverse (approved):

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -20,6 +20,8 @@ export function Viewer() {
   const { client: authClient } = useAuthorizationContext();
   const viewCreatorOptions = useViewCreatorOptions();
   const { iTwinId, iModelId } = useIModelInfo();
+  const [BingMapsKey] = useState(process.env.IMJS_BING_MAPS_KEY ?? "");
+  const [CesiumKey] = useState(process.env.IMJS_CESIUM_ION_KEY ?? "");
```
saskliutas, line 24: Looks like these can be moved outside `Viewer` component and don't need to be stored in React state.

saskliutas (approved):

PR #456, opened 2023-05-02
[G&M] Fixed missing group properties

Guillar1:


packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
```diff
@@ -160,9 +160,9 @@ export const GroupPropertyAction = ({
 
       if (!iModelConnection) return;
 
-      const result = await getHiliteIdsAndKeysetFromGroup(iModelConnection, group, hilitedElementsQueryCache);
+      const result = await manufactureKeys(group.query, iModelConnection);
```
MindaugasButkus, line 163: Query result caching is lost.

Guillar1: The cached query result first included both hilite element IDs and keysets. But, Presentation just needs the keysets. To create hilite IDs, you need an active ScreenViewport, which usually happens when the viewer is up and running.

I thought about a few options when there's no viewer. One was to tweak the getHiliteIds function to give back only the keysets and an empty ID list, basically caching a bad result. Another choice was to cache the normal query result, but that didn't make much sense since only viewer-related stuff uses these cached IDs, so the result would still be incomplete.

In the end, I chose to skip this whole process and just grab the keysets.

MindaugasButkus: It would be good to have the caching back eventually. For now I'll approve the PR, but we will need to come back to this.

MindaugasButkus (approved):

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
Guillar1: Later, I can maybe split the cache into keySets and hiliteId's.

PR #457, opened 2023-05-02
Tree widget: Update to AppUI and Presentation 4.0 (#457)Co-authored-by: Arun George <11051042+aruniverse@users.noreply.github.com>

saskliutas:
Updated AppUI and Presentation dependencies to 4.0
Injected tree-widget into breakdown-trees. Otherwise they need to have same devDependencies versions


aruniverse:
If you're getry ready to publish the new minor might want to update the version policy and the versions manually

common/changes/@itwin/tree-widget-react/tree_widget_appui_4.0_2023-05-02-12-44.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/tree-widget-react",
+      "comment": "Updated to AppUI 4.0 and Presentation 4.0",
+      "type": "none"
```
aruniverse, line 6:
```suggestion
      "type": "minor"
```

saskliutas:
@aruniverse Do you have any ideas what to do with `breakdown-trees` dependency on `tree-widget`? If they use same devDependencies versions everything works ok. If they have different versions then build fails. I tried adding `injected` metadata property but in that case it looks like `rush install` should be run after `tree-widget` is built in order to copy it's lib to `breakdown-trees` node_modules. 

Maybe `breakdown-trees` should pull `tree-widget` from npm? 

Or choose easy way and just update `breakdown-trees` and `tree-widget` dependencies together?

aruniverse:
> @aruniverse Do you have any ideas what to do with `breakdown-trees` dependency on `tree-widget`? If they use same devDependencies versions everything works ok. If they have different versions then build fails. I tried adding `injected` metadata property but in that case it looks like `rush install` should be run after `tree-widget` is built in order to copy it's lib to `breakdown-trees` node_modules.
> 
> Maybe `breakdown-trees` should pull `tree-widget` from npm?
> 
> Or choose easy way and just update `breakdown-trees` and `tree-widget` dependencies together?

I think long term the breakdown trees pkg will be killed off. For the time being i think we could just update deps together.
Jake has opened a pr to update breakdown trees, we can let your pr go in first, and then merge your changes into his.
https://github.com/iTwin/viewer-components-react/pull/459

saskliutas:
> > @aruniverse Do you have any ideas what to do with `breakdown-trees` dependency on `tree-widget`? If they use same devDependencies versions everything works ok. If they have different versions then build fails. I tried adding `injected` metadata property but in that case it looks like `rush install` should be run after `tree-widget` is built in order to copy it's lib to `breakdown-trees` node_modules.
> > Maybe `breakdown-trees` should pull `tree-widget` from npm?
> > Or choose easy way and just update `breakdown-trees` and `tree-widget` dependencies together?
> 
> I think long term the breakdown trees pkg will be killed off. For the time being i think we could just update deps together. Jake has opened a pr to update breakdown trees, we can let your pr go in first, and then merge your changes into his. #459

You will need to force merge this because build will fail.

PR #463, opened 2023-05-04
grouping-mapping-widget to 3.7.4

arnobmallickbsw:


packages/itwin/grouping-mapping-widget/src/widget/GroupingMappingWidget.tsx
```diff
@@ -23,19 +21,13 @@ export class GroupingMappingProvider implements UiItemsProvider {
...
-    if (
-      (location === StagePanelLocation.Left &&
-        section === StagePanelSection.Start &&
-        stageUsage === StageUsage.General) ||
-      zonelocation === AbstractZoneLocation.CenterLeft
-    ) {
-      const GroupingMappingWidget: AbstractWidgetProps = {
+  ): ReadonlyArray<CommonWidgetProps> {
+    const widgets: CommonWidgetProps[] = [];
+    if (location === StagePanelLocation.Left && section === StagePanelSection.Start) {
```
MindaugasButkus, line 29: Why does this no longer use `stageUsage` to determine if widget needs to be added?

arnobmallickbsw: So I was thinking initially, if we restrict the location the widget can appear, then `StageUsage` isn't useful to the framework and might cause problems if an app developer wants a `StageUsage` other than `StageUsage.General` on `StagePanelLocation.Left`.

But now I'm actually thinking we could go the opposite way. Remove the `StagePanelLocation` restriction and only gate on `StageUsage.General`.

Guillar1: We can just add additional optional props in the constructor to let the user provide the location and just leave these as default.

arnobmallickbsw: The 4.x changes will have another refactor of this. I'll revert whatever changes I made now, and implement the optional parameter suggestion in the 4.x PR.

PR #464, opened 2023-05-04
[measure-tools] Adding cleanup function to prevent node leaks during tests

Jake-Screen:
Adding cleanup function to prevent node leaks during tests. these were occurring the pipeline during the windows run. adding the cleanup function waits for promises to resolve, closing leaks.


mdastous-bentley (approved):

aruniverse (changes requested):
can you actually delete the iModelJs and CoreTools.json. They will overwrite what gets delivered by the core packages when this is actually consumed.

See if the leak is resolved with the testutils change you made

aruniverse (approved):

Issue #448, opened 2023-04-26
Tree Widget: Add API release tags


grigasp:
At the moment the only difference between public and internal APIs is the barrel export. It's very inconvenient as it makes it hard to know whether a file is exported through the barrel (it may be exported indirectly) and error prone as it's very easy to mistakenly make a function public by adding an internal function to an exported file.

At the very least we should tag our functions. Using API exporter would help us identify possible issues (e.g. public function returning internal type) and track breaking changes in public API.

PR #467, opened 2023-05-05
Tree widget: API extractions

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/448


packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
```diff
@@ -10,14 +10,31 @@ import { CategoriesTreeComponent } from "./trees/category-tree/CategoriesTreeCom
...
 }
 
+/**
+ * Id of the tree widget created by [[TreeWidgetUiItemsProvider]].
+ * @public
+ */
 export const TreeWidgetId = "tree-widget-react:trees";
+
+/**
+ * A [[UiItemsProvider]] implementation that provides a [[TreeWidgetComponent]] into
```
saskliutas, line 35: `into widget`?

saskliutas (approved):

packages/itwin/tree-widget/src/components/TreeWidgetUiItemsProvider.tsx
grigasp: must've been interrupted 🤣 isn't it providing "widgets" into "stage panels"? that would mean the fix should be `into a stage panel`.

saskliutas: Yes, looks like it should be `into a stage panel`. 🤦

grigasp: fixed

grigasp:
@aruniverse any objections?

packages/itwin/tree-widget/src/components/TreeFilteringState.ts
```diff
@@ -3,8 +3,9 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 import React, { useState } from "react";
-import type { IPresentationTreeDataProvider } from "@itwin/presentation-components";
+import { IPresentationTreeDataProvider } from "@itwin/presentation-components";
```
jason-crow, line 6: was it intentional to remove the `type` import?

grigasp: Yes, I did that for consistency with other imports in the package. 

I guess I don't see the value of using these `import type` imports since I have never seen any issues with simple `import`, so I just choose to use the latter everywhere. Did Design Review have any problems with that?

aruniverse (approved):

packages/itwin/tree-widget/src/components/TreeFilteringState.ts
jason-crow: @grigasp no its no issue, but I would recommend we make it the standard for imports that can be done as type since that will alleviate the import from getting pulled in the javascript, which an isolated case like this is unlikely to have any impact but if done consistently it can

grigasp: I wonder if there's any lint rule that would help us with that, otherwise it would certainly be very hard to not make a mess of those imports, especially when auto-importing. 

In any case, if this is not a problem at the moment, let's get this in and we can get back to **consistently** using `import type` afterwards.

grigasp: Filed a [backlog item](https://github.com/iTwin/viewer-components-react/issues/470) for this and closing the thread.

PR #469, opened 2023-05-05
Tree Widget: Add more unit tests

saskliutas:
Added unit tests for TreeHeader, SearchBox, TreeWidgetComponent, TreeWidget and TreeWidgetUiItemsProvider


grigasp (changes requested):
I believe `aria-label` attributes should be localized

saskliutas:
> I believe `aria-label` attributes should be localized

Added localization

grigasp (approved):

Issue #472, opened 2023-05-08
Models Tree: Let consumers configure `Element` classes that are loaded into the hierarchy


grigasp:
At the moment `ModelsTree` uses all types of `bis.GeometricElement3d` in the hierarchy. Some consumers may want to specify a more concrete subclass or exclude some subclasses. We should make this configuration option available.

PR #473, opened 2023-05-11
Models tree: Hierarchy configuration

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/472.

- **BREAKING:** `ModelsTreeProps.enableElementsClassGrouping` has been moved to `ModelsTreeProps.hierarchyConfig.enableElementsClassGrouping`.
- Added an option to specialize the class of geometric elements loaded into the hierarchy (see `ModelsTreeHierarchyConfiguration.elementClassSpecification`)

The issue mentions support for excluded classes (e.g. pick all `bis.GeometricElement3d` except `x.MyClass`) which I intentionally left out of this PR. That's not an immediate requirement for the consumer this change is being made for and implementing it in the Models Tree hierarchy would introduce one of the following:
- Some `bis.Subject` nodes showing a chevron, indicating that they have child nodes, but clicking on it resulting in no child nodes and chevron becoming hidden.
- Performance penalty as we would have to do a more expensive children check for `bis.Subject` nodes.

We can revisit support for excluded classes when (if) we restructure Models Tree hierarchy.


packages/itwin/tree-widget/src/components/trees/index.ts
```diff
@@ -4,13 +4,17 @@
...
+export * from "./imodel-content-tree/IModelContentTreeComponent";
+
 export * from "./category-tree/CategoriesTree";
 export * from "./category-tree/CategoriesTreeComponent";
 export * from "./category-tree/CategoryVisibilityHandler";
+
 export * from "./models-tree/ModelsTree";
 export * from "./models-tree/ModelsTreeComponent";
 export * from "./models-tree/ModelsVisibilityHandler";
+export * from "./models-tree/Utils";
```
saskliutas, line 20: I think this export can be removed. Everything in Utils.ts is internal so there is no need to reexport it through barrel file.

saskliutas (approved):

packages/itwin/tree-widget/src/components/trees/index.ts
grigasp: Our perf tests are going to use this internal API to get the ruleset. We could access the module directly, but doing that through the barrel is cleaner and has smaller chance of breaking our perf tests in case the API is moved to another file.

Issue #416, opened 2023-04-04
Property Widget: Review API


grigasp:
There is a lot of cleanup that need to be done regarding API of the Property Widget, e.g.:
- Props are not documented at all or documented very poorly.
- Prop `favoritePropertiesScope` makes no sense if you don't set `enableFavoriteProperties`.
- Prop `headerContent` makes no sense as it's overriden internally.
- Some context menu options are enabled through flags (e.g. `enableCopyingPropertyText`) and some by providing an array of options (`additionalContextMenuOptions`)

... and a bunch of other issues.

PR #481, opened 2023-05-15
Property grid: Refactor property grid components

saskliutas:
Reviewed Property-grid API and refactored it. Part of https://github.com/iTwin/viewer-components-react/issues/416.
- Split huge `PropertyGrid` component into custom hooks.
- Separated `PropertyGrid` component into `PropertyGrid` that uses unified selection and `SingleEmenentPropertyGrid` that show properties of specific instance.
- Changed default property data provider from auto expanding into simple `PresentationPropertyDataProvider` because it was wapped into `AutoExpandingPropertyFilterDataProvider` anyway.
- Added `createDataProvider` prop for providing custom data provider.


packages/itwin/property-grid/src/hooks/UseDataProvider.ts
```diff
@@ -0,0 +1,65 @@
...
+  /** Flag that specified whether properties favorite properties support should be enabled or not. */
+  enableFavoriteProperties?: boolean;
+  /** Flag that specified whether nested property categories are enabled. */
+  enablePropertyGroupNesting?: boolean;
+  /** Id of ruleset that should be used when pulling data. */
+  rulesetId?: string;
+  /** Callback that creates custom data provider that should be used instead of default one. */
+  createDataProvider?: (imodel: IModelConnection) => PresentationPropertyDataProvider;
+}
+
```
grigasp: I believe this should be a union:
```ts
{
  /** Flag that specified whether properties favorite properties support should be enabled or not. */
  enableFavoriteProperties?: boolean;
  /** Flag that specified whether nested property categories are enabled. */
  enablePropertyGroupNesting?: boolean;
  /** Id of ruleset that should be used when pulling data. */
  rulesetId?: string;
} | {
  /** Callback that creates custom data provider that should be used instead of default one. */
  createDataProvider: (imodel: IModelConnection) => PresentationPropertyDataProvider;
}
```

Also, I guess `createDataProvider` could return `IPresentationPropertyDataProvider`?

grigasp (approved):

PR #487, opened 2023-05-17
Migrate geo-tools to itwinjs-core 4.0.x

mdastous-bentley:


packages/itwin/geo-tools/package.json
```diff
@@ -43,25 +43,25 @@
     "react-autosuggest": "^10.1.0"
   },
   "devDependencies": {
-    "@itwin/appui-abstract": "^3.7.1",
+    "@itwin/appui-abstract": "^4.0.0-dev.95",
```
aruniverse, line 46: please use the rc, so `^4.0.0-dev.104` not 95

mdastous-bentley: should look better now

PR #498, opened 2023-05-19
Property grid: Remove `customOnDataChanged` property

saskliutas:
Removed `customOnDataChanged` callback from property grid props. This callback was invoked when `onDataChanged` event was raised. Same behavior can be replicated using `createDataProvider` and `onDataChanged` event:

```typescript
new PropertyGridUiItemsProvider({
  propertyGridProps: {
    createDataProvider: (imodel) => {
      const provider = new PresentationPropertyDataProvider({ imodel });
      provider.onDataChanged.addListener(() => { someCallback(); });
      return provider;
    }
  }
})
```


grigasp:
@jason-crow please have a look at this too

PR #508, opened 2023-05-23
Fix an issue with MapLayerPreferences that would not save with the proper key with replacing an existing source.

mdastous-bentley:
.. also added `MapLayersPrefBrowserStorage` as a simple/testing implementation of `UserPreferencesAccess`.


mdastous-bentley:
@calebmshafer we might want to move `PreferencesBrowserStorage` in a more "common" package if you think it might be useful for other applications.

mdastous-bentley:
@aruniverse please review

packages/itwin/map-layers/src/PreferencesBrowserStorage.ts
```diff
@@ -0,0 +1,125 @@
...
+* @beta
+*/
+export interface PreferencesBrowserStorageOptions {
+  /** indicate whether or not delete function should throw an error if the provided key cannot be found. */
+  throwOnDeleteMissingKey?: boolean;
+}
+
+/** User preferences implementation using browser's local storage.
+  * It creates storage item key in the form of:
+  *     <storageItemKeyPrefix>.<project_guid?>.<imodel_guid?>
```
aruniverse, line 25: we're not using the term project anymore, can you replace with itwin?

aruniverse: L27 too

aruniverse (approved):

mdastous-bentley:
@eringram please review

packages/itwin/map-layers/src/PreferencesBrowserStorage.ts
mdastous-bentley: @aruniverse should be good now...

eringram (approved):

PR #514, opened 2023-05-25
Update geo-tools's jsdom dependency

mdastous-bentley:
Fix security warning


mdastous-bentley:
@aruniverse sorry another one to look at.

PR #544, opened 2023-06-20
[EC3W] Fixed inconsistent styling

Guillar1:


packages/itwin/ec3-widget/src/components/HorizontalTile.scss
```diff
@@ -9,7 +9,7 @@
   justify-content: space-between;
   border-radius: 5px;
   background-color: var(--iui-color-background-2);
-  height: $iui-baseline * 4;
+  min-height: $iui-baseline * 4;
```
MindaugasButkus, line 12: How can `min-height` be more consistent than `height`?

Guillar1: Height is a pushover against `display: flex` sometimes. I pushed another way of fixing this. What is happening here is when there is not enough items to fill up the HorizontalTile (e.g. no description) and the container of the list overflows it means it does not have enough space to lay out all of its children according to their height, flex then starts to shrink them. Overriding the `height`. `min-height` prevents this. So does `flex-shrink: 0;` which stops the shrinking.

MindaugasButkus (approved):

Issue #494, opened 2023-05-18
Tree Widget: Context menu in tree components


grigasp:
Should support rendering a context menu in tree components. Menu items should be provided by consumers.

PR #549, opened 2023-06-29
Tree widget: Add ContextMenu to trees

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/494


packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTree.tsx
```diff
@@ -38,10 +39,16 @@ export interface ModelsTreeHierarchyConfiguration {
 }
 
 /**
- * Props for [[ModelsTree]] component
+ * Base props for [[ModelsTree]] component.
  * @public
  */
-export interface ModelsTreeProps extends BaseFilterableTreeProps {
+export type ModelsTreeBaseProps = BaseFilterableTreeProps & TreeContextMenuProps;
```
grigasp, line 45: do we need  / use this type anywhere besides using it below?

packages/itwin/tree-widget/src/test/trees/TreeRenderer.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      />
+    );
+
+    const nodeElement = await waitFor(() => getByText("Test Node"));
+    await user.pointer({ keys: "[MouseRight>]", target: nodeElement });
+
+    await waitFor(() => getByText("Test Item"));
+  });
+
+  it("doesn't open context menu if there are no items", async () => {
```
grigasp, line 54: please add a similar test for the case when `contextMenuItems=[() => null]`

grigasp (changes requested):

grigasp:
Just to confirm... I suppose you didn't add context menu support to `IModelContentTree` and `ExternalSourcesTree` because they don't have the same UX and `ModelsTree` and `CategoryTree` (don't use the same node renderers)?

saskliutas:
> Just to confirm... I suppose you didn't add context menu support to `IModelContentTree` and `ExternalSourcesTree` because they don't have the same UX and `ModelsTree` and `CategoryTree` (don't use the same node renderers)?

No, I think I just missed those trees. I made it possible to use ContextMenu without all the visibility tree styling so added context menu to other trees too.

packages/itwin/tree-widget/src/test/trees/TreeRenderer.test.tsx
saskliutas: Added

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTree.tsx
saskliutas: No it used as base type for `ModelsTreeProps`.

packages/itwin/tree-widget/src/components/trees/VisibilityTreeRenderer.tsx
```diff
@@ -23,10 +25,18 @@ import type { VisibilityTreeFilterInfo } from "./Common";
...
+export interface VisibilityTreeRendererProps extends TreeRendererBaseProps {
+  nodeRendererProps: VisibilityTreeNodeRendererProps;
+}
+
+/**
+ * Props for base tree node renderer used by visibility trees.
  * @public
  */
-export interface VisibilityTreeRendererProps {
+export interface VisibilityTreeNodeRendererProps {
```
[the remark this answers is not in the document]
saskliutas: Done in https://github.com/iTwin/viewer-components-react/pull/554

aruniverse:
Would it be possible to update so that the entry point for the context menu is shown on hover? I noticed for the property-grid yesterday, only shows when you right click

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTree.tsx
saskliutas: Applied suggestion

grigasp:
> Would it be possible to update so that the entry point for the context menu is shown on hover? I noticed for the property-grid yesterday, only shows when you right click

Can you give more details on why you think that's needed? Or give some example where it's implemented? I looked at several components in multiple applications and don't see this feature anywhere. Or maybe I just misunderstand what you're asking...

aruniverse:
> Can you give more details on why you think that's needed? Or give some example where it's implemented? I looked at several components in multiple applications and don't see this feature anywhere. Or maybe I just misunderstand what you're asking...

I haven't seen any examples of this in our apps either, might not have looked enough.

I found the "entry point" for our context menus a little hard to find. I noticed in google drive the other day they always show an ellipses for an entry point to their context menu. And on hover they show actions you can take like favorite and etc.

image

Here's the property grid right now with just the copy properties:
image

idk how users will know they will be able to copy the cell value unless they accidentally right click on the row

If you look at teams when you hover over a contacts profile and then proceed to hover over some information it'll show an icon to copy which i think is more intuitive 
image

I dont necessarily think we need to extract the copy to outside the context menu, just think the entry point (rather than right clicking) showing an ellipses on hover would be better UX

grigasp:
Thanks. I believe we should consult with UX first, so let's complete this PR and we can follow up with discoverability improvement afterwards (filed a separate [issue](https://github.com/iTwin/viewer-components-react/issues/556))

grigasp (approved):

PR #558, opened 2023-07-12
[G&M] Added a dropdown arrow to the Add Group button.

nyrb7:
Added a dropdown arrow to the Add Group button.

Here is what the button looks like after the change: 
https://github.com/iTwin/viewer-components-react/assets/22688247/15ef63a4-326a-42b4-93f4-280fa7fbb92a

### DIFF
--- a/common/changes/@itwin/grouping-mapping-widget/add-arrow_2023-07-12-16-13.json
+++ b/common/changes/@itwin/grouping-mapping-widget/add-arrow_2023-07-12-16-13.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "Added a dropdown arrow for the Add Group button.",
+      "type": "patch"
+    }
+  ],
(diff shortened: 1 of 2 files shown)


Guillar1:
Did you look into this component? https://itwin.github.io/iTwinUI-react/?path=/story/buttons-dropdownbutton--basic

Guillar1:
I'm wondering. Can you replace it with this component? https://itwin.github.io/iTwinUI-react/?path=/story/buttons-dropdownbutton--basic

nyrb7:
> I'm wondering. Can you replace it with this component? https://itwin.github.io/iTwinUI-react/?path=/story/buttons-dropdownbutton--basic

I do not believe it is possible to use this component and also have it look like a high-visibility button. The only style option for a DropdownButton is the default grey. 

This is what the default style looks like: 
https://github.com/iTwin/viewer-components-react/assets/22688247/15ef63a4-326a-42b4-93f4-280fa7fbb92a

Guillar1 (approved):

PR #559, opened 2023-07-12
[G&M] Added a modal to confirm a property name change.

nyrb7:
Added a modal to confirm a property name change. This helps prevent the user from experiencing issues in Custom Calculation formulas caused by renaming a property after already using it in a formula.

Here is what the modal looks like:
https://github.com/iTwin/viewer-components-react/assets/22688247/5891d1eb-3b89-45c9-9161-51ca27f78df7


packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
```diff
@@ -205,11 +208,21 @@ export const GroupPropertyAction = ({
...
+  const onSaveClick = () => {
     if (!validator.allValid()) {
       showValidationMessage(true);
       return;
     }
+    if (oldPropertyName != propertyName && oldPropertyName != "") {
+      setShowSaveModal(true);
+    }
+    else {
+      console.log("test");
```
Guillar1, line 220: Unnecessary console log.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
```diff
@@ -512,6 +525,33 @@ export const GroupPropertyAction = ({
...
+          <Button
+            onClick={() => {
+              setShowSaveModal(false);
+            }}
+            styleType='default'
+          >
+            Cancel
+          </Button>
+        </ModalButtonBar>
+      </Modal>
```
Guillar1, line 554: This looks like it can be made into its own component with its own file.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
nyrb7: This should be removed now.

packages/itwin/grouping-mapping-widget/src/widget/components/GroupPropertyAction.tsx
nyrb7: Made into a separate component.

packages/itwin/grouping-mapping-widget/src/widget/components/SaveModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+          >
+            Cancel
+          </Button>
+        </ModalButtonBar>
+      </Modal>
+    </>
+  );
+};
+
+export default SaveModal;
```
Guillar1, line 51: Let's avoid default exports. Overall, they make things more difficult. Let's use named imports from now on.

nyrb7: sounds good!

PR #561, opened 2023-07-13
[G&M] Added progress indicator when trying to Color By Group.

asharm10:
1) Changed it from Blocking Overlay to Progress Bar.
2) Used useRef to make useEffect not run when the page loads. It was running twice when getting back to the list of Groups after entering a Group with 'Color By Groups' toggle on. This made the overlay and buttons enabled and disabled twice.

https://github.com/iTwin/viewer-components-react/assets/69795197/78a63b94-ad8d-4990-8efa-6ea11548a96a


Guillar1:
image
That doesn't look right.

asharm10:
> image That doesn't look right.

Yeah it has been there since I cloned the cospace. Its not from the changes I made to the GroupsVisualization file.

Guillar1:
> > image That doesn't look right.
> 
> Yeah it has been there since I cloned the cospace. Its not from the changes I made to the GroupsVisualization file.

Oh ok definitely a bug.

Guillar1 (approved):

MindaugasButkus:
Is the user allowed to go out of the page that does the group visualization? If not - it might be annoying. Let's say user clicks on a mapping by accident and it starts visualizing groups - they would have no way of getting back to the mappings list. They would need to wait for the groups to be visualized completely until they could go back.

I would be against hiding components behind an overlay while you are loading things. Overlays are annoying.

asharm10:
> Is the user allowed to go out of the page that does the group visualization? If not - it might be annoying. Let's say user clicks on a mapping by accident and it starts visualizing groups - they would have no way of getting back to the mappings list. They would need to wait for the groups to be visualized completely until they could go back.
> 
> I would be against hiding components behind an overlay while you are loading things. Overlays are annoying.

Instead of an overlay should I add a loading spinner next to Add Group?

MindaugasButkus:
> > Is the user allowed to go out of the page that does the group visualization? If not - it might be annoying. Let's say user clicks on a mapping by accident and it starts visualizing groups - they would have no way of getting back to the mappings list. They would need to wait for the groups to be visualized completely until they could go back.
> > I would be against hiding components behind an overlay while you are loading things. Overlays are annoying.
> 
> Instead of an overlay should I add a loading spinner next to Add Group?

I don't know. It would be best to consult with the UX team. Having too many spinners is also not great.

arnobmallickbsw:
> > > image That doesn't look right.
> > 
> > 
> > Yeah it has been there since I cloned the cospace. Its not from the changes I made to the GroupsVisualization file.
> 
> Oh ok definitely a bug.

@asharm10, can you create a bug on the backlog for this?

asharm10:
> > > > image That doesn't look right.
> > > 
> > > 
> > > Yeah it has been there since I cloned the cospace. Its not from the changes I made to the GroupsVisualization file.
> > 
> > 
> > Oh ok definitely a bug.
> 
> @asharm10, can you create a bug on the backlog for this?

Already created and fixed the bug.

Guillar1:
> > > > image That doesn't look right.
> > > 
> > > 
> > > Yeah it has been there since I cloned the cospace. Its not from the changes I made to the GroupsVisualization file.
> > 
> > 
> > Oh ok definitely a bug.
> 
> @asharm10, can you create a bug on the backlog for this?

Yup, I created the bug item and assigned him to it. It was fixed last week.

PR #563, opened 2023-07-13
[G&M] Fully Show Group Creation Options

smmr-dn:
Fixed bug where it prevents group creation options to be fully shown when there is space.
image


packages/itwin/grouping-mapping-widget/src/widget/components/customUI/DefaultGroupingUI.tsx
```diff
@@ -27,7 +27,7 @@ export const defaultGroupingUI: GroupingMappingCustomUI[] = [
   },
   {
     name: "Manual",
-    displayLabel: "Manual",
+    displayLabel: "Manual Query",
```
Guillar1, line 30: Why was this renamed?

smmr-dn: I saw in both the backlog item image for this bug (there was a "..." following "Manual") and the readme file under the grouping-mapping-widget folder where it was specified as "Manual Query". I think it is clearer too now that we have space, but I should have asked first.

Guillar1: No worries. I think I renamed it to Manuel Query when I did a whole refactor of this widget then it was caught my QA asking the same question, I believe they took that screenshot when I made that mistake.

Guillar1: Yes

Guillar1 (approved):

PR #564, opened 2023-07-14
[G&M] Fixed bug not accepting Unicode characters

smmr-dn:
Allow Mapping/Grouping/Property names to take in Unicode characters
unicode-characters


packages/itwin/grouping-mapping-widget/src/widget/hooks/useValidator.ts
```diff
@@ -15,13 +15,13 @@ export const Validators = {
   OnlyBeginsWithLetterOrUnderscore: {
     message: "Name can only start with a letter or underscore.",
     rule: (val: string) => {
-      return !val.match(/^[\W\d]+/i);
+      return !val.match(/^[\p{Nd}\p{S}\p{Z}\p{Po}]+/iu);
```
MindaugasButkus, line 18: The regexes do not look right. Here is the documentation which contains a regex for OData simple identifier:
http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_SimpleIdentifier

Mapping, Group, and GroupProperty names are OData simple identifiers.

The first character must match `[\p{L}\p{Nl}_]`
All the other characters must match `[\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]`
Length should not exceed 128 characters.

smmr-dn: Thank you, Mindaugas, for pointing this out. I was struggling a bit with how String.prototype.match works. I just realized
that this function only returns an object of RegEx instead of a boolean value (which is required for the IRule interface when initializing the **validator** variable). As first I tried something like this:
image

But then the UI would return true by default since the validator returns an object. The backend instead did not take this input and returned a null name, which caused an error of "Error! Something went wrong!", but the UI was supposed to return the "Name can only start with a letter or an underscore." Therefore, I tried the approach of !val.match(...) so that the validator could output a boolean variable.

However, the Regex.prototype.test() actually returns a boolean output so I did figure out how to approach the problem properly by not sticking to match() in my second commit.

common/changes/@itwin/grouping-mapping-widget/uyen-unicode-character_2023-07-14-14-36.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "Refixed unicode bug to meet OData simple identifier requirement",
```
MindaugasButkus, line 5: Don't add a separate changelog entry. How can you fix and refix in the same version? In this case you just fixed, not refixed.

smmr-dn: Oh okay, I see! Sorry, I'm still new to rush. Do I just remove the second changelog?

Guillar1: Yes you can remove the file or modify the comment. I do like the more detailed description of this one besides the "Refixed" part.

packages/itwin/grouping-mapping-widget/src/widget/hooks/useValidator.ts
```diff
@@ -15,13 +15,15 @@ export const Validators = {
...
-      return !val.match(/^[\W\d]+/i);
+      var regexPattern = new RegExp(/^[\p{L}\p{Nl}_]+/u)
+      return regexPattern.test(val);
     },
   },
   FollowedByLettersUnderscoresAndDigits: {
     message: "Name can only contain letters, underscores, or digits",
     rule: (val: string) => {
-      return !val.match(/\W+/i);
+      var regexPattern = new RegExp(/[\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]+$/u)
```
MindaugasButkus, line 25: Do not use `var`. Use `const`.
Also, you are missing a semicolon at the end of this line. Run `rush lint` to check for these errors.

MindaugasButkus (approved):

PR #566, opened 2023-07-14
PropertyWidget & TreeWidget: Persist scroll position when switching between widgets

saskliutas:
Added `useTransientState` call to make sure that PropertyGrid and Tree scroll positions are persisted when switching between widgets.

Prerequisite: https://github.com/iTwin/appui/pull/410


grigasp (changes requested):
Need to update `README` to reflect the `TreeWidgetComponent` -> `SelectableTree` rename and the new hooks.

grigasp (approved):

aruniverse (approved):
Approving lockfile

PR #578, opened 2023-07-18
[G&M] Expands 'Selected Items' section by default.

asharm10:
https://github.com/iTwin/viewer-components-react/assets/69795197/bc4ccaa9-0027-4acd-8278-6ed428ace1af


packages/itwin/grouping-mapping-widget/src/widget/components/customUI/GroupQueryBuilderCustomUI.tsx
```diff
@@ -26,16 +26,21 @@ import { PropertyAction } from "../PropertyAction";
...
-): PresentationPropertyDataProvider => {
+): Promise<PresentationPropertyDataProvider> => {
   const dataProvider = new PresentationPropertyDataProvider({
     imodel: iModelConnection,
     ruleset: DEFAULT_PROPERTY_GRID_RULESET,
   });
   dataProvider.keys = keys;
   dataProvider.isNestedPropertyCategoryGroupingEnabled = true;
+  const data = await dataProvider.getData();
+  const selectedCategory = data.categories.find((category) => category.label === "Selected Item(s)");
```
Guillar1, line 40: I believe the category is localized, so it might not always be "Selected Item(s)". Try this `IModelApp.localization.getLocalizedString("Presentation:selectedItems.categoryLabel")`

Just an fyi, these localizations need to be initialized somewhere. Basically, the locale `json` files need to be loaded, this one is `Presentation`. I think the reason why we are seeing it localized into English is because something else is doing the initialization. Could be the other widget in the viewer, or the viewer itself. But if nothing is initializing it it would return just the locale key itself "`selectedItems.categoryLabel`". Remember that one time I was presenting when trying out the integration tests and the text looked different? It was because the locales were not loading. Just the keys themselves that pointed to the localized words. This widget is not localized completely, a lot of the text is hardcoded. The Reports Config widget is though. You can look at how all the different text is accessed rather then hardcoded.

asharm10: Changed it to "ModelApp.localization.getLocalizedString("Presentation:selectedItems.categoryLabel")". Yeah was looking into it just now. Reports Config widget has ReportsConfigWidget.ts file which handles the localization.

Guillar1 (approved):

PR #587, opened 2023-07-25
Fix various issues with MapLayers widget

mdastous-bentley:
Fix following issues:

-  Deleted MapLayersource would come back when MapLayers widget get reloaded  within the same session.
- Base map combo box becomes out of sync with the viewport state after performing  API calls.

Improved tests and fix broken ones.


packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx
```diff
@@ -248,17 +234,25 @@ export function MapLayerManager(props: MapLayerManagerProps) {
...
+      const sources: MapLayerSource[] = [];
+      const bases: MapLayerSource[] = [];
+      const addSource = (source: MapLayerSource) => source.baseMap ? bases.push(source) : sources.push(source);
+      sourceLayers?.allSource.forEach(addSource);
+      preferenceSources.forEach((source) => {
+        // Do not add duplicate
+        if (!sources.find((curSource) => source.name === curSource.name))
+          addSource(source);
+      });
+      sources.sort((a: MapLayerSource, b: MapLayerSource) => compareStrings(a.name.toLowerCase(), b.name.toLowerCase()));
```
eringram, line 247: What is the purpose of the sort here? Just trying to understand

eringram (approved):

packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx
mdastous-bentley: Simply to keep sources sorted in the UI.

aruniverse (approved):

packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx
mdastous-bentley: @eringram please resolve

Issue #592, opened 2023-08-01
Property Grid: Selected elements buttons is not shown when too many elements selected


saskliutas:
`Selected elements` button should be shown when property grid shows error message: `Too many elements selected`

PR #593, opened 2023-08-02
Property Widget: Render header when too many element selected

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/592

### DIFF
--- a/common/api/property-grid-react.api.md
+++ b/common/api/property-grid-react.api.md
@@ -72,9 +72,8 @@ export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildC
 
 // @public
 export interface FilteringPropertyGridProps extends VirtualizedPropertyGridWithDataProviderProps {
-    // (undocumented)
     autoExpandChildCategories?: boolean;
-    // (undocumented)
+    // @internal (undocumented)
     filterer: IPropertyDataFilterer;
 }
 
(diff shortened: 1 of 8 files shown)


grigasp (commented):
Is there an advantage of moving rendering of `FilteringPropertyGrid` up to `PropertyGrid` component over the following approaches?
- Pass `isOverLimit` flag to `PropertyGridContent` and render the message there.
- Render the outer `div` and header in `PropertyGrid`. This would also allow us avoid the addition to our public API.

saskliutas:
> Is there an advantage of moving rendering of `FilteringPropertyGrid` up to `PropertyGrid` component over the following approaches?
> 
> * Pass `isOverLimit` flag to `PropertyGridContent` and render the message there.
> * Render the outer `div` and header in `PropertyGrid`. This would also allow us avoid the addition to our public API.

* `PropertyGridContent` is resused between `PropertyGrid` that is hooked into unified selection and `SingleElementPropertyGrid` so introducing `isOverLimit` and rendering `Too many element` message does not make sense for `SingleEmelementPropertyGrid` use case.
* Rendering div and header in `PropertyGrid` means that we need to duplicate same code in `SingleElementPropertyGrid`

I don't see an issue introducing `dataRenderer` property on `PropertyGridContentProps` interface. This interface is public only because parts of it are used to define `PropertyGridProps` and `SingleElementPropertyGridProps` interface. `PropertyGridContent` is `internal` component and the new `dataRenderer` prop is not exposed through `PropertyGridProps` or `SingleElementPropertyGridProps`.

grigasp:
Thanks, that makes sense.

> I don't see an issue introducing `dataRenderer` property on `PropertyGridContentProps` interface. This interface is public only because parts of it are used to define `PropertyGridProps` and `SingleElementPropertyGridProps` interface. `PropertyGridContent` is internal component and the new `dataRenderer` prop is not exposed through `PropertyGridProps` or `SingleElementPropertyGridProps`.

Should the property be marked `@internal` then?

saskliutas:
> Thanks, that makes sense.
> 
> > I don't see an issue introducing `dataRenderer` property on `PropertyGridContentProps` interface. This interface is public only because parts of it are used to define `PropertyGridProps` and `SingleElementPropertyGridProps` interface. `PropertyGridContent` is internal component and the new `dataRenderer` prop is not exposed through `PropertyGridProps` or `SingleElementPropertyGridProps`.
> 
> Should the property be marked `@internal` then?

Yes, I was thinking about it. Probably all properties from `PropertyGridContentProps` that are omitted on `PropertyGridProps` should be marked `internal`. I will add that.

grigasp (approved):

Issue #595, opened 2023-08-03
Property Widget: Let override default context menu item actions


saskliutas:
Currently there is no way to override actions of default context menu items. There might be useful to have an ability to some additional stuff (like cache or clear cache) when one of default context menu items is clicked.

PR #597, opened 2023-08-03
Property Widget: Add ability to override default context menu items

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/595

### DIFF
--- a/packages/itwin/property-grid/src/hooks/UseContextMenu.tsx
+++ b/packages/itwin/property-grid/src/hooks/UseContextMenu.tsx
@@ -4,7 +4,6 @@
 *--------------------------------------------------------------------------------------------*/
 
 import { Fragment, useState } from "react";
-import { assert } from "@itwin/core-bentley";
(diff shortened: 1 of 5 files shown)


packages/itwin/property-grid/src/hooks/UseContextMenu.tsx
```diff
@@ -81,18 +92,24 @@ export interface FavoritePropertiesContextMenuItemProps extends ContextMenuItemP
...
+  const defaultAction = async () => Presentation.favoriteProperties.add(field, imodel, currentScope);
+
   return (
     <PropertyGridContextMenuItem
       id="add-favorite"
       onSelect={async () => {
-        assert(field !== undefined);
-        await Presentation.favoriteProperties.add(field, imodel, currentScope);
+        if (onSelect) {
+          onSelect(defaultAction);
```
grigasp, line 108: `defaultAction` here is async, but we pass it as non-async to `onSelect`. Maybe `onSelect` should accept an async function to allow consumers await on the `defaultAction` if necessary?

grigasp (approved):

Issue #611, opened 2023-08-08
Property widget: "Selected elements" button doesn't show up when selection changes rapidly


grigasp:
**Repro:**
1. Make sure the tree allows selecting nodes:
   ```
   diff --git a/apps/test-viewer/src/UiProvidersConfig.tsx b/apps/test-viewer/src/UiProvidersConfig.tsx
   index a2e9e33..f62d25d 100644
   --- a/apps/test-viewer/src/UiProvidersConfig.tsx
   +++ b/apps/test-viewer/src/UiProvidersConfig.tsx
   @@ -11,6 +11,7 @@ import { SampleSpatialTree } from "./components/SampleSpatialTree";
    import { DefaultMapFeatureInfoTool, FeatureInfoUiItemsProvider, MapLayersPrefBrowserStorage, MapLayersUI, MapLayersUiItemsProvider } from "@itwin/map-layers";
    import { GeoTools, GeoToolsAddressSearchProvider } from "@itwin/geo-tools-react";
    import { MapLayersFormats } from "@itwin/map-layers-formats";
   +import { SelectionMode } from "@itwin/components-react";
   
    export interface UiProvidersConfig {
      initialize: () => Promise<void>;
   @@ -63,7 +64,7 @@ const configuredUiItems = new Map<string, UiItem>([
            trees: [{
              id: ModelsTreeComponent.id,
              getLabel: ModelsTreeComponent.getLabel,
   -          render: () => <ModelsTreeComponent />
   +          render: () => <ModelsTreeComponent selectionMode={SelectionMode.Extended} selectionPredicate={() => true} />
            }, {
              id: CategoriesTreeComponent.id,
              getLabel: CategoriesTreeComponent.getLabel,
   ```
2. Select multiple nodes in a tree - notice the "selected elements" button shows up.
3. Ctrl+double click one of the nodes to deselect and quickly select it again - notice the "selected elements" button is not there, although we have multiple elements selected.

   ![2023-08-08_09-29-07](https://github.com/iTwin/viewer-components-react/assets/35135765/69678b2a-e1ee-46b2-832a-83def3ca6771)

**Expected behavior:** The "selected elements" button is visible always when more than 1 element is selected.

PR #613, opened 2023-08-08
Property widget: Fix selection handling

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/611


grigasp (approved):
LGTM. I believe you may need to regenerate e2e test screenshots after merging with master.

packages/itwin/property-grid/playwright.config.ts
```diff
@@ -75,6 +75,13 @@ export default defineConfig({
     // },
   ],
 
+  expect: {
+    toHaveScreenshot: {
+      /* Allow some difference between screenshots as rendering might differ on different machines. */
+      maxDiffPixelRatio: 0.01
```
grigasp, line 81: why do we need this? 😞

packages/itwin/property-grid/src/e2e-tests/PropertyGrid.test.ts-snapshots/should-match-image-snapshot-stage-panel-size---narrow-multiple-elements-selected-1-chromium-win32.png
grigasp, line 1: looks like property labels and values are in different fonts - mind having a look at this?

saskliutas: They might be in different font or maybe just different size. I tried to find font similar to the one @medikornov used to created initial snapshots. 

From what I found out it looks like that different machines might render same page differently https://github.com/microsoft/playwright/issues/20097#issuecomment-1419220152
Probably the best solution would be to have a docker container and run e2e test in it.

packages/itwin/property-grid/playwright.config.ts
saskliutas: This should answer https://github.com/iTwin/viewer-components-react/pull/613/files/f2c46feeef0e972e253035a5b75ae86b44cf8993..460b500656c196b72374d2e2ca83fa59ee6b85c0#r1288295845

packages/itwin/property-grid/src/e2e-tests/PropertyGrid.test.ts-snapshots/should-match-image-snapshot-stage-panel-size---narrow-multiple-elements-selected-1-chromium-win32.png
grigasp: That's sad. I filed https://github.com/iTwin/viewer-components-react/issues/614 to handle the situation, in the meantime let's keep the tolerance threshold.

PR #615, opened 2023-08-10
[G&M] Grouping Component Props 

Guillar1:


packages/itwin/grouping-mapping-widget/src/grouping-mapping-widget.ts
```diff
@@ -12,7 +12,9 @@ export * from "@itwin/insights-client";
 
 /** Internal components for custom UIs */
 export { Mappings, MappingsProps } from "./widget/components/Mapping";
+export { useGroupingsOperations, GroupingsOperationsProps } from "./widget/components/hooks/useGroupingsOperations";
 export { MappingAction, MappingActionProps } from "./widget/components/MappingAction";
+export { GroupingsView, GroupingsViewProps } from "./widget/components/GroupingsView";
```
arnobmallickbsw, line 17: kinda unfortunate situation we're ending up in with use of both terms 'Groupings' and 'Groups'...
is this a good time to fix that? before we get more usage of the package for components and not just the UIProvider?

Guillar1: Yeah I agree, let's rip the band aid off right now before their is more usage.

arnobmallickbsw (approved):

PR #434, opened 2023-04-18
Tree-Widget: Fix invert button in categories tree

simasjar:
closes #405

PR #621, opened 2023-08-16
[Measure Tools] Switch to Toolbar instead of ToolbarComposer for Measure Actions toolbar

wilmaier:
From Raphael:

PopupToolbar should not use ToolbarComposer as this is meant to be filled with UiItemProviders. #434 will allow proper handling of conditional toolbar items (which I guess is why ToolbarComposer was used in the first place) and so the Toolbar component should be enough for this usecase.

Hi! The Toolbar component have been updated to properly handle ConditionalXXXValues type of properties on toolbars, which should be a better fit for your need than the ToolbarComposer. This fix was released today on 4.4.0, give it a try and let us know if this fix your issue, even if it does not add new ToolbarUsage.

image


simnorm (approved):

packages/itwin/measure-tools/src/widgets/MeasurementActionToolbar.tsx
```diff
@@ -444,8 +444,6 @@ export class MeasurementActionToolbar {
     });
     return <PopupToolbar
```
aruniverse, line 445: format

aruniverse (commented):
missing rush change; patch

aruniverse:
>  This fix was released today on 4.4.0

this means you need to bump the minimum appui deps from ^4.0.0 to ^4.4.0;
please update the peers and dev deps
this is technically breaking, so actually make the rush change a minor bump

common/changes/@itwin/measure-tools-react/fix-measurement-action-toolbar_2023-08-16-20-55.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/measure-tools-react",
+      "comment": "[Measure Tools] Switch to Toolbar instead of ToolbarComposer for Measure Actions toolbar",
```
aruniverse, line 5:
```suggestion
      "comment": "Switch to Toolbar instead of ToolbarComposer for Measure Actions toolbar",
```

aruniverse (approved):

PR #645, opened 2023-09-12
[LCA] Migrated UI items for iTwin 4.x upgrade for One Click LCA Widget

smmr-dn:
Upgraded CSS variables and replaced deprecated packages/components for One Click LCA widget.


smmr-dn:
Conflicts in local branch

PR #647, opened 2023-09-13
Expand ModelsTree component to accept visibilityHandler as a function

ViliusBa:
Expand the _ModelsTree_ component to accept `visibilityHandler` as a callback which returns a `visibilityHandler` object. 

With these changes ModelTree creation workflows look like this:
- If `visibilityHandler` is passed as an object, we pass it on and use it without taking care of its disposal.
- If `visibilityHandler` is passed as a function, we call the function which creates the object and take care of its disposal.

Also, had to change the way that we are mocking `visibilityHandler` in our _ModelTree_ tests since it was mocking the `visibilityHandler` as a function and not as an object, which led to false positives when checking what is the type of the `visibilityHandler`.


packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -175,34 +175,35 @@ describe("ModelsTree", () => {
         const node = createModelNode();
         setupDataProvider([node]);
 
-        visibilityHandlerMock.setup(async (x) => x.getVisibilityStatus(moq.It.isAny())).returns(async () => ({ state: "hidden", isDisabled: false })).verifiable(moq.Times.exactly(2));
-        const result = render(<ModelsTree {...sizeProps} iModel={imodelMock.object} modelsVisibilityHandler={visibilityHandlerMock.object} activeView={mockViewport().object} />);
+        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "hidden", isDisabled: false });
+        const getVisibilityStatusSpy = sinon.spy(visibilityHandlerMock, "getVisibilityStatus");
```
grigasp, line 179:
```suggestion
        const getVisibilityStatusSpy = sinon.stub().resolves({ state: "hidden", isDisabled: false });
        visibilityHandlerMock.getVisibilityStatus = getVisibilityStatusSpy;
```

This suggestion also applies to some other tests.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -252,13 +254,36 @@ describe("ModelsTree", () => {
         }));
       });
 
+      it("should call visibilityHandler with correct parameters when it is passed as a function", async () => {
+        const node: TreeNodeItem = { id: "test", label: PropertyRecord.fromString("test-node"), isCheckboxVisible: true };
+        setupDataProvider([node]);
+        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "hidden" });
+        const getVisibilityStatusSpy = sinon.spy(visibilityHandlerMock, "getVisibilityStatus");
+        const visibilityHandlerCallback = (_: ModelsVisibilityHandlerProps) => {return visibilityHandlerMock;};
```
grigasp, line 262: This test is not doing what it says - it says it's verifying that the factory is called with correct props, but really it's checking if the handler's `getVisibilityStatus` function is called with correct props. We should spy on the factory instead.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -252,13 +254,36 @@ describe("ModelsTree", () => {
...
+        await waitFor(() => result.getByText("test-node"), { container: result.container });
+        expect(getVisibilityStatusSpy.calledOnceWith(node));
+      });
+
+      it("should dispose of the visibilityHandler when ModelTree component is unmounted", async () => {
+        setupDataProvider([{ id: "test", label: PropertyRecord.fromString("test-node"), isCheckboxVisible: true }]);
+        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "hidden" });
+        visibilityHandlerMock.dispose = () => {};
+        const disposeSpy = sinon.spy(visibilityHandlerMock, "dispose");
+        const visibilityHandlerCallback = (_: ModelsVisibilityHandlerProps) => {return visibilityHandlerMock;};
```
grigasp, line 273:
```suggestion
        const visibilityHandlerCallback = () => visibilityHandlerMock;
```

grigasp (changes requested):
Need to run `extract-api`.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -175,34 +175,35 @@ describe("ModelsTree", () => {
...
+        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "visible", isDisabled: false });
         visibilityChangeEvent.raiseEvent();
         await waitFor(() => {
           const renderedNode = result.getByTestId("tree-node");
           if (!isNodeChecked(renderedNode))
             throw new Error("expecting checked node");
           return renderedNode;
         });
-        visibilityHandlerMock.verifyAll();
+        expect(getVisibilityStatusSpy.calledTwice);
```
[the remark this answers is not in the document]
ViliusBa: I was sure that it is the first point, but now that I think of it, I can't find anything that would support this statement. (Except that the test passes.) I will change it to something that is more clear on how it works.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -254,23 +255,22 @@ describe("ModelsTree", () => {
...
 
-      it("should call visibilityHandler with correct parameters when it is passed as a function", async () => {
+      it("should invoke visibilityHandler factory when it is passed as a function", async () => {
         const node: TreeNodeItem = { id: "test", label: PropertyRecord.fromString("test-node"), isCheckboxVisible: true };
         setupDataProvider([node]);
-        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "hidden" });
-        const getVisibilityStatusSpy = sinon.spy(visibilityHandlerMock, "getVisibilityStatus");
-        const visibilityHandlerCallback = (_: ModelsVisibilityHandlerProps) => {return visibilityHandlerMock;};
+        const visibilityHandlerCallback = () => {return visibilityHandlerMock;};
+        const visibilityHandlerSpy = sinon.spy(visibilityHandlerCallback);
```
grigasp, line 262:
```suggestion
        const visibilityHandlerSpy = sinon.stub().returns(visibilityHandlerMock);
```

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -252,13 +255,35 @@ describe("ModelsTree", () => {
...
+        const result = render(<ModelsTree {...sizeProps} iModel={imodelMock.object} modelsVisibilityHandler={visibilityHandlerCallback} activeView={mockViewport().object} />);
+        await waitFor(() => result.getByText("test-node"), { container: result.container });
+        expect(visibilityHandlerSpy.calledOnce);
+      });
+
+      it("should dispose of the visibilityHandler when ModelTree component is unmounted", async () => {
+        setupDataProvider([{ id: "test", label: PropertyRecord.fromString("test-node"), isCheckboxVisible: true }]);
+        visibilityHandlerMock.getVisibilityStatus = async () => ({ state: "hidden" });
+        visibilityHandlerMock.dispose = () => {};
+        const disposeSpy = sinon.spy(visibilityHandlerMock, "dispose");
```
grigasp, line 272: I think you could actually even do this:
```
visibilityHandlerMock.dispose = sinon.spy();
```
And then
```
expect(visibilityHandlerMock.dispose).to.be.calledOnce;
```

The same can be applied to the tests above where you're stubing and spying on `getVisibilityStatus`.

grigasp (changes requested):

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTree.test.tsx
```diff
@@ -258,23 +254,21 @@ describe("ModelsTree", () => {
       it("should invoke visibilityHandler factory when it is passed as a function", async () => {
         const node: TreeNodeItem = { id: "test", label: PropertyRecord.fromString("test-node"), isCheckboxVisible: true };
         setupDataProvider([node]);
-        const visibilityHandlerCallback = () => {return visibilityHandlerMock;};
-        const visibilityHandlerSpy = sinon.spy(visibilityHandlerCallback);
-        const result = render(<ModelsTree {...sizeProps} iModel={imodelMock.object} modelsVisibilityHandler={visibilityHandlerCallback} activeView={mockViewport().object} />);
+        const visibilityHandlerSpy = sinon.stub().returns(visibilityHandlerMock);
+        const result = render(<ModelsTree {...sizeProps} iModel={imodelMock.object} modelsVisibilityHandler={visibilityHandlerSpy} activeView={mockViewport().object} />);
         await waitFor(() => result.getByText("test-node"), { container: result.container });
         expect(visibilityHandlerSpy.calledOnce);
```
grigasp, line 260: I believe this doesn't check anything

```suggestion
        expect(visibilityHandlerSpy).to.be.calledOnce;
```

grigasp (approved):

PR #655, opened 2023-09-23
[G&M] Expose extraction log messages

smmr-dn:
This change creates a new status icon button that shows a color representation of the status of the most recent extraction job. It also creates a modal that opens when one of the extraction logs contains issues.
[custom filter.webm](https://github.com/iTwin/viewer-components-react/assets/56598021/e1f6b1c7-379b-4f95-8463-e7b68d68f58f)


packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,82 @@ const fetchMappings = async (
...
+  extractionClient: IExtractionClient,
+  setIconStatus: React.Dispatch<React.SetStateAction<"negative" | "positive" | "warning">>,
+  setIconMessage: React.Dispatch<React.SetStateAction<string>>,
+  setExtractionMessageData: React.Dispatch<React.SetStateAction<ExtractionMessageData[]>>
+) => {
+  setIconStatus("warning");
+  setIconMessage("Extraction status pending.");
+  try {
+    const accessToken = await getAccessToken();
+    const extractions = await extractionClient.getExtractionHistory(accessToken, iModelId);
```
MindaugasButkus, line 46: Since you are only interested in the latest extraction, you can pass the `top` value of `1` to `getExtractionHistory`. This will request the API only for the latest extraction history entry.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,82 @@ const fetchMappings = async (
...
+    const jobId = extractions[0].jobId;
+    const status = await extractionClient.getExtractionStatus(accessToken, jobId);
+    if (status.containsIssues) {
+      setIconStatus("negative");
+      setIconMessage("Extraction contains issues. Click to view extraction logs.");
+      let logs = await extractionClient.getExtractionLogs(accessToken, jobId);
```
MindaugasButkus, line 52: Use `const`. Filter the logs in the same statement. If you need this for debugging you will always have the Network tab in dev tools.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,82 @@ const fetchMappings = async (
...
+            level: log.level,
+            message: String(log.message),
+          }
+        ));
+      setExtractionMessageData(extractionMessageData);
+    } else {
+      setIconStatus("positive");
+      setIconMessage("Extraction Successful.");
+    }
+  } catch (error: any) { }
```
MindaugasButkus, line 68: What is the IconStatus or IconMessage when error occurs? Let's say the `getExtractionLogs` fails for some reason. The icon status will stay negative but you will have no data to display.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  extractionMessageData: ExtractionMessageData[];
+  timestamp: string;
+}
+
+export const ExtractionMessageModal = ({ isOpen, onClose, extractionMessageData, timestamp }: ExtractionMessageModalProps) => {
+  const [formattedTimestamp, setFormattedTimestamp] = useState<string>("");
+  useEffect(() => {
+    formatTimestamp;
```
MindaugasButkus, line 25: What does this do?

smmr-dn: The hook will call the function where the stamp is formatted as soon as it retrieves the timestamp at the moment the component mounts.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: I have updated this part.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: I have changed let to const.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: I have updated this part.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
Guillar1: The hook is not doing anything. `formatTimestamp` is not being invoked.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+    clear: "Clear",
+  }), []);
+  const columns = useMemo(
+    (): Column<CreateTypeFromInterface<ExtractionMessageData>>[] => [
+      {
+        id: "category",
+        Header: "Category",
+        accessor: "category",
+        fieldType: "text",
+        Cell: ({ value }: CellProps<ExtractionMessageData>) => String(value),
```
Guillar1, line 38: This looks redundant.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+                ) : level === "Info" ? (
+                  <StatusIcon status='informational' />
+                ) : <StatusIcon status='trace' />
+              }
+            >
+              {level}
+            </DefaultCell>
+          );
+        },
+        Cell: ({ value }: CellProps<ExtractionMessageData>) => String(value),
```
Guillar1, line 66: This might also be redundant.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,82 @@ const fetchMappings = async (
...
+  try {
+    const accessToken = await getAccessToken();
+    const extractions = await extractionClient.getExtractionHistory(accessToken, iModelId);
+    const jobId = extractions[0].jobId;
+    const status = await extractionClient.getExtractionStatus(accessToken, jobId);
+    if (status.containsIssues) {
+      setIconStatus("negative");
+      setIconMessage("Extraction contains issues. Click to view extraction logs.");
+      let logs = await extractionClient.getExtractionLogs(accessToken, jobId);
+      logs = logs.filter((log) => log.message != null);
```
Guillar1: This is a loose inequality check. Please use `!==`

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,82 @@ const fetchMappings = async (
...
 export const useMappingsOperations = ({ iModelId, getAccessToken, mappingClient }: MappingsOperationsProps) => {
   const [showImportModal, setShowImportModal] = useState<boolean | undefined>(false);
   const [showDeleteModal, setShowDeleteModal] = useState<Mapping | undefined>(undefined);
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [mappings, setMappings] = useState<Mapping[]>([]);
+  const extractionClient = useExtractionClient();
   const [errorMessage, setErrorMessage] = useState<string | undefined>(undefined);
   const [isTogglingExtraction, setIsTogglingExtraction] = useState<boolean>(false);
+  const [iconStatus, setIconStatus] = useState<"negative" | "positive" | "warning">("warning");
+  const [iconMessage, setIconMessage] = useState<string>("");
```
Guillar1, line 94: I think `iconStatus` and `iconMessage` can be merged into one state.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
```diff
@@ -30,22 +31,85 @@ const fetchMappings = async (
...
+  setIconStatus: React.Dispatch<React.SetStateAction<"negative" | "positive" | "warning">>,
+  setIconMessage: React.Dispatch<React.SetStateAction<string>>,
+  setExtractionMessageData: React.Dispatch<React.SetStateAction<ExtractionMessageData[]>>
+) => {
+  setIconStatus("warning");
+  setIconMessage("Extraction status pending.");
+  try {
+    const accessToken = await getAccessToken();
+    const extractions = await extractionClient.getExtractionHistory(accessToken, iModelId, 1);
+    const jobId = extractions[0].jobId;
```
rkkr, line 47: When is `fetchExtractionStatus` called? I have an array index out of bounds feeling,

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }), []);
+  const columns = useMemo(
+    (): Column<CreateTypeFromInterface<ExtractionMessageData>>[] => [
+      {
+        id: "category",
+        Header: "Category",
+        accessor: "category",
+        fieldType: "text",
+        Cell: ({ value }: CellProps<ExtractionMessageData>) => String(value),
+        Filter: ExtractionLogCustomFilter,
```
rkkr, line 39: Where does `Filter` and `filter` come from? Naming looks suspicious.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
MindaugasButkus: Good catch! If the iModel had no extractions this will throw.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
Guillar1: I also think it would be easier to read if that logic was instead in the `useEffect` itself rather then in that `useMemo`.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
smmr-dn: To what I understand from the iTwin UIDoc, `Filter` is the JSX element that contains the filter options and selection type (such as checkboxes or radio buttons) since in this case it's a customized filter. `filter` is the type of comparison, such as "text" for string matching and "equals" for equivalent values.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
smmr-dn: Oh, you're right, since it's in a `useMemo`.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: `fetchExtractionStatus` is called when the component mounts since the `useMappingsOperations.ts` is initialized inside `Mappings.tsx`. I wrapped it in an if that changes `extractionIconData`, but I'm not sure this is an efficient way to deal with this issue.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: I have compressed these two in the object `extractionIconData`.

packages/itwin/grouping-mapping-widget/src/widget/components/hooks/useMappingsOperations.ts
smmr-dn: I have had this fixed.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
smmr-dn: I have had this removed.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
smmr-dn: I have had this removed.

packages/itwin/grouping-mapping-widget/src/widget/components/ExtractionMessageModal.tsx
smmr-dn: I have moved it inside `useEffect`.

MindaugasButkus (approved):

PR #656, opened 2023-09-26
Improved UI to better control the state of multiple map layers in a single operation

mdastous-bentley:
image

1. You can now easily unlink all map layers (or a sub-set of it), by clicking checkboxes.
2. New buttons were added to 'Hide', 'Show', and 'Invert' all map-layers in a single operation.
3. For better UI clarity, individual map layer menu are hidden until the mouse is moved over a specific entry of the list.

_Also part of this PR, linting packages /config were added to have a better feedback of React code in VS Code._


mdastous-bentley:
**! NOTE:** Change log type is set to 'none' on purpose because I want to publish a minor with another PR right after this.

eringram (approved):

mdastous-bentley:
@ben-polinsky Do you think the failing build is a fluke or something that be investigated?

ben-polinsky:
> @ben-polinsky Do you think the failing build is a fluke or something that be investigated?

Looks like there's an issue with mac tests potentially being flaky - I see failures from various runs unrelated to this PR. It should be investigated but probably shouldn't hold up this PR unless you cannot get the build to pass at all.

ben-polinsky (approved):

PR #664, opened 2023-10-03
Several UI fixes and fine tuning

mdastous-bentley:
**'Select Layers To Add' dialog:**
- No longer resize the dialog when a status message appear/disappear.
- Adjusted margins
- Fixed a bug where adding/remove a filter would reset the current set of selected layers.

**Other fixes:**
- Uniformized terminology to 'Add' layer instead of 'Create' or 'Attach'
- Added margin and removed border to the 'tech preview' badge in the type select control.


packages/itwin/map-layers/src/ui/widget/MapUrlDialog.scss
```diff
@@ -83,8 +79,45 @@ $default-font-size: --iui-font-size-1;
   white-space: pre-wrap;
 }
 
+
+.map-layer-source-footer {
+  display: flex;
+  margin-top: 10px;
+  }
```
eringram, line 86: Nitpick: spacing could be more consistent in this file

eringram (approved):

packages/itwin/map-layers/src/ui/widget/MapUrlDialog.scss
mdastous-bentley: @eringram done.

PR #670, opened 2023-10-09
[G&M] Limited one selection for radio buttons

smmr-dn:
Only allowed one selection for radio buttons instead of allowing multiple options selected.


common/changes/@itwin/grouping-mapping-widget/uyen-radio-button-selection_2023-10-09-23-19.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/grouping-mapping-widget",
+      "comment": "Only allowed one selection for radio buttons.",
```
Guillar1, line 5: Be a little bit more specific, like for which component?

Guillar1 (approved):

PR #675, opened 2023-10-18
[G&M] Extraction Status Icon Consistency

smmr-dn:
Kept extraction status icon consistent throughout Mapping and Group component navigation. Status can be reloaded using the refresh icon.

extraction-icon


packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -33,29 +36,25 @@ const fetchExtractionStatus = async (
   iModelId: string,
   getAccessToken: GetAccessTokenFn,
   extractionClient: IExtractionClient,
-  setExtractionIconData: React.Dispatch<React.SetStateAction<ExtractionIconData>>,
-  setExtractionMessageData: React.Dispatch<React.SetStateAction<ExtractionMessageData[]>>
+  setExtractionStatusIcon: (extractionStatusIcon: ExtractionStatusData | ((extractionStatusIcon: ExtractionStatusData) => ExtractionStatusData)) => void,
+  setExtractionMessageData: (extractionMessageData: ExtractionMessageData[] | ((extractionMessageData: ExtractionMessageData[]) => ExtractionMessageData[])) => void
```
Guillar1, line 40: Do we still need the previous state callback type?

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -33,29 +36,25 @@ const fetchExtractionStatus = async (
...
   try {
+    setExtractionStatusIcon({
+      iconStatus: undefined,
+      iconMessage: "Loading...",
+    });
     const accessToken = await getAccessToken();
-    const extractions = await extractionClient.getExtractionHistory(accessToken, iModelId, 1);
-    if (extractions.length === 0) {
-      setExtractionIconData({
+    const extraction = await extractionClient.getExtractionHistory(accessToken, iModelId, 1);
```
MindaugasButkus, line 48: Use `getExtractionHistoryIterator` here with `top = 1` and take only the first history entry. Using `getExtractionHistory` causes the whole history to be loaded in pages of size 1 (top = 1).

Using `getExtractionHistoryIterator` should produce better results, since it loads pages as you iterate, so if you take only 1 element, it will query only once.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
smmr-dn: I have removed the callback type.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
smmr-dn: I have changed it up to `getExtractionHistoryIterator`.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
MindaugasButkus: What about the `setExtractionStatusIcon`? Do we need the previous state callback there?

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -33,29 +36,25 @@ const fetchExtractionStatus = async (
...
+    setExtractionStatusIcon({
+      iconStatus: undefined,
+      iconMessage: "Loading...",
+    });
     const accessToken = await getAccessToken();
-    const extractions = await extractionClient.getExtractionHistory(accessToken, iModelId, 1);
-    if (extractions.length === 0) {
-      setExtractionIconData({
+    const extraction = extractionClient.getExtractionHistoryIterator(accessToken, iModelId, 1);
+    if((await extraction.next()).done){
```
MindaugasButkus, line 49: Do `extraction.next()` only once and assign the result to a `const`. Now you should be getting an unwanted result - the second to last extraction jobId should be set on line 55. The first `next` reads the latest extraction jobId. The second next will read the second and so on.

packages/itwin/grouping-mapping-widget/src/components/context/ExtractionStatusDataContext.ts
```diff
@@ -0,0 +1,50 @@
...
+  setExtractionStatusIcon: (extractionStatusIcon: ExtractionStatusData | ((extractionStatusIcon: ExtractionStatusData) => ExtractionStatusData)) => void;
+  setExtractionMessageData: (extractionMessageData: ExtractionMessageData[]) => void;
+}
+
+export const ExtractionStatusDataContext = createContext<IExtractionStatusDataProps>({
+  extractionStatusIcon: {
+    iconStatus: undefined,
+    iconMessage: "",
+  },
+  extractionMessageData: [{
```
MindaugasButkus, line 31: Wouldn't it be better to have the default `extractionMessageData` set to an empty array? Is there anything that relies on `extractionMessageData` being a non-empty array?

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
```diff
@@ -114,13 +114,13 @@ export const MappingsView = ({
           </div>
           <div className="gmw-button-spacing">
             <ExtractionStatusIcon
-              iconStatus={extractionIconData.iconStatus}
+              iconStatus={extractionStatusData.iconStatus}
```
MindaugasButkus, line 117: What happens when the `iconStatus` is undefined?

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -104,31 +95,35 @@ export const useMappingsOperations = ({ iModelId, getAccessToken, mappingClient
...
   const [errorMessage, setErrorMessage] = useState<string | undefined>(undefined);
   const [isTogglingExtraction, setIsTogglingExtraction] = useState<boolean>(false);
-  const [extractionIconData, setExtractionIconData] = useState<ExtractionIconData>({
-    iconStatus: "warning",
-    iconMessage: "",
-  });
+  const {extractionStatusIcon, extractionMessageData, setExtractionStatusIcon, setExtractionMessageData} = useExtractionStatusDataContext();
   const [showExtractionMessageModal, setShowExtractionMessageModal] = useState<boolean>(false);
-  const [extractionMessageData, setExtractionMessageData] = useState<ExtractionMessageData[]>([]);
+  const previousStatusIcon = useRef<ExtractionStatusData>();
```
MindaugasButkus, line 100: Could you explain why you need to know the `previousStatusIcon`?

smmr-dn: This is to check whether there is any status (not undefined) previously stored in context to decide whether we should call `fetchExtractionStatus` to retrieve the status again.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
smmr-dn: I didn't use it so I just had it removed.

packages/itwin/grouping-mapping-widget/src/components/context/ExtractionStatusDataContext.ts
smmr-dn: I don't think it would affect anything, so I have changed it to an empty array.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
smmr-dn: Oh, I just realized that logic. I just assigned `extraction.next()` to `latestExtractionResult`.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
smmr-dn: `iconStatus` is initialized in context so whenever it's blue it appears in color blue.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -104,31 +96,35 @@ export const useMappingsOperations = ({ iModelId, getAccessToken, mappingClient
...
+        setExtractionStatusIcon,
+        setExtractionMessageData
+      );
+    }
+  }, [iModelId, getAccessToken, extractionClient, isLoading, setExtractionStatusIcon, setExtractionMessageData]);
 
   const refresh = useCallback(async () => {
     setMappings([]);
+    void fetchExtractionStatus(iModelId, getAccessToken, extractionClient, setExtractionStatusIcon, setExtractionMessageData);
     await fetchMappings(setMappings, iModelId, setIsLoading, getAccessToken, mappingClient, setErrorMessage);
```
Guillar1, line 126: You can wrap both of these calls into a `Promise.all`

Guillar1 (approved):

Issue #676, opened 2023-10-20
Property Widget: Add support for react 18


grigasp:
- Ensure the widget works fine with React 18:
  - Bump react dependency of test viewer to version 18.
  - Add additional e2e tests to make sure different pieces of the property widget are exercised.
- Bump React peer dependency to `18`.

Issue #672, opened 2023-10-16
Tree Widget: Add support for react 18


saskliutas:
Tree widget does not work with react 18 in `StrictMode` due to double rendering.

muthubentley:
@saskliutas All the below packages doesnt have react 18 as their peer dependency. 
1. @itwin/breakdown-trees-react
2. @itwin/measure-tools-react
3. @itwin/property-grid-react
4. @itwin/tree-widget-react

All of them belongs to this repo. Will you be able to fix all of them with this issue or do you need a separate issue for fixing each package?

grigasp:
> @saskliutas All the below packages doesnt have react 18 as their peer dependency.
> 
> 1. @itwin/breakdown-trees-react
> 2. @itwin/measure-tools-react
> 3. @itwin/property-grid-react
> 4. @itwin/tree-widget-react
> 
> All of them belongs to this repo. Will you be able to fix all of them with this issue or do you need a separate issue for fixing each package?

@muthubentley I believe it's best if we have an issue per package. I filed [one for the property grid](https://github.com/iTwin/viewer-components-react/issues/676).

**Note:** We don't own all the mentioned packages, so we're only going to look into `tree-widget-react` and `property-grid-react`. Other teams will need to look into others (adding @aruniverse).

muthubentley:
Thanks @grigasp
Created new issues  for @itwin/breakdown-trees-react and @itwin/measure-tools-react
https://github.com/iTwin/viewer-components-react/issues/677
https://github.com/iTwin/viewer-components-react/issues/678

PR #679, opened 2023-10-20
[Tree Widget & Property Grid]: Add support for react 18

saskliutas:
Added `^18.0.0` to react peerDependency.
Added custom `useDisposable` hook implementation that works in StrictMode with React 18.

Couldn't bump react dev dependencies to 18 because build for other packages started failing. Might be because of mismatching versions of `@types/react`. Looks like react dev dependency should be bump for all packages in repo at the same time.

Closes https://github.com/iTwin/viewer-components-react/issues/676
Closes https://github.com/iTwin/viewer-components-react/issues/672


packages/itwin/property-grid/src/hooks/UseDisposable.ts
```diff
@@ -0,0 +1,29 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import type { IDisposable } from "@itwin/core-bentley";
+import { useEffect, useRef, useState } from "react";
+
+// TODO: Remove when `useDisposable` from `@itwin/core-react` works fine with React 18 strict mode
+/** @internal */
+export function useDisposable<TDisposable extends IDisposable>(factory: () => TDisposable): TDisposable {
```
aruniverse, line 11: Could you copy this into AppUi?

aruniverse (approved):

grigasp (commented):
@aruniverse @calebmshafer, we can't run any tests using React 18 until every package in the repo upgrades to it. Until that happens we have to just _trust_ that whatever works with React 17 should also work with React 18... 

Honestly, I don't like that situation and I'd like to defer telling consumers we support React 18 (through peer deps) until we have our tests running against it. Which means we'd wait for all other packages in this repo to upgrade. What do you think?

packages/itwin/property-grid/src/hooks/UseDisposable.ts
saskliutas: Yes, I have local changes made in AppUI repo. Will open PR soon as I made changes to some other hooks in AppUI too.

saskliutas: https://github.com/iTwin/appui/pull/546/files

saskliutas:
> > defer telling consumers we support React 18 (through peer deps)
> 
> I'm in agreement, I have become not a fan of saying we support React 18, because a lot of our pkgs do break under strict mode, so we have iffy support.
> 
> Would we be able to use [rush variants](https://rushjs.io/pages/advanced/installation_variants/) to test against both React 17 & 18? AppUI is using this to test against both itwinjs-core ^3.6 and 4.x, [see](https://github.com/iTwin/appui/pull/187)

I don't think `rush variants` will work in this case as running tests against React 18 code changes will be required compared to running against React 17 because `renderHook` was moved to `@testing-library/react` in version `13.1.0` (https://www.npmjs.com/package/@testing-library/react-hooks#a-note-about-react-18-support). So when testing hooks against React 17 we need to use `import { renderHook } from "@testing-library/react-hooks"` while running against React 18 we would need to change that import to `import { renderHook } from "@testing-library/react"`.

MichaelBelousov:
> So when testing hooks against React 17 we need to use `import { renderHook } from "@testing-library/react-hooks"` while running against React 18 we would need to change that import to `import { renderHook } from "@testing-library/react"`.

popping in because I saw this in my email and the transformer also has [non-rush based tests for support across major versions of itwin.js](https://github.com/iTwin/imodel-transformer/blob/093be1ade668f683562baf302d2aae866de3032b/.github/workflows/ci.yml#L57), done by dynamically generating a matrix in our gh action.

Also, you can always fallback to a dynamic require to support multiple versions:

```ts
import * as Semver from "semver";
const reactVersion = require("react/package.json").version;
const renderHook = Semver.satisfies(reactVersion, "^18")
  ? require("@testing-library/react-hooks").renderHook
  : require("@testing-library/react").renderHook;
```

grigasp:
We won't be able to do this until all packages in the repo support React 18.

I see the following options:
1. Create a separate pipeline for packages that support React 18. Long term, only the unmaintained packages should stay in the old pipeline and all supported ones be in the new one. I believe the packages would still require the hack @MichaelBelousov showed, because release pipeline would still build against React 17.
2. Move tree and property widgets into `presentation` repo. This wouldn't solve the problem for other packages, but would make thinks a lot easier for my team.

aruniverse:
> 2\. Move tree and property widgets into `presentation` repo. This wouldn't solve the problem for other packages, but would make thinks a lot easier for my team.

I'm not strongly opposed to this, with your team now owning these pkgs it makes some sense. However we'd be losing the gold standard for pkgs in this monorepo.

@calebmshafer thoughts?

calebmshafer:
@aruniverse @grigasp I'd prefer if we didn't start shifting packages out of the repo. The more we shift them to other repos, the less crowdsourcing we'll have to ensure the single build system is working for everyone.

Given that I think the first option is better in the long run.

saskliutas:
CLosing in favor of https://github.com/iTwin/viewer-components-react/pull/724. Will open a separate PR for tree-widget.

PR #683, opened 2023-10-25
[ReportsConfig] Fix flaky tests

Guillar1:


packages/itwin/reports-config-widget/src/test/Reports.test.tsx
```diff
@@ -208,29 +156,16 @@ describe("Reports View", () => {
...
-      screen.getByTestId(/rcw-loading-delete/i)
-    );
     await waitForElementToBeRemoved(() => screen.getByRole("dialog"));
 
-    // Should be one less report
-    expect(screen.getAllByTestId("horizontal-tile")).toHaveLength(
-      mockedReportsOriginalLength - 1
-    );
+    expect(mockDeleteReport).toBeCalled();
+    expect(mockGetReports).toBeCalledTimes(2);
```
smmr-dn, line 162: Why do we expect to call `mockGetReports` twice?

packages/itwin/reports-config-widget/src/test/Reports.test.tsx
```diff
@@ -169,27 +134,10 @@ describe("Reports View", () => {
...
-          mockedReports.reports = mockedReports.reports.filter(
-            (report) => report.id !== mockedReports.reports[0].id
-          );
-          return res(ctx.delay(100), ctx.status(204));
-        }
-      )
-    );
 
-    const { user } = render(<Reports />);
+    mockGetReports.mockReturnValue(mockedReports.reports);
```
smmr-dn, line 138: Is expecting to call `mockGetReports` twice the reason why we use `mockReturnValue` here instead of `mockReturnValueOnce` in other functions?

packages/itwin/reports-config-widget/src/test/Reports.test.tsx
Guillar1: When it first loads, then after the deletion happens. It should refresh the list which is the second time. Left a comment.

packages/itwin/reports-config-widget/src/test/Reports.test.tsx
Guillar1: Yes.

smmr-dn (approved):

packages/itwin/reports-config-widget/src/test/Reports.test.tsx
```diff
@@ -43,38 +41,30 @@ const reportsFactory = (): ReportCollection => ({
...
-      )
-    );
-
+    mockGetReports.mockReturnValueOnce([]);
     const onClickAddMock = jest.fn();
 
-    const { user } = render(<Reports onClickAddReport={onClickAddMock} />);
+    const { user } = render(<Reports onClickAddReport={onClickAddMock} />, { reportsClient: mockReportsClient.object });
 
     await waitForElementToBeRemoved(() => screen.getByText(/loading/i));
```
MindaugasButkus, line 69: I think the most flakyness was caused by these `waitForElementToBeRemoved` by the element text.

PR #688, opened 2023-11-08
E2E test improvements

grigasp:
- Fix the script that starts test viewer for e2e tests to run on both - Windows and Unix.
- Get rid of the env. variable that tells to update screenshots or not. Instead, always update them when not in CI and never do when in CI.
- Provide some contributing guidelines. Include instructions for running e2e tests.


ben-polinsky:
>  Instead, **always update them when not in CI** and never do when in CI.

Does this make sense? If someone makes a bad change

grigasp:
> > Instead, **always update them when not in CI** and never do when in CI.
> 
> _(Sorry hand slipped and hit close/submit)_
> 
> Does this make sense? If someone makes a bad change, then it'll be committed and validated/accepted as passing in CI, right?

It will not be automatically committed. Running the tests will update the screenshots and you'll see them as changed files in the repository. As the new `CONTRIBUTING.md` files state, you should verify the changes, if any, are expected and valid before committing them.

ben-polinsky:
> > > Instead, **always update them when not in CI** and never do when in CI.
> > 
> > 
> > _(Sorry hand slipped and hit close/submit)_
> > Does this make sense? If someone makes a bad change, then it'll be committed and validated/accepted as passing in CI, right?
> 
> It will not be automatically committed. Running the tests will update the screenshots and you'll see them as changed files in the repository. As the new `CONTRIBUTING.md` files state, you should verify the changes, if any, are expected and valid before committing them.

Gotcha. That makes sense.

ben-polinsky (approved):

aruniverse (approved):

saskliutas (approved):

ViliusBa (approved):

JonasDov (approved):

Issue #697, opened 2023-11-17
Tree Widget: Enter 2.0 pre-release mode


grigasp:
```[tasklist]
### Tasks
- [x] Release currently pending changes.
- [x] Enter 2.0 major version pre-release mode for the package.
```

grigasp:
On hold until https://github.com/iTwin/viewer-components-react/pull/707 merges

saskliutas:
Entered with https://github.com/iTwin/viewer-components-react/pull/707

PR #701, opened 2023-11-29
Tree Widget: Enter prerelease mode for 2.0.0

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/697

Need to wait for https://github.com/iTwin/viewer-components-react/pull/700 to complete and release patch version.


saskliutas:
@aruniverse I ran `rush version --bump` locally to test it out and saw that `tree-widget` peer dep in `breakdown-trees` package is bumped which bumps `breakdown-tree` package version also. Do you know how to configure rush to avoid that?

aruniverse:
> @aruniverse I ran `rush version --bump` locally to test it out and saw that `tree-widget` peer dep in `breakdown-trees` package is bumped which bumps `breakdown-tree` package version also. Do you know how to configure rush to avoid that?

Does it bump breakdown-tree with a minor or a major?

I think to get around this, you would have to change the dep on tree-widget in breakdown-trees to not use `workspace` and set it to the actual version

saskliutas:
> > @aruniverse I ran `rush version --bump` locally to test it out and saw that `tree-widget` peer dep in `breakdown-trees` package is bumped which bumps `breakdown-tree` package version also. Do you know how to configure rush to avoid that?
> 
> Does it bump breakdown-tree with a minor or a major?
> 
> I think to get around this, you would have to change the dep on tree-widget in breakdown-trees to not use `workspace` and set it to the actual version

It does patch bump but I guess it's because major version is still at 0. I guess not using `workspace` in version is good solution for this I will make a change.

mdastous-bentley (approved):

aruniverse (approved):

grigasp (approved):

PR #702, opened 2023-12-01
[G&M] React Query Mappings

Guillar1:


packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
```diff
@@ -9,16 +9,16 @@ import {
...
 } from "@itwin/itwinui-react";
-import React, { useState } from "react";
-import { handleError } from "../../common/utils";
+import { useMutation } from "@tanstack/react-query";
+import React, { useCallback, useState } from "react";
 import "./DeleteModal.scss";
 import { LoadingSpinner } from "./LoadingSpinner";
 
 export interface DeleteModalProps {
   entityName?: string;
```
MindaugasButkus, line 18: Can the `entityName` ever be `undefined`?

packages/itwin/grouping-mapping-widget/package.json
```diff
@@ -94,13 +96,15 @@
     "@itwin/presentation-components": "^4.1.0",
     "@itwin/presentation-frontend": "^4.1.0",
     "@itwin/webgl-compatibility": "^4.1.0",
+    "@tanstack/query-core": "^4.36.1",
```
MindaugasButkus, line 99: This is used as a non-dev dependency.

Guillar1: Rush/pnpm refuses to install this transient dependency. So the builds fail.

MindaugasButkus: Then move it to non-dev dependency list.

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
Guillar1: Nope. Use to but not anymore.

packages/itwin/grouping-mapping-widget/src/components/GroupingMappingContext.tsx
```diff
@@ -62,6 +64,31 @@ export interface GroupingMappingContextProps {
 const authorizationClientGetAccessToken = async () =>
   (await IModelApp.authorizationClient?.getAccessToken()) ?? "";
 
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+      // staleTime: 20 * 60 * 1000, // 20 minutes
```
MindaugasButkus, line 71: Remove commented out code.

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
```diff
@@ -27,28 +27,27 @@ export const DeleteModal = ({
   onDelete,
   refresh,
 }: DeleteModalProps) => {
-  const [isDeleting, setIsDeleting] = useState<boolean>(false);
+  const [localEntityName] = useState(entityName);
```
MindaugasButkus, line 30: What's the purpose of this state? Can't you just use the string from props?

packages/itwin/grouping-mapping-widget/src/components/GroupingMappingContext.tsx
Guillar1: Actually this stays. Accidently pushed it. Fixed

MindaugasButkus: I don't understand. Why do we want to keep commented out configuration?

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
Guillar1: It's a copy of the original string. When the deletion happens, for a brief moment it becomes and empty string.

packages/itwin/grouping-mapping-widget/src/components/GroupingMappingContext.tsx
Guillar1: This is the cache time. I disabled the cache to debug. I accidently pushed it.

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
MindaugasButkus: Why does this become an empty string?

packages/itwin/grouping-mapping-widget/src/components/GroupingMappingContext.tsx
MindaugasButkus: In the reply you said `Fixed`. How?

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/DeleteModal.tsx
Guillar1: Because the original array holding the reference to this string gets updated and that reference disappears when the list updates and the entity is gone. The modal has a closing animation, so while it closes you can see the string disappear and become empty for a brief second.

packages/itwin/grouping-mapping-widget/src/components/GroupingMappingContext.tsx
Guillar1: I uncommented it.

packages/itwin/grouping-mapping-widget/package.json
Guillar1: This is a quirk with pnpm and workspaces. If you are consuming this package normally it should install on its own.

MindaugasButkus: Quirk or not, you are referencing it directly in code. Add it as a dependency.

Guillar1: Why, when react-query already has it as a dependency?
image

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+  iModelId,
+  getAccessToken,
+  extractionClient,
+}: {
+  iModelId: string;
+  getAccessToken: GetAccessTokenFn;
+  extractionClient: IExtractionClient;
+}) => {
+  return useQuery({
+    queryKey: ["extractionStatus", iModelId],
```
MindaugasButkus, line 21: The `extractionStatus` can conflict with a specific extraction, so I would recommend using a more specific name. Since this is tracking the latest iModel extraction, I would go with `iModelExtractionStatus`. Then if you would be tracking a specific extraction (not by iModelId, but by extractionId) you could use `extractionStatus` with the extractionId.

packages/itwin/grouping-mapping-widget/package.json
MindaugasButkus: So why do you even need to add the `@tanstack/query-core` into any kind of dependency list if "react-query already has it as a dependency"?

If you would be referencing `query-core` directly in tests I would say keep it this way. But now you are referencing it directly in production code. I don't get it why you should make it a dev dependency.

I think if you wan to keep it this way then for consistency you should move `classnames` (transitive dependency from `@itwin/itwinui-react`) as well.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+        const jobId = latestExtractionResult.value.jobId;
+        const status = await extractionClient.getExtractionStatus(accessToken, jobId);
+
+        if (status.containsIssues) {
+          const logs = await extractionClient.getExtractionLogs(accessToken, jobId);
+          extractionMessageData = logs.filter((log) => log.message !== null).map((log) => ({
+            date: log.dateTime,
+            category: log.category,
+            level: log.level,
+            message: String(log.message),
```
MindaugasButkus, line 47: Why is this `String` conversion required here?

packages/itwin/grouping-mapping-widget/package.json
Guillar1: I think I get you're point. Actually, I didn't realize that react-query was reexporting the packages. Sadly, I still need this as a dev dependency, since it is not hoisting it correctly.

MindaugasButkus: I am not saying that you don't need it. I am saying that you are using it as a dependency, not as a devDevependency.

Guillar1: Yup I get it. Nice catch. I was too focused on the pnpm issue at hand that I tunneled visioned.

Guillar1: Changed the import.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
Guillar1: Not needed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+        const jobId = latestExtractionResult.value.jobId;
+        const status = await extractionClient.getExtractionStatus(accessToken, jobId);
+
+        if (status.containsIssues) {
+          const logs = await extractionClient.getExtractionLogs(accessToken, jobId);
+          extractionMessageData = logs.filter((log) => log.message !== null).map((log) => ({
+            date: log.dateTime,
+            category: log.category,
+            level: log.level,
+            message: log.message ?? "",
```
MindaugasButkus, line 47: Can the `log.message` ever be nullish, even after the `filter(x => x.message !== null)` applied earlier?

Guillar1: The type coming from insights-client is defined as potentially undefined. Shouldn't it be `string | null`?

MindaugasButkus: Yes, it should `string | null`. I checked, and there are some entries (state-change) that return null message, but the `message` property is always there.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionMessageModal.tsx
```diff
@@ -24,58 +32,75 @@ export interface ExtractionMessageModalProps {
...
+    }, []);
+  }, [extractionMessageData]);
+
+  // useQueries to fetch all group names
+  const groupQueriesResults = useQueries({
+    queries: extractionInfo.map(({ mappingId, groupId }) => ({
+      queryKey: ["groups", mappingId, groupId],
+      queryFn: async () => {
+        const accessToken = await groupingMappingApiConfig.getAccessToken();
+        const groups = await mappingClient.getGroups(
```
MindaugasButkus, line 68: How many requests does this result in if there is more than 1 unique group in the extractionInfo array?

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatusIcon.tsx
```diff
@@ -6,8 +6,13 @@ import { IconButton, NotificationMarker, Tooltip } from "@itwin/itwinui-react";
...
 import { SvgProcess } from "@itwin/itwinui-icons-react";
 
-export interface ExtractionStatusIconProps {
+export interface ExtractionStatusData {
   iconStatus: "negative" | "positive" | "warning" | undefined;
+  iconMessage: string;
+}
+
+export interface ExtractionStatusIconProps {
+  iconStatus: ExtractionStatusData["iconStatus"];
```
MindaugasButkus, line 15: Maybe it would be better to create a separate type for the `iconStatus` instead of interface properties referencing one another.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/ConfirmMappingsImport.tsx
```diff
@@ -79,30 +78,39 @@ const ConfirmMappingImport = ({
     setSelectedMappings(newState);
   };
 
+  const importMutation = useMutation({
+    mutationFn: async (selectedMapping: IMappingTyped) => {
+      const accessToken = await getAccessToken();
+      await mappingClient.copyMapping(
+        accessToken,
+        sourceiModelId,
+        selectedMapping.id ?? "",
```
MindaugasButkus, line 87: Source mapping is required for copying a mapping. Providing an empty string will result in an error.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/ConfirmMappingsImport.tsx
```diff
@@ -79,30 +78,39 @@ const ConfirmMappingImport = ({
...
+        {
+          targetIModelId: iModelId ?? "",
```
MindaugasButkus, line 89: `targetIModelId` is required. Providing an empty string will result in an error.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectIModel.tsx
```diff
@@ -22,38 +23,26 @@ const defaultDisplayStrings = {
...
-    for await (const iModel of iModelIterator) {
-      iModels.push(iModel);
-    }
-    setIModels(iModels);
-  } catch (error: any) {
-    handleError(error.status);
-  } finally {
-    setIsLoading(false);
+  const accessToken = await getAccessToken();
+  const iModelIterator = iModelsClient.iModels.getRepresentationList({
```
MindaugasButkus, line 31: Does this really need the whole iModel representation? Maybe minimal would be enough?

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectIModel.tsx
```diff
@@ -22,38 +23,26 @@ const defaultDisplayStrings = {
...
+    authorization: async () => ({
```
MindaugasButkus, line 32: Use `AccessTokenAdapter`: https://github.com/iTwin/imodels-clients/blob/main/docs/IModelsClientManagement.md#authorization

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectITwin.tsx
```diff
@@ -31,32 +32,18 @@ const tabsWithIcons = [
...
-        iTwinsResponse = await iTwinsClient.queryAsync(accessToken, ITwinSubClass.Project);
-    }
-    setITwins(iTwinsResponse.data!);
-  } catch (error: any) {
-    handleError(error.status);
-  } finally {
-    setIsLoading(false);
+  const accessToken = await getAccessToken();
+  switch (iTwinType) {
+    case 0:
```
MindaugasButkus, line 41: What are these magic numbers?

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectITwin.tsx
```diff
@@ -76,43 +64,37 @@ const SelectITwin = ({
...
   const iTwinsClient = useITwinsClient();
-  const [isLoading, setIsLoading] = useState<boolean>(true);
-  const [iTwins, setITwins] = useState<ITwin[]>([]);
   const [iTwinType, setITwinType] = useState<number>(defaultITwinType);
 
-  useEffect(() => {
-    void fetchITwins(setITwins, setIsLoading, getAccessToken, iTwinsClient, iTwinType);
-  }, [getAccessToken, iTwinsClient, setIsLoading, iTwinType]);
+  const { data: iTwins, isFetching: isLoading } = useQuery({
+    queryKey: ["iTwin", iTwinType],
```
MindaugasButkus, line 70: `iTwinsByType` instead of `iTwin`. What if you need to fetch a single iTwin? Of course it will have different value (`iTwinType`) but still - I think it would be better to keep these buckets different.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
```diff
@@ -41,14 +41,15 @@ export const mappingViewDefaultDisplayStrings = {
 };
 
 export interface MappingsViewProps {
-  mappings: Mapping[];
+  mappings: Mapping[] | undefined;
```
MindaugasButkus, line 44: Does the requirement to support `undefined` come from the `react-query`? I don't think that it makes sense to render the `MappingsView` if you do not have the `Mappings`. The `mappings` property should remain mandatory.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
```diff
@@ -91,63 +17,54 @@ export interface MappingsOperationsProps extends GroupingMappingApiConfig {
...
+    },
+  });
 
-  const onDelete = async (mapping: Mapping) => {
-    const accessToken = await getAccessToken();
-    await mappingClient.deleteMapping(accessToken, iModelId, mapping.id);
-    await refresh();
-  };
+  const isLoading = isLoadingMappings || isLoadingExtractionStatus || isTogglingExtraction || isDeletingMapping;
+  const extractionStatusGated = extractionStatus ??  {extractionStatusIcon: {
```
MindaugasButkus, line 64: Looks like you have a double space there after `??`.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectIModel.tsx
Guillar1: It needs it. The minimal only provides the display name and id. This table also shows the description of the iModel.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
Guillar1: Mappings isn't immediately available, so I have to render something in its place while it loads. There can be a point where the mappings fetch can error out as well and what is returned is undefined.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
Guillar1: Changed to `iModelExtractionStatus`

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionMessageModal.tsx
Guillar1: I just rewrote it to just fetching groups. Added in a new hook from the work I've been doing with the groups components.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatusIcon.tsx
Guillar1: I agree. Done.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/ConfirmMappingsImport.tsx
Guillar1: Left over from incorrect types in the past. Fixed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/ConfirmMappingsImport.tsx
Guillar1: Left over from incorrect types in the past. Fixed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectIModel.tsx
Guillar1: Done

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectITwin.tsx
Guillar1: No more magic.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Import/SelectITwin.tsx
Guillar1: Good point. Renamed the key.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useMappingsOperations.ts
Guillar1: Fixed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStatus.ts
Guillar1: I opened a PR https://github.com/iTwin/insights-client/pull/46

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
MindaugasButkus: The component which requires the mappings defines what it needs. This component needs a mapping list, therefore if the mappings fetch errors out and returns undefined, it should be translated to an empty array there instead of this component handling undefined mappings list.

MappingsView should always receive a list to display. It can be empty, but it must be a list.

common/config/rush/pnpm-lock.yaml
```diff
@@ -509,7 +509,9 @@ importers:
       '@itwin/ecschema-metadata': ^4.1.0
       '@itwin/eslint-plugin': ^3.7.7
       '@itwin/imodel-components-react': ^4.1.0
-      '@itwin/imodels-client-management': ^3.1.0
+      '@itwin/imodels-access-common': ^4.1.2
```
MindaugasButkus, line 512: Any reason why this specific version is picked? Does 4.1.0 have what you need? Maybe even the 4.0.0 has what you need?

MindaugasButkus (approved):

common/config/rush/pnpm-lock.yaml
Guillar1: I don't think it should be an issue either way? This is just a local dependency. If this would be a peer dependency yes. Package managers can handle multiple versions.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
Guillar1: Sure, fixed. I was initially steering towards that but this convinced me again.

common/config/rush/pnpm-lock.yaml
MindaugasButkus: Yes, it's not an issue. I added this comment and thought that if users have 4.x.x in their project this will not cause any breaking changes.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
MindaugasButkus: "Fixed", but I don't see any changes.

Guillar1: git push connection refused 👎

MindaugasButkus (approved):

Guillar1:
@aruniverse can you review?

aruniverse (approved):
lockfile changes look fine

PR #708, opened 2023-12-08
escape escql queries

aruniverse:


grigasp:
@aruniverse what you changed is not an ECSQL query but rather an ECExpression. It does get converted to ECSQL internally, but keywords also get escaped during the process, so no need for this change.

PR #714, opened 2023-12-15
Bugfix/missing peerdeps

a-gagnon:
~~Add missing peerDependencies that are required at runtime by the map-layers package.
Besides `type` imports, anything that is used by the package should be a direct `dependency`, or be in both `devDependency` and `peerDependency`.~~
- Move some `devDependencies` to direct `dependencies`

Found the problem when adding the @itwin/map-layers package to my test-app. I didn't get a warning for missing peerDependency, but it complained about missing `@itwin/map-layers-formats` at build time.


packages/itwin/map-layers/package.json
```diff
@@ -115,6 +115,9 @@
     "@itwin/core-frontend": "^4.1.2",
     "@itwin/core-react": "^4.0.0",
     "@itwin/imodel-components-react": "^4.0.0",
+    "@itwin/itwinui-variables": "^2.0.0",
+    "@itwin/itwinui-react": "^2.11.2",
```
aruniverse, line 119: itwinui should not be a peer

a-gagnon: If this package consumes it as a direct dep, then it ought to be a peer to communicate to the package users that it's required.
i guess it is *assumed* this package lives in the larger itwinjs bubble so the dependency on itwinui stuff is implied?

aruniverse: itwinui should be a direct dep imo

aruniverse: itwinui should be direct;
need to review if map-layers-formats can be a direct dep. i think that might actually need to be a peer if it has statics and is not pure

a-gagnon:
> @a-gagnon This is old and no one approved...does it need additional update before we can complete it?

Haven't played with the app since then, but those should really be deps. I think it's fine to approve.

common/changes/@itwin/map-layers/bugfix-missing-peerdep_2023-12-15-14-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/map-layers",
+      "comment": "Some packages were pulled as devDependency when they really should be direct dependencies",
```
aruniverse, line 5:
```suggestion
      "comment": "Updated dependencies",
```

packages/itwin/map-layers/package.json
```diff
@@ -38,6 +38,9 @@
   "dependencies": {
     "@itwin/itwinui-icons-color-react": "^2.1.0",
     "@itwin/itwinui-icons-react": "^2.6.0",
+    "@itwin/itwinui-react": "^2.11.2",
+    "@itwin/itwinui-variables": "^2.0.0",
+    "@itwin/map-layers-formats": "^4.1.2",
```
aruniverse, line 43: why is this a direct dep? is this dep pure / has no side effects?
if its not, needs to be dev + peer dep

PR #717, opened 2023-12-18
Fix flaky e2e tests in PropertyGrid

### LINKED ISSUE #713: Property widget: Property list not rendered when selection changes rapidly
Reproduced by flaky e2e tests:
- `multiple elements selected`
- `multiple elements selected - search bar expanded`

Looks like the problem is that the second node is selected while properties for the first one are not loaded yet, which causes the properties for final selection to not load either.
```[tasklist]
### Tasks
- [ ] Header should show label of last selected item (https://github.com/iTwin/viewer-components-react/pull/717)
- [ ] Spinner is not visible when property grid is rendered in widget
- [ ] https://github.com/iTwin/appui/issues/641
```

ViliusBa:
Closes #713 

Fixed the issue where titles for in property widget would load for such items for which the `getData()` call completed faster, rather than displaying the values for which the last call was made. Added a test case which simulates this situation.

Edited e2e tests to additionally wait for the property widget to load before clicking other item in the tree widget.


ben-polinsky (approved):

grigasp:
> It seems that ensuring that properties are loaded for the first selected item helped.

Could you reproduce this without the change? I don't agree that waiting for the load to complete before clicking on the second node is the solution - users may not know they need to wait :)

common/changes/@itwin/property-grid-react/property-grid-fix-flaky-tests_2023-12-18-17-19.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/property-grid-react",
+      "comment": "",
```
grigasp, line 5: Looks like a legit patch - need a change message.

packages/itwin/property-grid/src/e2e-tests/PropertyGrid.test.ts
```diff
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
- * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
- * See LICENSE.md in the project root for license terms and full copyright notice.
- *--------------------------------------------------------------------------------------------*/
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
```
grigasp, line 3: should changes in this file be reverted?

packages/itwin/property-grid/src/hooks/UseInstanceInfo.ts
```diff
@@ -22,18 +23,25 @@ export interface UseLoadedInstanceInfoProps {
...
+        setItem({ label: propertyData.label, className: propertyData.description ?? "" });
+      }
     };
 
     const removeListener = dataProvider.onDataChanged.addListener(onDataChanged);
     void onDataChanged();
 
     return () => { removeListener(); };
-  }, [dataProvider]);
+  }, [dataProvider, dataProvider.onDataChanged]);
```
saskliutas, line 44:
```suggestion
  }, [dataProvider]);
```

saskliutas (changes requested):

saskliutas (approved):

PR #720, opened 2023-12-21
Bump AppUI version and unskip related e2e tests

### LINKED ISSUE #710: Property widget: Wrong property list layout after selecting element from list
**Repro:**
1. Make the property widget narrow enough for property label / value to be layout in a column rather than a row.
2. Select multiple elements in the tree widget.
3. Notice the properties layout is fine.
4. Enter elements list, select one of the elements.
5. Notice the property label / value are now layout in a row rather than a column.

This is also _often_ reproduced by the following e2e tests:
- `single element selected from elements list`
- `single element selected from elements list - search bar expanded`

ViliusBa:
Closes #710


aruniverse (approved):

packages/itwin/property-grid/package.json
```diff
@@ -68,10 +68,10 @@
   },
   "devDependencies": {
     "@itwin/appui-abstract": "^4.0.0",
-    "@itwin/appui-layout-react": "^4.5.0",
-    "@itwin/appui-react": "^4.5.0",
+    "@itwin/appui-layout-react": "^4.8.0",
+    "@itwin/appui-react": "^4.8.0",
     "@itwin/build-tools": "^4.0.0",
-    "@itwin/components-react": "^4.5.0",
+    "@itwin/components-react": "^4.8.0",
```
saskliutas, line 74: You need to bump appui version in test-viewer to see appui fix in e2e tests

saskliutas (changes requested):

grigasp:
Tree widget tests fail due to an issue in appui - waiting for [the fix](https://github.com/iTwin/appui/pull/668) to be released.

saskliutas (approved):

PR #727, opened 2024-01-10
Make e2e containers more stable

grigasp:
This PR allows running e2e tests in docker containers multiple times without having to rebuild the image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,0 @@
-**/node_modules/
-lib/
-playwright-report/
-test-results/
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -9,5 +9,8 @@ ARG PACKAGE_NAME=""
 WORKDIR /workspaces/viewer-components-react/packages/itwin/${PACKAGE_NAME}
 RUN npx playwright install chromium
 
+# Copy test snapshots into container before running the tests
+COPY /packages/itwin/${PACKAGE_NAME}/src/e2e-tests src/e2e-tests
+
 # Set the entry point to run the tests
 CMD ["npm", "run", "test:e2e:local"]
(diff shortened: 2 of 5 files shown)


grigasp:
A few E2E tests were failing due to https://github.com/iTwin/imodel-native/issues/586 - skipped until the issue is fixed on the backend.

saskliutas (approved):

change/@itwin-property-grid-react-944e621e-779d-495c-8b49-f5a79733ce64.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Skip tests that are failing due to a backend issue",
```
aruniverse, line 3:
```suggestion
  "comment": "",
```

aruniverse (approved):

change/@itwin-property-grid-react-944e621e-779d-495c-8b49-f5a79733ce64.json
grigasp: Out of curiosity, why do you see this as a problem? The `type: "none"` will make sure it's not going to appear in the changelog.

aruniverse: oh i assumed it was like rush, where even if we specify none now, when we do publish itd show up in the changelogs eventually. if thats not the case, please ignore

grigasp: I confirmed two things for changes with type "none":
- When only "none" changes exist for specific package, it's version doesn't get bumped.
- They don't get placed into `CHANGELOG.md`, only in `CHANGELOG.json`, which is fine, I think.

PR #733, opened 2024-01-12
[G&M] Trigger extraction in mapping UI

smmr-dn:
Added features to trigger extraction and show live extraction status on mapping UI
- Added a button to run extraction.
- Allowed users to multiselect/deselect mappings.
- Added status icons in each mapping list item to show live extraction status.

final_list_extraction


packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatus.tsx
```diff
@@ -0,0 +1,71 @@
...
+        animation={fadeOut}
+        onAnimationEnd={onAnimationEnd}
+      />);
+    case ExtractionStates.Failed:
+      return (<FailedExtractionState
+        animation={fadeOut}
+        onAnimationEnd={onAnimationEnd}
+      />);
+    default:
+      return <></>;
```
Guillar1, line 69: I think you can return null here.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  mapping: Mapping;
+  jobStartEvent: BeEvent<(mappingId: string) => void>;
+  onClickMappingTitle?: (mapping: Mapping) => void;
+  onClickMappingModify?: (mapping: Mapping) => void;
+  onRefreshMappings: () => Promise<void>;
+  onToggleExtraction: (mapping: Mapping) => Promise<void>;
+  setShowDeleteModal: (mapping?: Mapping) => void;
+}
+
+export const MappingHorizontalTile = (props: MappingHorizontalTileProps) => {
```
Guillar1, line 29: `MappingList` or `MappingHorizontalTile`? File name does not match.

smmr-dn: I have changed the component name.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatus.tsx
smmr-dn: I have changed this to null.

packages/itwin/grouping-mapping-widget/package.json
```diff
@@ -61,9 +61,10 @@
     "@itwin/imodels-access-common": "^4.1.2",
     "@itwin/imodels-access-frontend": "^4.1.2",
     "@itwin/imodels-client-management": "^4.2.1",
-    "@itwin/insights-client": "^0.4.0",
+    "@itwin/insights-client": "^0.6.0",
```
MindaugasButkus, line 64: I think this means that this whole PR can no longer be a `patch`. Pre-release (>1.0.0) minor version changes can have breaking changes, therefore this package upgrade might be a breaking change for some users.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStates/SucceededExtractionState.tsx
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import React from "react";
+import { SvgStatusSuccess } from "@itwin/itwinui-icons-color-react";
+import { ANIMATION_DELAY, ANIMATION_DURATION } from "../../../Constants";
+
+interface ExtractionStateProps {
```
MindaugasButkus, line 9: Both Failed and Succeeded extraction states use the same interface, but it is duplicated. The only difference between these components is `title` and the `SvgStatus*` node.

MindaugasButkus: What you could do is keep the `Failed` and `Succeeded` extraction state components, but move the common things to a new component `TerminalExtractionState` that would be used by `Failed` and `Succeeded` extraction states, passing in the `title` and `SvgStatus*` as props.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatus.tsx
```diff
@@ -0,0 +1,71 @@
...
+interface ExtractionStatusProps {
+  state: ExtractionStates | undefined;
+  clearExtractionState: () => void;
+}
+
+export const ExtractionStatus = ({
+  state,
+  clearExtractionState,
+}: ExtractionStatusProps) => {
+
```
MindaugasButkus, line 33: Don't add empty lines at the beginning or the end of a block of code.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.scss
```diff
@@ -0,0 +1,21 @@
+.gmw-list-item-container{
```
MindaugasButkus, line 1: Missing copyright header.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, [extractionStateData, isJobDone, isJobStarted]);
+
+  return (
+    <ListItem actionable
+      className="gmw-list-item-container"
+      active={props.selected}
+      key={props.mapping.id}
+      onClick={onClickTile}
+      title={props.mapping.mappingName}>
+      <ListItem.Content>
```
MindaugasButkus, line 65: The contents of `ListItem.Content` are hard to read as they currently are. Could you split the inner components into multiple lines? Something similar like you have for `ListItem` - the component starts at one line and each of the properties get their own line. Then the ternary operator could also be split into multiple lines.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
```diff
@@ -60,6 +61,7 @@ export interface MappingsViewProps {
   onClickAddMapping?: () => void;
   onClickMappingTitle?: (mapping: Mapping) => void;
   onClickMappingModify?: (mapping: Mapping) => void;
+  onRunExtraction: (mappings: Mapping[]) => Promise<void>;
```
MindaugasButkus, line 64: Adding a mandatory prop to a component that is exported from the package is a breaking change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  mapping: Mapping;
+  jobStartEvent: BeEvent<(mappingId: string) => void>;
+  onClickMappingTitle?: (mapping: Mapping) => void;
+  onClickMappingModify?: (mapping: Mapping) => void;
+  onRefreshMappings: () => Promise<void>;
+  onToggleExtraction: (mapping: Mapping) => Promise<void>;
+  setShowDeleteModal: (mapping?: Mapping) => void;
+}
+
+export const MappingList = (props: MappingListProps) => {
```
MindaugasButkus, line 29: This is a misnomer. This is not a _list_, but rather a single a single _list item_.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { ExtractionRequestMapping, ExtractionRunRequest, Mapping } from "@itwin/insights-client";
+import { useState } from "react";
+import { ExtractionStates } from "../Extraction/ExtractionStatus";
+import { STATUS_CHECK_INTERVAL } from "../../Constants";
+
+export const useFetchExtractionStates = ({
+  iModelId,
+  getAccessToken,
+}: GroupingMappingApiConfig) => {
+
```
MindaugasButkus, line 18: Do not add empty lines at the beginning or the end of a block of code.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
```diff
@@ -0,0 +1,76 @@
...
+  const { data: extractionStateData, isError: isExtractionStateError, isLoading: isExtractionStatusLoading } = useQuery({
+    queryKey: ["extractionState"],
+    staleTime: Infinity,
+    placeholderData: undefined,
+    queryFn: async () => {
+      const accessToken = await getAccessToken();
+      const extractionHistoryIterator = extractionClient.getExtractionHistoryIterator(accessToken, iModelId, 1);
+      const extractionHistoryResult = await extractionHistoryIterator.next();
+      const jobId = extractionHistoryResult.value.jobId;
```
MindaugasButkus, line 43: This query will fail if used on an iModel with no extractions.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
```diff
@@ -0,0 +1,76 @@
...
+    },
+  });
+
+  const { data: extractionStateData, isError: isExtractionStateError, isLoading: isExtractionStatusLoading } = useQuery({
+    queryKey: ["extractionState"],
+    staleTime: Infinity,
+    placeholderData: undefined,
+    queryFn: async () => {
+      const accessToken = await getAccessToken();
+      const extractionHistoryIterator = extractionClient.getExtractionHistoryIterator(accessToken, iModelId, 1);
```
MindaugasButkus, line 41: Querying extraction history on every state check if very inefficient. The state checks themselves cause a lot of network traffic, since it is polling for state every couple of seconds. Adding the extraction history query will double the amount of network traffic.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
```diff
@@ -0,0 +1,76 @@
...
+    },
+    enabled: isJobStarted,
+    refetchInterval: STATUS_CHECK_INTERVAL,
+  });
+
+  const finalExtractionState = async (state: ExtractorState | undefined) => {
+    if (state === undefined)
+      return ExtractionStates.None;
+
+    if (state === ExtractorState.Failed || isExtractionStateError)
```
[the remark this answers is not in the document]
smmr-dn: It's supposed to be Failed. I will change this into a switch statement.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
smmr-dn: I have worked around and found a way to cache the jobId accordingly to the selected mapping(s) every time users run extraction so when it comes to querying status, it refers to that cached jobId without going to the route of retrieving history.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingsView.tsx
smmr-dn: I have moved this inside `MappingView`.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStates/SucceededExtractionState.tsx
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/Extraction/ExtractionStatus.tsx
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.scss
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.tsx
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingList.tsx
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
smmr-dn: I have made this change.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
```diff
@@ -0,0 +1,87 @@
...
+    case ExtractorState.Queued:
+      return { mappingId, finalExtractionStateValue: ExtractionStates.Queued};
+    case ExtractorState.Succeeded:
+      return { mappingId, finalExtractionStateValue: ExtractionStates.Succeeded};
+    default:
+      return { mappingId, finalExtractionStateValue: ExtractionStates.None};
+  }
+};
+
+export const createQueryForEachJobId = (mappingId: string, jobId: string, getAccessToken: GetAccessTokenFn, extractionClient: IExtractionClient, enabled: boolean) => ({
```
MindaugasButkus, line 57: How can it create something "for each", when only 1 thing is given?

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
MindaugasButkus: What if user refreshes the web page? Then all of the state is lost. So we still need to have the extraction history query somewhere. Completely removing it is not the right way.

smmr-dn: Personally, I think having jobID mapped with each mapping then using `getExtractionStatus` to retrieve status for that specific mapping will be sufficient since users can start extraction on mapping(s) at different time. Extraction history will potential return the latest job(s) but it might cause running mappings to reset their status as soon as another job is started. That is also why I stored a hash map in context with keys are mapping IDs and values are their assigned jobIDs.

smmr-dn: But I think it's a good idea to think of a way to keep/store extraction status such as passing down a list of query results when `Mapping` mounts so that we won't lose status when users refresh the whole page.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchExtractionStates.ts
smmr-dn: I have made it to check whether `jobId` exists, if not just simply returns the "None" status. I'm not sure if this works effectively.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
smmr-dn: I have changed the method's name.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
```diff
@@ -0,0 +1,102 @@
...
+import { useMemo } from "react";
+import { useExtractionStateJobContext } from "../../context/ExtractionStateJobContext";
+
+export interface MappingExtractionStatusProps extends GroupingMappingApiConfig {
+  isMounted: boolean;
+  mapping: Mapping;
+  enabled: boolean;
+}
+
+export interface MappingQueryResults {
```
Guillar1, line 23: I think it would look neater if you defined it as `MappingQueryResult` and then used `MappingQueryResult[]` where needed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingListItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+      setIsJobStarted(false);
+      setMappingIdJobInfo((prevMap: Map<string, string>) => {
+        const newMap = new Map(prevMap);
+        newMap.delete(props.mapping.id);
+        return newMap;
+      });
+      await resetMappingExtractionStatus(props.mapping.id, queryClient);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [statusQuery, props.mapping.id]);
```
Guillar1, line 74: Why not pull all required dependencies in here? This callback might become stale.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingListItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }, [isMounted]);
+
+  // Check whether the job is still running when users refresh the mapping list
+  // or modify any mappings
+  useEffect(() => {
+    if(mappingIdJobInfo.get(props.mapping.id)){
+      setIsJobStarted(true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Guillar1, line 60: Is this still needed?

smmr-dn: It needed `props.mapping.id` as well, I just updated the file.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingListItem.tsx
smmr-dn: I included all the dependencies.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
smmr-dn: I have changed it to `MappingQueryResult`.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingListItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  useEffect(() => {
+    // Only apply to all mappings when the page is reloaded, not the list component
+    // or when users modify mappings
+    if(props.isMappingPageRefreshed && props.setIsMappingPageRefreshed){
+      setIsListItemMounted(true);
+      setIsJobStarted(true);
+      props.setIsMappingPageRefreshed(false);
+    }
+  }, [props]);
```
Guillar1, line 55: Is this on purpose or does it need to be this broad? Couldn't it just be `[props.isMappingPageRefreshed, props.setIsMappingPageRefreshed]`

smmr-dn: For some reason, eslint errors on my end. Should I just include the exhaustive deps comment?
image

Guillar1: Hmm. Just get rid of props. Deconstruct it in the params.

smmr-dn: I have updated the params.

packages/itwin/grouping-mapping-widget/src/components/Mappings/MappingListItem.tsx
```diff
@@ -27,66 +27,79 @@ export interface MappingListItemProps {
...
       setIsJobStarted(true);
-      props.setIsMappingPageRefreshed(false);
+      setIsMappingPageReloaded(false);
     }
-  }, [props]);
+  }, [isMappingPageReloaded, setIsMappingPageReloaded]);
 
   // Check whether the job is still running when users refresh the mapping list
-  // or modify any mappings
+  // or modify any mappingsA
```
Guillar1, line 71: `mappings`

smmr-dn: oops, i changed it back

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
```diff
@@ -66,12 +62,12 @@ export const useFetchMappingExtractionStatus = ({
 }: MappingExtractionStatusProps) => {
   const extractionClient = useExtractionClient();
   const { mappingIdJobInfo } = useExtractionStateJobContext();
-  const jobId = mappingIdJobInfo.get(mapping.id);
+  const jobId = mappingIdJobInfo.get(mapping.id) ?? "";
```
MindaugasButkus, line 65: `jobId` should never be an empty string. It should either be null / undefined or a valid value (guid).

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useRunExtraction.ts
```diff
@@ -0,0 +1,50 @@
...
+  const extractionClient = useExtractionClient();
+  const [isJobStarted, setIsJobStarted] = useState<boolean>(false);
+  const { mappingIdJobInfo, setMappingIdJobInfo } = useExtractionStateJobContext();
+
+  const { mutateAsync: runExtraction, isLoading: isRunExtractionLoading, isSuccess: isRunExtractionSuccess } = useMutation({
+    mutationKey: ["runExtraction"],
+    mutationFn: async (mappings: Mapping[]) => {
+      const accessToken = await getAccessToken();
+      const mappingIds: ExtractionRequestMapping[] = mappings.length > 0 ? mappings.map((mapping) => { return { id: mapping.id }; }) : [];
+      const extractionRequest: ExtractionRunRequest | undefined = {
```
MindaugasButkus, line 29: What's the point of having `| undefined` if it is `const` and initialized with a non-undefined value?

MindaugasButkus (approved):

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useRunExtraction.ts
Guillar1: Removed.

packages/itwin/grouping-mapping-widget/src/components/Mappings/hooks/useFetchMappingExtractionStatus.ts
Guillar1: Made it more clear in the code.

PR #740, opened 2024-01-18
Disable breakdown-trees publishing

saskliutas:
Disabled `breakdown-trees` publishing. Beachball tries to bump `tree-widget` peer dep and publish `breakdown-trees` to npm and fails. Chose to disable `breakdown-trees` publishing for a moment to avoid blocking publishing of other packages.

Fixed changelog generation to avoid git command failures when commit hash is not available.

### DIFF
--- a/beachball.config.js
+++ b/beachball.config.js
@@ -9,17 +9,17 @@ module.exports = {
   bumpDeps: false,
   access: "public",
   tag: "latest",
-  scope: ["packages/itwin/*", "!packages/itwin/tree-widget"],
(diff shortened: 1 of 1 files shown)


aruniverse (approved):

aruniverse:
thinking about this more, im ok just taking this pkg out of the `pnpm-workspace.yaml` file so we dont build it.
imo this pkg is dead without an owner so it shouldnt force you guys to keep it up to date and compiling and what not

PR #741, opened 2024-01-18
ModelsTree: move onNodeDoubleClick to ModelsTreeEventHandler

ViliusBa:
This refactor was needed because having the current implementation in `VisibilityTreeEventHandler` would only work with `ModelsTree`, so creating a new `ModelsTreeEventHandler` which would hold logic that is specific for `ModelsTree` is more correct.

### DIFF
--- a/packages/itwin/tree-widget/src/components/trees/common/UseVisibilityTreeState.tsx
+++ b/packages/itwin/tree-widget/src/components/trees/common/UseVisibilityTreeState.tsx
@@ -10,7 +10,7 @@ import { VisibilityTreeEventHandler } from "../VisibilityTreeEventHandler";
 import type { IModelConnection } from "@itwin/core-frontend";
 import type { Ruleset } from "@itwin/presentation-common";
 import type { IFilteredPresentationTreeDataProvider, PresentationTreeEventHandlerProps, UsePresentationTreeStateProps } from "@itwin/presentation-components";
-import type { IVisibilityHandler, VisibilityTreeSelectionPredicate } from "../VisibilityTreeEventHandler";
+import type { IVisibilityHandler, VisibilityTreeEventHandlerParams, VisibilityTreeSelectionPredicate } from "../VisibilityTreeEventHandler";
 import type { VisibilityTreeFilterInfo } from "./Types";
 
 /**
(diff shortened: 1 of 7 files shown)


packages/itwin/tree-widget/src/components/trees/common/UseVisibilityTreeState.tsx
```diff
@@ -53,13 +56,21 @@ export function useVisibilityTreeState({
...
+            nodeLoader: params.nodeLoader,
+            visibilityHandler,
+            selectionPredicate,
+          })) ??
+        new VisibilityTreeEventHandler({
+          nodeLoader: params.nodeLoader,
+          visibilityHandler,
+          selectionPredicate,
+        })
+      );
```
grigasp, line 71:
```suggestion
      const eventHandlerProps: VisibilityTreeEventHandlerParams = {
        nodeLoader: params.nodeLoader,
        visibilityHandler,
        selectionPredicate,
      };
      return eventHandler 
        ? eventHandler(eventHandlerProps) 
        : new VisibilityTreeEventHandler(eventHandlerProps);
```

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTree.tsx
```diff
@@ -205,6 +206,7 @@ function useTreeState({
           : selectionPredicateRef.current(node.key, ModelsVisibilityHandler.getNodeType(node)),
       [],
     ),
+    eventHandler: useCallback((props: VisibilityTreeEventHandlerParams) => new ModelsTreeEventHandler(props), []),
```
grigasp, line 209: Maybe better to define this factory outside of render function? Wouldn't need `useCallback` then.

grigasp (changes requested):

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTree.tsx
ViliusBa: Not having it wrapped causes infinite re-renders. `selectionPredicate` parameter is passed in the same way. Also, it makes sense to wrap the event handler in a` useCallback` with no dependencies, since this passed `eventHandler` factory is specific for models tree and is not subject to change.

ViliusBa: Grigas clarified his suggestion in private, applied it.

grigasp (approved):

saskliutas (approved):

PR #745, opened 2024-01-26
[G&M] Localized Presentation categories

Guillar1:


packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyAction.tsx
```diff
@@ -279,6 +280,19 @@ export const GroupPropertyAction = ({
     }
   }, [searchInput, setSearched, clearSearch]);
 
+  const getLocalizedStringPresentation = useCallback((input: string): string => {
```
MindaugasButkus, line 283: It looks like this function could be used anywhere where presentation is used. Maybe move it to some kind of utils?

Guillar1: This is the only place this can be used as of now, because this is rendering the descriptor data directly. But I do like the idea of moving it somewhere else. I moved it.

packages/itwin/grouping-mapping-widget/src/common/utils.ts
```diff
@@ -95,3 +96,16 @@ export const enableExperimentalFeatures = async (iModelConnection: IModelConnect
...
 };
+
+export const getLocalizedStringPresentation = (input: string): string => {
+  // https://www.itwinjs.org/presentation/advanced/localization/
+  // Check if the string starts and ends with "@"
+  if (input.startsWith("@") && input.endsWith("@")) {
+    // Remove the leading and trailing "@" from the string
+    const trimmedString = input.substring(1, input.length - 1);
+
+    return IModelApp.localization.getLocalizedString(trimmedString);
```
MindaugasButkus, line 107: I wonder if this adds a requirement to initialize the localization before rendering the Grouping & Mapping widget. I remember that Reports config widget has an initialization step because of localization. Is it the same here?

Guillar1: Yes, the viewer itself already does the initialization. 
https://github.com/iTwin/viewer/blob/a58000781f538ac468368e875072d587a415186d/packages/modules/viewer-react/src/services/BaseInitializer.ts#L196

MindaugasButkus (approved):

grigasp:
Don't you think it would be worth reporting such an issue so it can be properly fixed rather than worked around like this? 🤔

PR #746, opened 2024-01-26
Make sure the Feature Info Widget is reopened every time selection change occurs

mdastous-bentley:


apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -66,29 +66,28 @@ function useViewCreatorOptions() {
...
-          }
-        }, 100);
-      });
-    };
-
-    tileTreesLoaded().finally(() => {
-      void IModelApp.tools.run(FitViewTool.toolId, viewPort, true, false);
-      viewPort.view.setStandardRotation(StandardViewId.Iso);
-    });
+    // const tileTreesLoaded = () => {
```
eringram, line 69: Is this intentionally commented out and could it be deleted if it's unused?

mdastous-bentley: Actually this is causing rejected promise error in the application because it times out.
@aruniverse Do we have any better way to achieve this?

aruniverse: i would just remove this, its not really needed. just added as an example in the starter templates

mdastous-bentley: @aruniverse @eringram ok removed `useViewCreatorOptions`

eringram (approved):

apps/test-viewer/src/components/Viewer.tsx
Guillar1: I think the Viewer itself loads this view config if nothing is passed in. It has been a while since I've checked but I think it still holds true. https://github.com/iTwin/viewer/blob/a58000781f538ac468368e875072d587a415186d/packages/modules/viewer-react/src/services/iModel/ViewCreator3d.ts#L71

When I had to deal with this, I would just pass in an empty object into `viewportConfigurer`.

mdastous-bentley:
@aruniverse I think your approval is needed.

aruniverse (approved):

PR #756, opened 2024-02-01
Hide User Preferences Options to store at the iTwin/iModel scope.

mdastous-bentley:
The following options are now hidden by default, and the layer sources are always on the iTwin scope:
image

A new `showUserPreferencesStorageOptions` option has been added to restored if neeed.


aruniverse (commented):
id want someone from a product chime in on the terminology here. "settings" is kinda misleading imo

@jason-crow @jjbeckman13

mdastous-bentley:
> id want someone from a product chime in on the terminology here. "settings" is kinda misleading imo
> 
> @jason-crow @jjbeckman13

Yea I agree, it's been like that since the beginning.  Anyway everyone is confused about those options, hence the reason to hide it.  The whole concept of user preferences is kind of distorted  TBH.

eringram (approved):

grigasp:
> @grigasp Looks like it's breaking e2e tests, can you help with that?

https://github.com/iTwin/viewer-components-react/pull/779 should fix this.

PR #757, opened 2024-02-05
Maplayer Feature Info Tech Preview badge option

wilmaier:
Add support to show tech preview badge on new maplayer feature info tool/widget


mdastous-bentley (approved):

mdastous-bentley:
@wilmaier thinking about it... why have an option for the badge?  I think we can simply add the badge and remove it when its no longer tech preview.

wilmaier:
Ok I'll go ahead and make that change then.

PR #760, opened 2024-02-07
Improved warning message when an incompatible map-layer type is provided.

mdastous-bentley:


eringram (approved):

grigasp:
> @grigasp e2e error here too.

https://github.com/iTwin/viewer-components-react/pull/779 should fix this.

PR #761, opened 2024-02-08
Tree Widget: allow opt-in hierarachy level size limiting

ViliusBa:


packages/itwin/tree-widget/src/components/trees/category-tree/CategoriesTree.tsx
```diff
@@ -54,10 +54,10 @@ export interface CategoryTreeProps extends BaseFilterableTreeProps {
    */
   viewManager?: ViewManager;
   /**
-   * Flag that determines if hierarchy level filtering will be enabled for this tree.
+   * Props for configuring hierarchy level filtering.
    * @beta
    */
-  isHierarchyLevelFilteringEnabled?: boolean;
+  hierarchyLevelFilteringProps?: HierarchyLevelFilteringProps;
```
grigasp, line 60: Need a better name for this prop. While it's recommended to have filtering enabled when limiting, the two aren't really related to each other. So putting limiting properties under "filtering" prop is not correct, IMO.

ViliusBa: Do you suggest not putting them under the same interface and have them as two separate props?

packages/itwin/tree-widget/src/components/trees/imodel-content-tree/IModelContentTree.tsx
```diff
@@ -63,7 +63,7 @@ export const IModelContentTree = (props: IModelContentTreeProps) => {
         state={state}
         selectionMode={selectionMode ?? SelectionMode.None}
         treeRenderer={(treeProps) =>
-          hierarchyLevelFilteringProps ? (
+          hierarchyLevelConfig ? (
```
grigasp, line 66: I feel like it would be more correct to check `hierarchyLevelConfig?.isFilteringEnabled` here

saskliutas (approved):

PR #763, opened 2024-02-08
Tests fix imodel react hooks

ben-bartholomew:
Some of the tests for useFeatureOverrides in the package imodel-react-hooks were skipped during a previous upgrade. These changes fix and re-enable them.


aruniverse (approved):

MichaelBelousov (approved):
LGTM except I hesitate to let this through without adding a test that confirms there are no longer any feature overrides in the provider list after unmounting...

ben-bartholomew:
Fixing exists tests that were commented out is one thing, but I don't think I have the bandwidth or the expertise to add a new test that does something like that.

FWIW I gave it a try, but its an internal variable that disappears when the component unmounts; something clever will have to be done to get access to it without exposing it to consumers of the package.

PR #767, opened 2024-02-12
Added `@itwin/itwinui-react` to peer dependencies

mdastous-bentley:


eringram (approved):

aruniverse:
itwinui should not be a peer, it should be a direct dep

PR #778, opened 2024-02-21
[G&M] Merge duplicate properties

Guillar1:


packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyUtils.ts
```diff
@@ -221,42 +231,52 @@ export const convertPresentationFields = (propertyFields: Field[]): PropertyMeta
...
+              });
             }
           }
         }
       }
     }
-    return [];
   });
 
+  return Array.from(uniquePropertiesMap.values());
```
MindaugasButkus, line 277: Storing values in a map and then retrieving listing all of them might change the order, since Maps are not ordered (I think). Is there anything that ensures a consistent ordering of these properties?

Guillar1: Maps remember insertion order. [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map](url)

MindaugasButkus (approved):

PR #782, opened 2024-02-27
Touch friendly Tree Widget

jasdom:
Increased the size of Tree Widget header, its buttons and node action buttons when using expanded layout.


grigasp (changes requested):
Please add e2e tests to capture screenshots of the changed components.

saskliutas (changes requested):

apps/test-viewer/src/components/ViewerOptions.tsx
```diff
@@ -0,0 +1,52 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { StatusBarItem, StatusBarSection, UiItemsProvider } from "@itwin/appui-react";
+import { SvgVisibilityShow } from "@itwin/itwinui-icons-react";
+import { createContext, PropsWithChildren, useContext, useState } from "react";
+
+export interface ViewerOptionsContext {
```
saskliutas, line 10: Maybe we can split this context into `ViewerOptionsContext` and `ViewerActionsContext`? I think that would be a little bit cleaner for components that only used options values and does not modify them.

apps/test-viewer/src/components/ViewerOptions.tsx
```diff
@@ -0,0 +1,52 @@
...
+      {
+        id: `${this.id}:expandedLayoutButton`,
+        tooltip: "Toggle expanded layout",
+        icon: (<SvgVisibilityShow />),
+        label: "expandedLayoutToggle",
+        section: StatusBarSection.Left,
+        itemPriority: 1,
+        execute: () => {
+          this.isEnlarged = !this.isEnlarged;
+          this.setDensity(this.isEnlarged ? "enlarged" : "default");
```
saskliutas, line 47: No need to track `isEnlarged`. You can use state update action. `this,setDensity((prev) => prev === "default" ? "enlarged" : "default");`

grigasp (commented):
Some observations from enlarged e2e test screenshots:
- Tree selector is not enlarged.
- Node icon doesn't look vertically aligned with label - it's a bit below the centerline.
- In some screenshots node labels don't fit, but there's no horizontal scrollbar.

apps/test-viewer/src/UiProvidersConfig.tsx
```diff
@@ -71,28 +72,47 @@ const configuredUiItems = new Map<string, UiItem>([
             {
               id: ModelsTreeComponent.id,
               getLabel: ModelsTreeComponent.getLabel,
-              render: () => (
-                <ModelsTreeComponent
+              render: () => {
```
saskliutas, line 75: Does changing arrow function to something like `render: function ModelsTreeComponentRenderer() {` resolves eslint error?

packages/itwin/tree-widget/src/components/tree-header/TreeHeader.scss
```diff
@@ -66,4 +70,38 @@
...
+      .search-open, .search-input, .search-close {
+        height: var(--enlarged-header-height);
+        min-width: var(--enlarged-header-height);
+      }
+    }
+
+    .button-container {
+      width: calc(100% - var(--enlarged-header-height));
+
+      .header-button {
```
saskliutas, line 97: Can we adjust header to accommodate button component from `itwinui-react` with default size?

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeComponent.test.tsx
```diff
@@ -245,6 +271,14 @@ describe("<ModelsTreeComponent />", () => {
     });
 
     describe("<View3DButton />", () => {
+      it("renders enlarged View3DButton", async () => {
+        const { getByRole } = render(
+          <ModelsTreeComponent.View3DButton models={[{ id: "modelTestId", isPlanProjection: true }]} viewport={mockViewport().object} density="enlarged" />,
+        );
+        const button = await waitFor(() => getByRole("button"));
+        expect(button.getAttribute("data-iui-size")).to.be.null;
```
saskliutas, line 279: Don't thin we can count on `data-iui-size` allways being there. It looks like internal implementation detail of itwinui-react. 

I think enlarged layout is better tested with e2e so for getting coverage up I think we can render `enlarged` button in one of the other tests.

saskliutas (changes requested):

apps/test-viewer/src/UiProvidersConfig.tsx
jasdom: Yes, using your suggestion did resolve the eslint error, however after refactoring it was no longer necessary.

packages/itwin/tree-widget/src/components/tree-header/TreeHeader.scss
jasdom: Switched to using larger buttons, so we would no longer need to additionally increase the size.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeComponent.test.tsx
jasdom: Removed the tests and added `enlarged` buttons to other tests.

apps/test-viewer/src/UiProvidersConfig.tsx
```diff
@@ -77,41 +78,42 @@ const configuredUiItems = new Map<string, UiItem>([
         await TreeWidget.initialize();
       },
       createUiItemsProviders: () => [
-        new TreeWidgetUiItemsProvider({
+        new TreeWidgetWithOptionsUiItemsProvider({
```
saskliutas, line 81: I think this can be simplified using new UIProvider API:
```tsx
createUIItemsProviders: () => [{
  id: "TreeWidgetUIProvider,
  getWidgets: () => {
    const trees = [ // tree definitions ]
    return [{
      id: "tree-widget",
      content: <TreeWidgetWithOptions trees={trees} />,
      layouts: {
        standard: {
           section: StagePanelSection.Start,
           location: StagePanelLocation.Right,
        }
      }
    }]
}]
```

Then you can move `TreeWidgetWithOptions` to this file and remove `TreeWidgetWithOptionsUiItemsProvider.tsx`

packages/itwin/tree-widget/src/components/SelectableTree.tsx
```diff
@@ -25,7 +25,7 @@ export interface TreeDefinition {
   /** Callback that is used to get tree label */
   getLabel: () => string;
   /** Callback that is used to render tree component */
-  render: () => React.ReactNode;
+  render: (density?: "enlarged" | "default") => React.ReactNode;
```
saskliutas, line 28: I would suggest defining props type: 
```typescript
render: (props: TreeRenderProps) => React.ReactNode;
```

packages/itwin/tree-widget/src/components/SelectableTree.tsx
```diff
@@ -61,13 +62,13 @@ function SelectableTreeContent(props: SelectableTreeProps & { imodel: IModelConn
 
   return (
     <div className="tree-widget-selectable-tree">
-      <SelectableContent {...getSelectableContentProps(trees)} />
+      <TreeSelector {...getTreeSelectorProps(trees, props.density)} />
```
saskliutas, line 65: No need to pass `density` to `getTreeSelectorProps`. You can do `<TreeSelector {...getTreeSelectorProps(trees)} density={props.density} />`

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export interface TreeContentDefinition {
+  id: string;
+  label: string;
+  render: (density?: "enlarged" | "default") => React.ReactNode;
+}
+
+export interface TreeSelectorProps {
+  defaultSelectedContentId: string;
+  children: TreeContentDefinition[];
```
saskliutas, line 19: I suggest renaming this to `definitions` or `trees` because it might be confusing as `children` prop name is used for child components.

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <div className="presentation-components-tree-selector-content" title={className}>
+      <div className="presentation-components-tree-selector-content-header">
+        {options.length > 0 && (
+          <Select
+            options={options}
+            value={selectedContent.id}
+            onChange={setSelectedContentId}
+            aria-label={props.selectAriaLabel}
+            disabled={props.disabled}
+            menuClassName={className}
```
saskliutas, line 44: Maybe we can use custom item renderer instead of customizing menu css?

packages/itwin/tree-widget/src/components/tree-header/TreeHeader.scss
```diff
@@ -66,4 +70,33 @@
...
+
+    .tree-widget-search-box {
+      height: var(--enlarged-header-height);
+      width: calc(100% - var(--enlarged-header-height));
+
+      &.contracted {
+        width: var(--enlarged-header-height);
+      }
+
+      .search-open, .search-input, .search-close {
```
saskliutas, line 88: Maybe these components can use `size="large"` instead of custom styling?

packages/itwin/tree-widget/src/test/tree-header/TreeHeader.test.tsx
```diff
@@ -35,6 +35,40 @@ describe("<TreeHeader />", () => {
     sinon.restore();
   });
 
+  it("renders non `enlarged` header", async () => {
```
saskliutas, line 38: I think same as for buttons this should be tested with e2e test instead of checking `data-iui-size`

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,64 @@
...
+                size={isEnlarged ? "large" : "default"}
+                isSelected={itemProps.isSelected}
+                onClick={() => {
+                  setSelectedContentId(option.value);
+                  itemProps.close();
+                }}
+                role='option'
+              >
+                {option.label}
+              </MenuItem>
```
saskliutas, line 56: I think you can simplify this:
```suggestion
              <MenuItem
                {...itemProps}
                size={isEnlarged ? "large" : "default"}
              >
                {option.label}
              </MenuItem>
```

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,64 @@
...
+  id: string;
+  label: string;
+  render: (props: TreeRenderProps) => React.ReactNode;
+}
+
+export interface TreeSelectorProps {
+  defaultSelectedContentId: string;
+  trees: TreeContentDefinition[];
+  selectAriaLabel?: string;
+  disabled?: boolean;
```
saskliutas, line 22: These are not used in `SelectableTree`. Maybe remove them?

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import "./TreeSelector.scss";
+import type { SelectOption } from "@itwin/itwinui-react";
+import { useMemo, useState } from "react";
+import { MenuItem, Select } from "@itwin/itwinui-react";
+import type { TreeRenderProps } from "./SelectableTree";
+
+export interface TreeContentDefinition {
```
saskliutas, line 12: Add `@internal` tags to exported types and functions.

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
```diff
@@ -30,36 +41,22 @@ test.describe("should match image snapshot", () => {
     const node = locateNode(treeWidget, "ProcessPhysicalModel");
     await node.getByTestId("tree-node-expansion-toggle").click();
 
-    // wait for node at the bottom to be visible/loaded
-    await locateNode(treeWidget, "Tag-Category").waitFor();
+    await new Promise((f) => setTimeout(f, 3000));
```
saskliutas, line 44: Why `locateNode` was changed to this timeout?

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
```diff
@@ -123,4 +119,33 @@ test.describe("should match image snapshot", () => {
...
+  });
+
+  test("tree selector", async ({ page }) => {
+    await treeWidget.getByRole("combobox").click();
+    await page.getByRole("listbox").waitFor();
+    await takeScreenshot(page, treeWidget);
+  });
+};
+
+test.describe("should match image snapshot", () => {
```
saskliutas, line 137: Maybe use similar layout to property grid e2e tests?
```typescript
test.describe("tree widget" () => {
  const testCases = () => {
    // ...
  }

  test.describe("default", () => {
    testCases();
  });

  test.describe("enlarged", () => {
    test.beforeEach(() => {
      // toggle expanded layout
    });
    testCases();
  });
})
```

saskliutas (changes requested):

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
jasdom: I believe the last node never gets loaded when using the expanded layout as it is too far away and needs to be scrolled to.

saskliutas: I think we could pass last node label to `testCases` function instead of setting 3s timeout. Can you go over all timeout you added and replace them with check for something to appear in UI?

apps/test-viewer/src/components/ViewerOptions.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export function useViewerOptionsContext() {
+  return useContext(viewerOptionsContext);
+}
+
+export function useViewerActionsContext() {
+  return useContext(viewerActionsContext);
+}
+
+export class ViewerOptionsUiItemsProvider implements UiItemsProvider {
```
saskliutas, line 39: Played around with test-viewer and it was refreshing whole app on each layout toggle. Creating custom button component allows to avoid refreshing whole app:
```tsx
export const statusBarActionsProvider: UiItemsProvider = {
  id: "ViewerOptionsUiItemsProvider",
  getStatusBarItems: () => [
    {
      id: `expandedLayoutButton`,
      tooltip: "Toggle expanded layout",
      content: <ToggleLayoutButton />,
      itemPriority: 1,
      execute: () => {},
      section: StatusBarSection.Left,
    },
  ],
};

function ToggleLayoutButton() {
  const { setDensity } = useViewerActionsContext();
  return (
    <IconButton styleType="borderless" onClick={() => setDensity((prev) => (prev === "default" ? "enlarged" : "default"))}>
      <SvgVisibilityShow />
    </IconButton>
  );
}
```

packages/itwin/tree-widget/src/components/TreeSelector.tsx
jasdom: Simplified.

packages/itwin/tree-widget/src/components/TreeSelector.tsx
jasdom: Removed.

packages/itwin/tree-widget/src/components/TreeSelector.tsx
jasdom: Added tags.

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
jasdom: Passed the node label to `testCases` and reverted to wait for the node to appear.

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
jasdom: Changed.

apps/test-viewer/src/components/ViewerOptions.tsx
jasdom: Applied the suggestion, however it was rendering two buttons so made some additional changes.

packages/itwin/tree-widget/src/test/TreeSelector.test.tsx
```diff
@@ -0,0 +1,75 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import { fireEvent, render } from "@testing-library/react";
```
saskliutas, line 7: Use `render` from `./TestUtils`. Additionally to rendering component it returns `user` for making interactions.

packages/itwin/tree-widget/src/components/TreeSelector.tsx
```diff
@@ -0,0 +1,69 @@
...
+                {...itemProps}
+                size={isEnlarged ? "large" : "default"}
+              >
+                {option.label}
+              </MenuItem>
+            )}
+          />
+        )}
+      </div>
+      <div className="presentation-components-tree-selector-content-wrapper">{selectedContent?.render(props)}</div>
```
saskliutas, line 66: Passing `props` will leak `defaultSelectedContentId` and `trees` to tree renderer.
```suggestion
      <div className="presentation-components-tree-selector-content-wrapper">{selectedContent?.render({ density: props.density })}</div>
```

saskliutas (changes requested):

grigasp (approved):

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
```diff
@@ -19,108 +20,137 @@ test.beforeEach(async ({ page, baseURL }) => {
...
+      await treeWidget.getByTitle("More").click();
+      await page.locator(".tree-header-button-dropdown-container").waitFor();
+      await takeScreenshot(page, treeWidget);
+    });
+
+    test("tree selector", async ({ page }) => {
+      await treeWidget.getByRole("combobox").click();
+      await page.getByRole("listbox").waitFor();
+      await takeScreenshot(page, treeWidget);
+    });
```
saskliutas, line 140: I ran e2e tests locally and it looks like these tests might be flaky depending on how quickly roots nodes are loaded. I would suggest adding `waitFor` to check if root node is visible before doing interactions and taking screenshot.

saskliutas (changes requested):

saskliutas (approved):

PR #783, opened 2024-02-28
[G&M] Fixed an issue where hidden groups were not hidden again

Guillar1:
When visualizing groups, hidden groups were not hidden again when returning to Groups View and when a group was recreated with the same query.
Happened because of slight race condition and incorrectly caching groups alongside their hiliteid results. Decoupled the group and query association from the caching and processed that after the cached result.


packages/itwin/grouping-mapping-widget/src/components/Groups/GroupsVisualization.tsx
```diff
@@ -105,24 +105,44 @@ export const GroupsVisualization = ({
...
-        : [],
-    [groupQueries, isGroupsQueriesReady]
-  );
+  const hiliteIds = useMemo(() => {
+    if (!isGroupsQueriesReady || !groups) return [];
+
+    const processedGroupIds = new Map<string, string[]>();
+
+    return groupQueries.flatMap((query) => {
+      // Find all groups that match the current query and haven't been processed yet
```
MindaugasButkus, line 114: What does `processed` mean here?

MindaugasButkus:
Could you provide reproduction steps for the issue that you are solving? It's unclear from just the title of PR what you are fixing.

packages/itwin/grouping-mapping-widget/src/components/Groups/GroupsVisualization.tsx
Guillar1: I wrote a comment above. I guess associated would be a better term. Processed means associated with hiliteid results. This whole logic exists to avoid issues when their are groups with the same query.

MindaugasButkus (approved):

PR #785, opened 2024-02-28
Added @itwin\map-layers-formats to peerDependencies

mdastous-bentley:


mdastous-bentley:
@aruniverse As requested.

aruniverse (changes requested):
lockfile changes?

mdastous-bentley:
@aruniverse I looked into it and coulndnt find anything.  Maybe it didn't get updated because it was already a devDep ? @ben-polinsky

aruniverse (approved):
This is technically breaking but I don't see how this worked without this before so

eringram (approved):

PR #804, opened 2024-04-04
Tree widget & Property grid: Remove border around search button

saskliutas:
`@itwin/itwinui-react@3.6.0` version brought some changes to buttons used in Searchbox which led to border appearing around search button:
image

Removed border around search button.
Adjusted search button positioning in the header.


grigasp (approved):

packages/itwin/tree-widget/src/components/tree-header/TreeHeader.scss
```diff
@@ -82,7 +76,7 @@
       width: calc(100% - var(--enlarged-header-height));
 
       &.contracted {
-        width: var(--enlarged-header-height);
+        width: unset;
```
grigasp, line 79: am I right to think that our e2e tests don't capture the effect of this? shouldn't they?

saskliutas: Borderless button with matches `var(--enlarged-header-height)` so there are no difference. Changed to `unset` to match styling of default density.

PR #809, opened 2024-04-10
[G&M] List components to iTwinUI

Guillar1:


packages/itwin/grouping-mapping-widget/src/components/Groups/GroupListItem.tsx
```diff
@@ -24,45 +25,44 @@ export interface GroupItemProps extends Omit<GroupsProps, "onClickAddGroup"> {
   ) => void;
 }
 
-export const GroupItem = ({
+export const GroupListItem = ({
```
MindaugasButkus, line 28: Is there an easy way to check what is exported from the package? I see you are renaming components, so I am not that confident that this will not cause breaking changes.

MindaugasButkus: @Guillar1 Could you set it up in a separate PR?

MindaugasButkus (approved):

julioalvahash (approved):

Issue #797, opened 2024-03-12
Models Tree - Request performance improvement during visibility on/off with multi selection


muthubentley:
[Request from User]
Select multiple elements and turn visibility on/off – Elements visibility is turned on/off one by one with a little bit of delay. If possible please turn them on/off at once.
This can be reproduced in design review as well . Refer to the video in the below path:
https://bentley.sharepoint.com/:v:/r/sites/Fukui-BentleyCollaboration/Shared%20Documents/iTwin%20-%20CIMPHONY%20Plus/SampleData/TreePerformance/ModelTreeMultiSelection.mp4?csf=1&web=1&e=kxT1qn

grigasp:
iModel: https://connect-imodelweb.bentley.com/imodeljs/?projectId=19c7abf7-77a6-4151-9ee9-f2fb9e2dfaab&iModelId=1395f52b-aba1-41e2-b8e1-8b9911c35807&ChangeSetId=0920ad2c63ad146b40740a64dadef427f2794c66

grigasp:
I believe this has something to do with https://github.com/iTwin/viewer-components-react/pull/359, but needs further investigation.

grigasp:
@muthubentley, [@itwin/tree-widget-react@2.1.0](https://www.npmjs.com/package/@itwin/tree-widget-react/v/2.1.0) with the fix is now released

muthubentley:
Thank you

PR #810, opened 2024-04-11
Tree Widget: Improve visibility change performance

saskliutas:
Fixes https://github.com/iTwin/viewer-components-react/issues/797

Improved visibility change performance: 
* Reduced amount of content request that are made to get element node children ids. Request is done only if node has children.
* Change how `neverDrawn`/`awaysDrawn` list are updated to support changing multiple element visibility at the same time instead of handling one at the time.
* Update tree events handler to reflect checkbox state immediately after visibility was changed instead of waiting for all nodes to handled.


packages/itwin/tree-widget/src/test/trees/models-tree/ModelsVisibilityHandler.test.ts
```diff
@@ -1683,6 +1672,32 @@ describe("ModelsVisibilityHandler", () => {
...
+        const vpMock = mockViewport({ viewState: viewStateMock.object });
+
+        vpMock.setup((x) => x.isAlwaysDrawnExclusive).returns(() => false);
+        vpMock.setup((x) => x.alwaysDrawn).returns(() => new Set());
+        vpMock.setup((x) => x.neverDrawn).returns(() => new Set()); 
+        vpMock.setup((x) => x.setNeverDrawn(moq.It.is((set) => set.size === 1 && set.has(key.id)))).verifiable(moq.Times.once());
+
+        await using(createHandler({ viewport: vpMock.object }), async (handler) => {
+          // note: need to override to avoid running a query on the imodel
+          const spy = sinon.spy(handler as any, "getAssemblyElementIds");
```
grigasp, line 1694: I don't think this note is necessary

packages/itwin/tree-widget/src/components/trees/VisibilityTreeEventHandler.ts
```diff
@@ -139,24 +139,24 @@ export class VisibilityTreeEventHandler extends UnifiedSelectionTreeEventHandler
     };
 
     new Observable<CheckboxStateChange[]>((subscriber) => event.stateChanges.subscribe(subscriber))
-      .pipe(
-        mergeMap((changes) => this.changeVisibility(changes)),
-      )
+      .pipe(mergeMap((changes) => this.changeVisibility(changes)))
       .subscribe({
+        next: (nodeId) => {
+          void this.updateCheckboxes([nodeId]);
```
grigasp, line 145: Do we need to call `updateCheckBoxes()` from `handleStateChanged` if we do that for each node individually?

saskliutas: This call updates only checkbox of the node specified by `nodeId`. But I believe other nodes might be impacted also. So calling `updateCheckboxes` on complete makes sure that we update all checkboxes based on viewport state after all changes.

grigasp (approved):

PR #812, opened 2024-04-11
Add extract-api step for the PR CI build

Yato333:
Currently, some packages uses api-extractor tool, which helps to check for breaking API changes.
Added a build step that will fail if `pnpm extract-api` produces a result that differs from the one that is committed.

### DIFF
--- a/.pipelines/templates/build-steps.yaml
+++ b/.pipelines/templates/build-steps.yaml
@@ -20,5 +20,5 @@ steps:
   # - script: pnpm exec prettier . --check
   #   displayName: Check formatting
 
-  - script: pnpm exec lage lint cover extract-api --grouped
+  - script: pnpm exec lage lint extract-api cover --grouped
     displayName: Build, Lint, Cover and check for unexpected API changes
--- a/.pipelines/templates/build-steps.yaml
+++ b/.pipelines/templates/build-steps.yaml
@@ -20,11 +20,8 @@ steps:
   # - script: pnpm exec prettier . --check
   #   displayName: Check formatting
 
-  - script: pnpm exec lage lint cover --grouped
-    displayName: Build, Lint and Cover
-
-  - script: pnpm exec lage extract-api --grouped
-    displayName: Extract API
+  - script: pnpm exec lage lint cover extract-api --grouped
+    displayName: Build, Lint, Cover and check for unexpected API changes
     env:
-      # Fail the build if there are uncommitted API changes
+      # Fail the build if package uses api-extractor and there are uncommitted API changes
       TF_BUILD: 1
--- a/.pipelines/templates/build-steps.yaml
+++ b/.pipelines/templates/build-steps.yaml
@@ -22,6 +22,3 @@ steps:
 
   - script: pnpm exec lage lint cover extract-api --grouped
     displayName: Build, Lint, Cover and check for unexpected API changes
-    env:
-      # Fail the build if package uses api-extractor and there are uncommitted API changes
-      TF_BUILD: 1
--- a/.pipelines/templates/build-steps.yaml
+++ b/.pipelines/templates/build-steps.yaml
@@ -22,3 +22,9 @@ steps:
 
   - script: pnpm exec lage lint cover --grouped
     displayName: Build, Lint and Cover
+
+  - script: pnpm exec lage extract-api --grouped
+    displayName: Extract API
+    env:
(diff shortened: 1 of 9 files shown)


packages/itwin/tree-widget/api/tree-widget-react.api.md
```diff
@@ -451,7 +453,7 @@ export interface TreeContextMenuProps {
 export interface TreeDefinition {
     getLabel: () => string;
     id: string;
-    render: () => React.ReactNode;
+    render: (props: TreeRenderProps) => React.ReactNode;
```
grigasp, line 456: @saskliutas @jasdom, looks like we missed a breaking change. Can props be made optional?

saskliutas: I don't think this is a breaking change. These props are always provided by `TreeSelector` when rendering selected tree. https://github.com/iTwin/viewer-components-react/blob/9d296a06dd2e2290364692e02b109e9ac591a5e9/packages/itwin/tree-widget/src/components/TreeSelector.tsx#L64

This callback should never be called by consumer. And consumers can keep passing render callback that does not need props:
```ts
 {
  id: CategoriesTreeComponent.id,
  getLabel: CategoriesTreeComponent.getLabel,
  render: () => <CategoriesTreeComponent />,
},
```

grigasp: Agreed.

.pipelines/templates/build-steps.yaml
```diff
@@ -20,5 +20,8 @@ steps:
   # - script: pnpm exec prettier . --check
   #   displayName: Check formatting
 
-  - script: pnpm exec lage lint cover --grouped
-    displayName: Build, Lint and Cover
+  - script: pnpm exec lage lint cover extract-api --grouped
+    displayName: Build, Lint, Cover and check for unexpected API changes
+    env:
+      # Fail the build if package uses api-extractor and there are uncommitted API changes
+      TF_BUILD: 1
```
saskliutas, line 27: Is this needed? It should be predefined (https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables-devops-services)

saskliutas:
I think CI fails because `extract-api` task is missing dependency on `build` https://github.com/iTwin/viewer-components-react/blob/9d296a06dd2e2290364692e02b109e9ac591a5e9/lage.config.js#L34

.pipelines/templates/build-steps.yaml
Yato333: I am not sure, I copied it from presentation repo

saskliutas: Presentation repo uses Github actions instead of Azure DevOps. So it should not be necessary in here,

Yato333: Let me check if this is the case

Yato333: Yes it works, thanks

saskliutas (approved):

change/@itwin-property-grid-react-6141cefa-e559-437e-85ff-3904f0fd207e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add extract api step",
```
aruniverse, line 3:
```suggestion
  "comment": "",
```

change/@itwin-tree-widget-react-1de32109-0cd0-442d-8e17-ccf77b483c50.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Add extract api step",
```
aruniverse, line 3:
```suggestion
  "comment": "",
```

aruniverse (approved):

change/@itwin-property-grid-react-6141cefa-e559-437e-85ff-3904f0fd207e.json
saskliutas: FYI, there is no need to make comment empty string. If type is set to `none` it won't impact `CHANGELOG.md` generation.

Issue #801, opened 2024-03-26
Tree widget: Performance telemetry reporting


grigasp:
Should add a way for widget consumers to optionally supply a function of type `(featureId: string, duration: number) => void`. Consumers can then call their telemetry client with this information to track features' performance.

The features I think we should initially track:
- `{tree}-initial-load` - from the moment we create a new tree model till initial tree nodes are rendered.
- `{tree}-hierarchy-level-load` - from the moment user clicks on the "expand" button till child nodes are rendered.

... where `{tree}` is the tree identifier.

Add a README section that explains the API, its purpose and reported features.

PR #813, opened 2024-04-12
Tree widget performance telemetry

jasdom:
Closes [801](https://github.com/iTwin/viewer-components-react/issues/801).


packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
```diff
@@ -0,0 +1,44 @@
+import { useCallback, useEffect, useRef } from "react";
```
grigasp, line 1: missing file header

jasdom: Added the header.

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * Enables performance reporting for a tree component.
+ * @internal
+ */
+export function usePerformanceReporting(props: UsePerformanceReportingProps): UsePerformanceReportingResult {
+  const { treeIdentifier, iModel, ruleset, onPerformanceMeasured } = props;
+  const firstLoadRef = useRef(true);
+
+  useEffect(() => {
+    firstLoadRef.current = true;
+  }, [iModel, ruleset]);
```
saskliutas, line 34: If I understand correctly dependency on `ruleset` and this `firstLoadRef` is needed for the special case we have in `ModelsTree` where we are using different node loader/ruleset for unfiltered/filtered tree. As this is a special case and from consumer perspective it's a single tree I think `ModelsTree` should report node load time only from unfiltered node loader.

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const onNodeLoaded = useCallback(
```
saskliutas, line 36: There is no need to wrap `onNodeLoaded` in a `useCallback`.

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
jasdom: Removed the `ruleset` dependency and disabled performance reporting in `ModelsTree` when filtered state is used.

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
jasdom: Right, removed.

packages/itwin/tree-widget/src/test/trees/common/UsePerformanceReporting.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    it("does not log multiple initial loads", async () => {
+      const { result } = renderHook((props) => usePerformanceReporting(props), { initialProps });
+      expect(result.current).to.not.be.undefined;
+
+      result.current.onNodeLoaded?.({ node: "root", duration: 20 });
+      result.current.onNodeLoaded?.({ node: "root", duration: 10 });
+
+      expect(onPerformanceMeasuredSpy.callCount).to.be.eq(2);
```
saskliutas, line 57: Does not look like condition matches test name. From the test name I would expect that calling `onNodeLoaded` with `root` node twice results in single call of `onPerformanceMeasuredSpy`

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function usePerformanceReporting(props: UsePerformanceReportingProps): UsePerformanceReportingResult {
+  const { treeIdentifier, iModel, onPerformanceMeasured } = props;
+  const firstLoadRef = useRef(true);
+
+  useEffect(() => {
+    firstLoadRef.current = true;
+  }, [iModel]);
+
+  const onNodeLoaded = ({ node, duration }: { node: string; duration: number }) => {
+    if (firstLoadRef.current && node === "root") {
```
saskliutas, line 35: I don't think we should log root nodes load as `-hierarchy-level-load`. In my opinion loading root nodes is `initial-load`

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
saskliutas, line 1: There is no JSX in this file so it's extension could be changed to `.ts`

saskliutas (changes requested):

change/@itwin-tree-widget-react-ca23f996-6160-4e15-9918-2444a229add8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Added ability to track performance of `TreeWidget` component features.",
```
grigasp, line 3: I think this should mention the requirement for `@itwin/presentation-components` to be at `5.1` for the reporting to work.

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.tsx
jasdom: Switched to always reporting root nodes as `initial-load`.

packages/itwin/tree-widget/src/test/trees/common/UsePerformanceReporting.test.tsx
jasdom: Removed the test since it was no longer relevant.

change/@itwin-tree-widget-react-ca23f996-6160-4e15-9918-2444a229add8.json
jasdom: Mentioned the requirement.

packages/itwin/tree-widget/src/components/trees/common/UsePerformanceReporting.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const onNodeLoaded = ({ node, duration }: { node: string; duration: number }) => {
+    if (node === "root") {
+      onPerformanceMeasured?.(`${treeIdentifier}-initial-load`, duration);
+      return;
+    }
+    onPerformanceMeasured?.(`${treeIdentifier}-hierarchy-level-load`, duration);
+  };
+
+  return { onNodeLoaded: onPerformanceMeasured ? onNodeLoaded : undefined };
```
saskliutas, line 32:
```suggestion
  if (!onPerformanceMeasured) {
    return { onNodeLoaded: undefined };
  }

  return {
    onNodeLoaded: ({ node, duration }: { node: string; duration: number }) => {
      const feature = `${treeIdentifier}-${node === "root" ? "initial-load" : "hierarchy-level-load"}`;
      onPerformanceMeasured(feature, duration);
    },
  };
```

saskliutas (approved):

PR #814, opened 2024-04-12
[G&M] isMounted fix

Guillar1:


julioalvahash:
What are we fixing in this PR ?, I want to make sure I'm looking for it on the UI.

Guillar1:
> What are we fixing in this PR ?, I want to make sure I'm looking for it on the UI.

`useIsMounted` is a hook checks whether the current component is currently mounted (running). When hiliting (for example `Color Group` in the properties menu) sometimes the query can generate a huge result. As of now, there is no way to cancel a long running query request, but we can handle what happens afterwards. We use `useIsMounted` to check whether the component is still mounted after the query request is finished and finish out the hiliting. If someone decides to cancel this and back out of the component the hiliting portion should not happen after the query request completes.

julioalvahash (approved):

PR #816, opened 2024-04-17
fix bad publish

aruniverse:


packages/itwin/tree-widget/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/tree-widget-react",
-  "version": "2.0.2",
+  "version": "2.2.0",
```
saskliutas, line 3: The version is incorrect. It should be 2.1.0. Did you run `beachball sync` and then `beachball bump`? It should have been enough to run `beachball bump` . Other package also have incorrect version

saskliutas (changes requested):

packages/itwin/tree-widget/package.json
aruniverse: thx for catching that, updated

saskliutas: Looks good. I think you can by pass checks and merge.

saskliutas (approved):

PR #818, opened 2024-04-23
[EC3] Replaced redundant types

Guillar1:


packages/itwin/ec3-widget/src/components/LabelAction.tsx
```diff
@@ -30,21 +31,21 @@ export const LabelAction = ({ template, onClose, label, setTemplate }: LabelActi
   const [name, setName] = useState<string>(label?.name ?? "");
   const [itemName, setItemName] = useState<string>(label?.elementNameColumn ?? "UserLabel");
   const [itemQuantity, setItemQuantity] = useState<string>(label?.elementQuantityColumn ?? "");
-  const [selectedMaterial, setSelectedMaterial] = useState<Material | undefined>();
+  const [selectedMaterial, setSelectedMaterial] = useState<EC3ConfigurationMaterial | undefined>();
   const [isLoading, setIsLoading] = useState<boolean>(true);
-  const [materials, setMaterials] = useState<Material[]>(label?.materials.map((x) => { return { nameColumn: x.nameColumn }; }) ?? [{ nameColumn: undefined }]);
+  const [materials, setMaterials] = useState<(EC3ConfigurationMaterial | undefined)[]>(label?.materials ?? [undefined]);
```
MindaugasButkus, line 36: What's the point of having `undefined` in the array? Why not an empty array in that case?

packages/itwin/ec3-widget/src/components/LabelAction.tsx
```diff
@@ -106,15 +102,15 @@ export const LabelAction = ({ template, onClose, label, setTemplate }: LabelActi
       setName(label.name);
       setItemName(label.elementNameColumn);
       setItemQuantity(label.elementQuantityColumn);
-      setMaterials(label.materials.map((x) => { return { nameColumn: x.nameColumn }; })); // creating a copy of array, so original (in the parent) isn't modified
+      setMaterials(label.materials.map((x) => x)); // creating a copy of array, so original (in the parent) isn't modified
```
MindaugasButkus, line 105: I think `[...label.materials]` would result in the same thing.

packages/itwin/ec3-widget/src/components/LabelAction.tsx
Guillar1: I did that originally, but the materials selection is written in such an odd way that it would require extensive refactoring. That would have to be in another PR.

packages/itwin/ec3-widget/src/components/LabelAction.tsx
Guillar1: Good point. Done.

MindaugasButkus (approved):

Issue #791, opened 2024-03-06
Property widget: Features telemetry reporting


grigasp:
Should add a way for widget consumers to optionally supply a function of type `(featureId: string) => void`. Consumers can then call their telemetry client with this information to find out which widget features are used.

The features I think we should initially track:
- `single-element` - showing properties for a single element.
- `multiple-elements` - showing merged properties for multiple elements.
- `elements-list` - showing elements list in case of multi-selection.
- `single-element-from-list` - showing properties for a single element after selecting it in elements list.
- `ancestor-navigation` - traversing the elements' hierarchy using ancestor navigation buttons (up / down).
- `context-menu` - context menu for a property is opened.
- `hide-empty-values-enabled` - property values loaded with "hide empty values" setting enabled.
- `hide-empty-values-disabled` - property values loaded with "hide empty values" setting disabled.
- `filter-properties` - properties are filtered; changing selection with an already filtered property grid should also count as feature usage.

Add a README section that explains the API, its purpose and reported features.

PR #819, opened 2024-04-24
Property widget feature telemetry

jasdom:
Closes [791](https://github.com/iTwin/viewer-components-react/issues/791).

### DIFF
--- a/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
+++ b/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
@@ -23,7 +23,6 @@ import type {
   VirtualizedPropertyGridWithDataProviderProps,
 } from "@itwin/components-react";
 import type { IDisposable } from "@itwin/core-bentley";
-import { useLatest } from "../hooks/UseLatest";
 
 /**
  * Properties for rendering a `FilteringPropertyGrid`.
@@ -34,25 +33,22 @@ export interface FilteringPropertyGridProps extends VirtualizedPropertyGridWithD
   autoExpandChildCategories?: boolean;
   /** @internal */
   filterer: IPropertyDataFilterer;
-  /** Callback called when property data is loaded. */
-  onDataLoaded?: () => void;
 }
 
 /**
  * Creates a filtered data provider before rendering a `VirtualizedPropertyGridWithDataProvider`.
  * @internal
  */
-export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, onDataLoaded, ...props }: FilteringPropertyGridProps) {
+export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, ...props }: FilteringPropertyGridProps) {
   const [filteringDataProvider, setFilteringDataProvider] = useState<AutoExpandingPropertyFilterDataProvider>();
-  const onDataLoadedRef = useLatest(onDataLoaded);
   useEffect(() => {
     const filteringProvider = new FilteringPropertyDataProvider(dataProvider, filterer);
-    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories, onDataLoadedRef.current);
+    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories);
     setFilteringDataProvider(provider);
     return () => {
       provider.dispose();
     };
-  }, [filterer, dataProvider, autoExpandChildCategories, onDataLoadedRef]);
+  }, [filterer, dataProvider, autoExpandChildCategories]);
 
   if (!filteringDataProvider) {
     return null;
@@ -115,16 +111,13 @@ class AutoExpandingPropertyFilterDataProvider implements IPropertyDataProvider, 
   public onDataChanged = new PropertyDataChangeEvent();
   private _removeListener: () => void;
   private _autoExpandChildCategories = true;
-  private _onDataLoaded?: () => void;
 
   constructor(
     private _wrapped: FilteringPropertyDataProvider,
     autoExpandChildCategories?: boolean,
-    onDataLoaded?: () => void,
   ) {
     // istanbul ignore next
     this._removeListener = this._wrapped.onDataChanged.addListener(() => this.onDataChanged.raiseEvent());
-    this._onDataLoaded = onDataLoaded;
     if (undefined !== autoExpandChildCategories) {
       this._autoExpandChildCategories = autoExpandChildCategories;
     }
@@ -148,7 +141,6 @@ class AutoExpandingPropertyFilterDataProvider implements IPropertyDataProvider, 
 
     const result = await this._wrapped.getData();
     expandCategories(result.categories);
-    this._onDataLoaded?.();
     return result;
   }
 }
--- a/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
+++ b/packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
@@ -23,6 +23,7 @@ import type {
   VirtualizedPropertyGridWithDataProviderProps,
 } from "@itwin/components-react";
 import type { IDisposable } from "@itwin/core-bentley";
+import { useLatest } from "../hooks/UseLatest";
 
 /**
  * Properties for rendering a `FilteringPropertyGrid`.
@@ -33,22 +34,25 @@ export interface FilteringPropertyGridProps extends VirtualizedPropertyGridWithD
   autoExpandChildCategories?: boolean;
   /** @internal */
   filterer: IPropertyDataFilterer;
+  /** Callback called when property data is loaded. */
+  onDataLoaded?: () => void;
 }
 
 /**
  * Creates a filtered data provider before rendering a `VirtualizedPropertyGridWithDataProvider`.
  * @internal
  */
-export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, ...props }: FilteringPropertyGridProps) {
+export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, onDataLoaded, ...props }: FilteringPropertyGridProps) {
   const [filteringDataProvider, setFilteringDataProvider] = useState<AutoExpandingPropertyFilterDataProvider>();
+  const onDataLoadedRef = useLatest(onDataLoaded);
   useEffect(() => {
     const filteringProvider = new FilteringPropertyDataProvider(dataProvider, filterer);
-    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories);
+    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories, onDataLoadedRef.current);
     setFilteringDataProvider(provider);
     return () => {
       provider.dispose();
     };
-  }, [filterer, dataProvider, autoExpandChildCategories]);
+  }, [filterer, dataProvider, autoExpandChildCategories, onDataLoadedRef]);
 
   if (!filteringDataProvider) {
     return null;
@@ -111,13 +115,16 @@ class AutoExpandingPropertyFilterDataProvider implements IPropertyDataProvider, 
   public onDataChanged = new PropertyDataChangeEvent();
   private _removeListener: () => void;
   private _autoExpandChildCategories = true;
+  private _onDataLoaded?: () => void;
 
   constructor(
     private _wrapped: FilteringPropertyDataProvider,
     autoExpandChildCategories?: boolean,
+    onDataLoaded?: () => void,
   ) {
     // istanbul ignore next
     this._removeListener = this._wrapped.onDataChanged.addListener(() => this.onDataChanged.raiseEvent());
+    this._onDataLoaded = onDataLoaded;
     if (undefined !== autoExpandChildCategories) {
       this._autoExpandChildCategories = autoExpandChildCategories;
     }
@@ -141,6 +148,7 @@ class AutoExpandingPropertyFilterDataProvider implements IPropertyDataProvider, 
 
     const result = await this._wrapped.getData();
     expandCategories(result.categories);
+    this._onDataLoaded?.();
     return result;
   }
 }
--- a/packages/itwin/property-grid/src/components/Header.tsx
+++ b/packages/itwin/property-grid/src/components/Header.tsx
@@ -45,7 +45,7 @@ function DebouncedSearchBox({ onSearchStringChange, onClose, onOpen, className }
 
     const timeoutId = setTimeout(() => {
       onSearchStringChangeRef.current(inputValue);
-    }, 25);
+    }, 500);
 
     return () => {
       clearTimeout(timeoutId);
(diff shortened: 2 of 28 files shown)


packages/itwin/property-grid/src/components/Header.tsx
```diff
@@ -45,7 +45,7 @@ function DebouncedSearchBox({ onSearchStringChange, onClose, onOpen, className }
 
     const timeoutId = setTimeout(() => {
       onSearchStringChangeRef.current(inputValue);
-    }, 25);
+    }, 500);
```
grigasp, line 48: While 25 is maybe too little, 500 seems a bit too much - why did you make the change?

jasdom: Because when it was set to 25 the `filter-properties` feature would be constantly reported while typing, so it seemed too low. Switched to 500 because that was the debounce timeout set in `UniqueValuesSelector`. What value should I use instead?

grigasp: The difference is that unique values selector may have to make an RPC request to get the filtered values values, while the property grid already has everything and can filter synchronously. Could we instead debounce reporting of the feature, e.g. if you filter 5 times in a timeframe of 1 second, report just one usage.

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -33,27 +35,32 @@ enum MultiElementPropertyContent {
...
     // show standard property grid when selection changes
     return Presentation.selection.selectionChange.addListener(() => {
       setContent(MultiElementPropertyContent.PropertyGrid);
     });
-  }, []);
+  }, [setContent]);
+
+  useEffect(() => {
+    const feature = getFeatureFromContent(content, selectedKeys.length);
+    feature && onFeatureUsed?.(feature);
```
saskliutas, line 62: Why `?.` is needed?

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
```diff
@@ -93,13 +94,36 @@ export function PropertyGridContent({
...
 
+  const filterTextRef = useLatest(filterText);
+  const showNullValuesRef = useLatest(showNullValues);
+
   const [{ width, height }, setSize] = useState({ width: 0, height: 0 });
   const handleResize = useCallback((w: number, h: number) => {
     setSize({ width: w, height: h });
   }, []);
 
+  const reportFilterFeature = useCallback(() => {
```
saskliutas, line 108: I think this should be done in `onSearchTextChange` callback.

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
```diff
@@ -93,13 +94,36 @@ export function PropertyGridContent({
...
+  }, [onFeatureUsed, filterTextRef]);
+
+  useEffect(() => {
+    reportFilterFeature();
+  }, [filterText, reportFilterFeature]);
+
+  useEffect(() => {
+    return dataProvider.onDataChanged.addListener(() => {
+      if (!dataProvider.keys.isEmpty) {
+        onFeatureUsed?.(showNullValuesRef.current ? "hide-empty-values-disabled" : "hide-empty-values-enabled");
```
saskliutas, line 121: I think this should be done in `useNullValueSettingContext`

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -33,27 +35,32 @@ enum MultiElementPropertyContent {
...
 
   useEffect(() => {
     // show standard property grid when selection changes
     return Presentation.selection.selectionChange.addListener(() => {
       setContent(MultiElementPropertyContent.PropertyGrid);
     });
-  }, []);
+  }, [setContent]);
+
+  useEffect(() => {
```
saskliutas, line 60: Can we instead of `useEffect` wrap `setContent` to report feature usage?

saskliutas (changes requested):

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
saskliutas: On the second though, we probably don't want to wrap setContent, and just call `onFeatureUsed` before `setContent`. In that case we know what feature to report and don't need additional checks that are done in `getFeatureFromContent`

jasdom: The reason it is currently in the `useEffect` is to report the initial content (before any `setContent` is called) and when either the content or selection changes. If we call `onFeatureUsed` before each `setContent` we would still need to handle the other cases.

jasdom: Moved the reporting closer to `setContent`.

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
jasdom: As discussed, added a `onDataLoaded` callback to `AutoExpandingPropertyFilterDataProvider` and used it to report this setting.

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
jasdom: Moved to `onSearchTextChange`.

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
jasdom: It is not, removed.

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -33,22 +34,25 @@ export interface FilteringPropertyGridProps extends VirtualizedPropertyGridWithD
...
  * @internal
  */
-export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, ...props }: FilteringPropertyGridProps) {
+export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildCategories, onDataLoaded, ...props }: FilteringPropertyGridProps) {
   const [filteringDataProvider, setFilteringDataProvider] = useState<AutoExpandingPropertyFilterDataProvider>();
+  const onDataLoadedRef = useLatest(onDataLoaded);
   useEffect(() => {
     const filteringProvider = new FilteringPropertyDataProvider(dataProvider, filterer);
-    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories);
+    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories, onDataLoadedRef.current);
```
saskliutas, line 50:
```suggestion
    const provider = new AutoExpandingPropertyFilterDataProvider(filteringProvider, autoExpandChildCategories, () => onDataLoadedRef.current?.());
```

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
```diff
@@ -33,29 +35,35 @@ enum MultiElementPropertyContent {
...
+    feature && onFeatureUsed(feature);
+  }, [selectedKeys, onFeatureUsed]);
 
   useEffect(() => {
     // show standard property grid when selection changes
     return Presentation.selection.selectionChange.addListener(() => {
       setContent(MultiElementPropertyContent.PropertyGrid);
     });
-  }, []);
+  }, [setContent]);
```
saskliutas, line 63:
```suggestion
  }, []);
```

packages/itwin/property-grid/src/test/hooks/UseInstanceSelection.test.tsx
```diff
@@ -312,4 +313,72 @@ describe("useInstanceSelection", () => {
       expect(result.current.ancestorsNavigationProps.canNavigateUp).to.be.true;
     });
   });
+
+  describe("feature usage reporting", () => {
+    const initialProps: InstanceSelectionProps = {
+      imodel,
+    };
+
+    it("reports when navigates to parent", async () => {
```
saskliutas, line 322: I think these tests can be combined into one, because both of them need to `navigateUp`

packages/itwin/property-grid/src/test/hooks/UseInstanceSelection.test.tsx
```diff
@@ -312,4 +313,72 @@ describe("useInstanceSelection", () => {
...
+      });
+
+      await act(async () => result.current.ancestorsNavigationProps.navigateUp());
+
+      await waitFor(() => {
+        expect(result.current.selectedKeys[0].id).to.be.eq(parentKey.id);
+        expect(result.current.ancestorsNavigationProps.canNavigateDown).to.be.true;
+      });
+
+      expect(selectionManager.replaceSelection).to.be.calledOnceWith(
```
saskliutas, line 366: Does not like this is relevant to the test.

packages/itwin/property-grid/src/test/components/PropertyGridContent.test.tsx
```diff
@@ -217,4 +221,218 @@ describe("<PropertyGridContent />", () => {
...
+      dataProvider.getData.reset();
+      dataProvider.getData.resolves({
+        categories: [],
+        records: {},
+        label: PropertyRecord.fromString("New Test Label"),
+        description: "NewTestClassName",
+      });
+      act(() => provider.onDataChanged.raiseEvent());
+
+      rerender(
```
saskliutas, line 419: Why is this `rerender` is needed? Looks like `PropertyGridContent` is rerendered with same props.

packages/itwin/property-grid/src/test/components/MultiElementPropertyGrid.test.tsx
```diff
@@ -9,7 +9,7 @@ import { PropertyRecord, PropertyValueFormat } from "@itwin/appui-abstract";
 import { KeySet } from "@itwin/presentation-common";
 import { PresentationLabelsProvider, PresentationPropertyDataProvider } from "@itwin/presentation-components";
 import userEvents from "@testing-library/user-event";
-import { AncestorsNavigationControls, MultiElementPropertyGrid, PropertyGridManager } from "../../property-grid-react";
+import { AncestorsNavigationControls, MultiElementPropertyGrid, PropertyGridManager, TelemetryContextProvider } from "../../property-grid-react";
```
saskliutas, line 12: Can you fix this to not import from barrel file?

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
```diff
@@ -197,3 +218,20 @@ function useFilterer({ showNullValues, filterText }: UseFiltererProps) {
...
+function useThrottled(func: () => void, timeout: number) {
+  const [isThrottled, setIsThrottled] = useState(false);
+  const isThrottledRef = useLatest(isThrottled);
+  const funcRef = useLatest(func);
+
+  return () => {
+    if (isThrottledRef.current) {
+      return;
+    }
+    setIsThrottled(true);
```
saskliutas, line 231: Why is state needed? Isn't `isThrottledRef` enough for this?

saskliutas (changes requested):

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
jasdom: It is not, leftover after refactoring, removed.

packages/itwin/property-grid/src/test/components/MultiElementPropertyGrid.test.tsx
jasdom: Fixed.

packages/itwin/property-grid/src/test/components/PropertyGridContent.test.tsx
jasdom: Without the `rerender` it would not update the displayed data. Removed the test as it no longer is relevant.

packages/itwin/property-grid/src/test/hooks/UseInstanceSelection.test.tsx
jasdom: You are right, removed.

packages/itwin/property-grid/src/test/hooks/UseInstanceSelection.test.tsx
jasdom: Merged.

packages/itwin/property-grid/src/components/MultiElementPropertyGrid.tsx
jasdom: Removed.

packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
jasdom: Removed as it was no longer necessary.

packages/itwin/property-grid/src/hooks/UseNullValuesSetting.tsx
```diff
@@ -25,26 +26,33 @@ export function NullValueSettingContext({ children }: PropsWithChildren<{}>) {
...
   const [showNullValues, setShowNullValues] = useState(true);
   const { getShowNullValuesPreference, setShowNullValuesPreference } = useNullValueStorage();
+  const { onFeatureUsed } = useTelemetryContext();
+  const firstReportRef = useRef(false);
 
   // Get value from preferences storage
   useEffect(() => {
     void (async () => {
       const res = await getShowNullValuesPreference();
+      if (!firstReportRef.current) {
```
saskliutas, line 36: I don't think we want to report only the first value. If preferences storage changes value will be updated and we should report that.

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
```diff
@@ -197,3 +205,19 @@ function useFilterer({ showNullValues, filterText }: UseFiltererProps) {
 
   return compositeFilterer;
 }
+
+function useThrottled(func: () => void, timeout: number) {
+  const isThrottledRef = useLatest(false);
+  const funcRef = useLatest(func);
```
saskliutas, line 211: I don't think we need this ref as `useThrottled` returns new callback on each call.

packages/itwin/property-grid/src/hooks/UseNullValuesSetting.tsx
jasdom: Would it not update when the `updateShowNullValues` below is called? The reason I added the `firstReportRef` is that now it runs multiple times at the start and reports more than once.

packages/itwin/property-grid/src/components/PropertyGridContent.tsx
jasdom: Removed.

packages/itwin/property-grid/src/hooks/UseNullValuesSetting.tsx
saskliutas: I guess it runs multiple times because of `React.StrictMode` rendering components twice for debugging purposes. (you can comment it out and it should report only once). `updateShowNullValues` will report when button is clicked. However, storage can be changed during runtime in which case setting value will change but won't be reported.

jasdom: Seems like you are right, removed the `firstReportRef`.

saskliutas (approved):

PR #824, opened 2024-04-26
geo-tools - Use ComboBox from itwinui-react instead of AutoSuggest 

ashar-bentley:
The Autosuggest component is causing a crash when using the geo-tools in an electron app. Autosuggest is 3rd party component that is no longer being maintained and mdastous-bentley suggested using a react component from itwinui-react; Combobox. 

Also added launch configuration for debugging the geo-tools tests

### DIFF
--- a/packages/itwin/geo-tools/src/components/GeoAddressSearch.scss
+++ b/packages/itwin/geo-tools/src/components/GeoAddressSearch.scss
@@ -5,22 +5,26 @@
 
 
 .geotools-geoaddresssearch__container {
-  width: 300px
+  width: 300px;
+
+  ._iui3-end-icon {
+    display: none; // we don't want the dropdown arrow icon to be shown
+  }
 }
 
 .geotools-geoaddresssearch__button {
   background-color: transparent;
   position: absolute;
-  top: 6px;
-  right: 5px;
+  top: 0.8rem;
+  right: 0.5rem;
   border: none;
   z-index: 1;   // Make it display over the input
 }
 
 .geotools-geoaddresssearch__autosuggest {
   width: 100%;
-  .uicore-autosuggest__input {
-    padding: 1.3px 30px 1.3px 5px;    //right padding avoid input text to be overlaid with the button
+  #iui-geoAddressSearch-input-cb {
+    padding: 0.2rem 2.2rem 0.2rem 0.5rem;    //right padding avoid input text to be overlaid with the button
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/itwin/geo-tools/src/components/GeoAddressSearch.scss
```diff
@@ -5,22 +5,26 @@
 
 
 .geotools-geoaddresssearch__container {
-  width: 300px
+  width: 300px;
+
+  ._iui3-end-icon {
```
jjbeckman13, line 10: We should avoid overriding styles in iTwinUi. Change on their side could unknowingly break us

mdastous-bentley (commented):
I have no problem re-using the Combobox component to create the search input, but we should make sure no artifacts are left.  

For example, when you first click in the input field, you can see an empty submenu shadow appear:
image

packages/itwin/geo-tools/src/components/GeoAddressSearch.scss
ashar-bentley: I have removed the iTwinUI change now. Will request a change on their end to allow hiding the end icon.

packages/itwin/geo-tools/src/components/GeoAddressSearch.scss
```diff
@@ -5,22 +5,26 @@
...
+  right: 0.5rem;
   border: none;
   z-index: 1;   // Make it display over the input
 }
 
 .geotools-geoaddresssearch__autosuggest {
   width: 100%;
-  .uicore-autosuggest__input {
-    padding: 1.3px 30px 1.3px 5px;    //right padding avoid input text to be overlaid with the button
+  #iui-geoAddressSearch-input-cb {
```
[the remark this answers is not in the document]
ashar-bentley: iTwinUI classes get renamed, so I wanted to await using an obscure class name selector. I have now removed the id selector, and realized that we can apply the padding directly to the container

ashar-bentley:
> I have no problem re-using the Combobox component to create the search input, but we should make sure no artifacts are left.
> 
> For example, when you first click in the input field, you can see an empty submenu shadow appear: image

I will try to get it resolved in the iTwinUI

ashar-bentley:
> I keep running into runtime errors when typing values in the input field: image

I was able to reproduce with the test-Viewer. The latest changes fix the issue.

aruniverse:
@ashar-bentley is this pr still needed? if not, plz close it

ashar-bentley:
I was waiting for iTwinUI changes to be resolved first. Maybe I will reopen once that is in place

PR #826, opened 2024-04-29
[ReportsConfig] Extract API

Guillar1:


packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
```diff
@@ -0,0 +1,129 @@
...
+import { default as React_2 } from 'react';
+import type { Report } from '@itwin/insights-client';
+import { ReportsClient } from '@itwin/insights-client';
+import { StagePanelLocation } from '@itwin/appui-react';
+import { StagePanelSection } from '@itwin/appui-react';
+import type { UiItemsProvider } from '@itwin/appui-react';
+import type { Widget } from '@itwin/appui-react';
+
+// @public (undocumented)
+export class BulkExtractor {
```
MindaugasButkus, line 23: I wonder if this should be exported. It's a pretty hacky solution for tracking running extractions.

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
```diff
@@ -0,0 +1,129 @@
...
+    onClickClose: () => void;
+    // (undocumented)
+    report: Report;
+}
+
+// @public
+export const Reports: ({ onClickAddReport, onClickReportModify, onClickReportTitle, }: ReportsProps) => JSX.Element | null;
+
+// @public (undocumented)
+export const REPORTS_CONFIG_BASE_URL = "https://api.bentley.com";
```
MindaugasButkus, line 69: I didn't see the base url exposed for other widgets.

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
Guillar1: It's out in the wild already. I can just mark it internal.

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
Guillar1: G&M widget does not need it yet.. EC3 does not have it defined anywhere, you can define it on your own. I don't even think its worth exporting this. What do you think?

MindaugasButkus: Yeah, I don't think it's needed. For now mark it as internal if it is already exposed.

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
Guillar1: Ouch. 
image
I believe the reason it's there is to allow mocking of the BulkExtractor.

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
Guillar1: Done.

MindaugasButkus (approved):

packages/itwin/reports-config-widget/api/reports-config-widget-react.api.md
MindaugasButkus: Sounds like a bigger change than I expected initially. I think it's better to defer fixing this for now.

PR #831, opened 2024-05-15
[G&M] Overlap progress fix

Guillar1:


packages/itwin/grouping-mapping-widget/src/components/Groups/GroupListItem.tsx
```diff
@@ -52,17 +52,24 @@ export const GroupListItem = ({
...
-          groupElementsInfo={groupElementsInfo}
-        />}
-      <GroupMenuActions
-        group={group}
-        disableActions={disableActions}
-        {...rest}
-      />
+      <div className="gmw-group-list-end">
+        <div className="gmw-group-menu-actions">
+          <GroupMenuActions
```
MindaugasButkus, line 57: I see that the order of components changed. Could you provide before/after screenshots of how it looks? Or the OverlapProgress was not displayed at all in any case?

Guillar1: I incorrectly ordered these components when I switched to the new list components. It was not displayed at all because of this. I must have forgotten to restore it to a working state when probably debugging something. This PR brings it back to how it was before the switch. 
Before (last working state):
image
Now (with a slight adjustment to the spacing):
image

MindaugasButkus (approved):

PR #834, opened 2024-05-16
Added copy property value button in Map Layers Info widget

mdastous-bentley:
Added a new 'copy' button for each property in the Map Layers info widget :
image


packages/itwin/map-layers/src/ui/widget/FeatureInfoWidget.tsx
```diff
@@ -58,6 +61,20 @@ export function MapFeatureInfoWidget({ featureInfoOpts }: MapFeatureInfoWidgetPr
     setSize({ width: w, height: h });
   }, []);
 
+  const copyButton = React.useCallback(
+    (props: ActionButtonRendererProps) => props.isPropertyHovered &&
+      <div>
+        <IconButton styleType="borderless" onClick={()=>{
+          const value = props.property.value;
+          if (value !== undefined && value.hasOwnProperty("displayValue") )
+            navigator.clipboard.writeText((value as PrimitiveValue).displayValue??"").catch((_) => { });
```
eringram, line 70: Does this call need a `.catch()` if its empty?

mdastous-bentley: > Does this call need a `.catch()` if its empty?

writetext is async, and wanted to avoid lint errors about no-floating-promises

eringram (approved):

mdastous-bentley:
@ben-polinsky I think we need your approval too

packages/itwin/map-layers/src/ui/widget/FeatureInfoWidget.tsx
```diff
@@ -58,6 +61,20 @@ export function MapFeatureInfoWidget({ featureInfoOpts }: MapFeatureInfoWidgetPr
...
+        }}>
```
ben-polinsky, line 71:
```suggestion
        <IconButton styleType="borderless" onClick={() => {
          const value = props.property.value;
          if (value !== undefined && value.hasOwnProperty("displayValue"))
            navigator.clipboard.writeText((value as PrimitiveValue).displayValue ?? "").catch((_) => { });
        }}>
```

ben-polinsky (approved):

PR #837, opened 2024-05-22
Distance and area measurements using sheet scale when measuring in sheets

Maxime-Brassard:
This uses already present schema data to scale measurements in drawing so that they represent real distances instead of sheet distances. This change is made for distance measurements and area measurements. Further tools will require more schema information before they can show accurate 3d data in sheets.

This does not affect measurements made outside sheets in any way and is activated from a parameter passed to the UIProvider of the ToolCommand.

This also features showing an outline and preventing clicks outside a drawing when the measurement starts in a drawing.

Here's 2 measurements. One made outside the drawing so it uses sheet coordinates and one inside that applies the scaling.
compare

Here's the outline displayed when starting a measurement in a drawing.
outline


packages/itwin/measure-tools/src/api/Polygon.ts
```diff
@@ -66,10 +68,19 @@ export class Polygon {
     this._textMarker.applyStyle(this._styleSet.getTextStyle(WellKnownTextStyleType.AreaMeasurement));
   }
 
-  constructor(points: Point3d[], copyPoints: boolean = true, styleSet?: StyleSet) {
+  public set sheetToWorldScale(scale: number | undefined) {
+    this._sheetToWorldScale = scale ? (scale * scale): undefined;
```
a-gagnon, line 72: Make sure to check against `undefined` when dealing with numbers, eg. `undefined !== scale` ? (scale * scale) : undefined

change/@itwin-measure-tools-react-94a7f921-b7a3-4caf-b3e2-9ab99a2fafb6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Measurement in sheets",
+  "packageName": "@itwin/measure-tools-react",
+  "email": "email not defined",
```
aruniverse, line 5: @Maxime-Brassard  can you add your gh email here?

packages/itwin/measure-tools/src/api/Polygon.ts
a-gagnon: Also, keep it as the raw value, do not square it. That makes it very inconsistent and unpredictable.

ex:
```ts
const p: Polygon
p.sheetToWorldScale = 3;
console.log(p.sheetToWorldScale); // 9
p.sheetToWorldScale = p.sheetToWorldScale;
console.log(p.sheetToWorldScale); // 81
```
I think you get the point.

change/@itwin-measure-tools-react-94a7f921-b7a3-4caf-b3e2-9ab99a2fafb6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
aruniverse, line 2:
```suggestion
  "type": "minor",
```

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -42,6 +42,7 @@ import { MeasureTools } from "../MeasureTools";
  */
 export interface AreaMeasurementProps extends MeasurementProps {
   polygonPoints: XYZProps[];
+  ratio?: number;
```
a-gagnon, line 45: Please use a name that is more meaningful. ratio can go either side (do I multiply or divide by ratio??)

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -433,18 +463,26 @@ export class AreaMeasurement extends Measurement {
...
-        value: fAreaXY,
-        aggregatableValue:
-          areaSpec !== undefined
-            ? { value: this._polygon.areaXY, formatSpec: areaSpec }
-            : undefined,
-      },
+      }
+    );
+
+    if (this._sheetToWorldScale === undefined) {
```
a-gagnon, line 469: Can you elaborate why we need to check for sheet to world scale? Seems we can compute the area either case.

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -51,6 +51,7 @@ export interface DistanceMeasurementProps extends MeasurementProps {
   startPoint: XYZProps;
   endPoint: XYZProps;
   showAxes?: boolean;
+  ratio?: number;
```
a-gagnon, line 54: Variable name is misleading.

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -474,9 +503,9 @@ export class DistanceMeasurement extends Measurement {
         QuantityType.LengthEngineering
       );
 
-    const distance = this._startPoint.distance(this._endPoint);
-    const run = this._startPoint.distanceXY(this._endPoint);
-    const rise = this._endPoint.z - this._startPoint.z;
+    const distance = this.sheetToWorldScale * this._startPoint.distance(this._endPoint);
+    const run = this.sheetToWorldScale * Math.abs(this._endPoint.x - this._startPoint.x);
+    const rise = this.sheetToWorldScale * Math.abs(this._endPoint.y - this._startPoint.y);
```
a-gagnon, line 508: Rise can be positive or negative, no need for `Math.abs`.
(Run is always positive since it's a distance, which is correct)

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -545,35 +574,41 @@ export class DistanceMeasurement extends Measurement {
...
-      {
-        label: MeasureTools.localization.getLocalizedString(
-          "MeasureTools:tools.MeasureDistance.endCoordinates"
-        ),
-        name: "DistanceMeasurement_EndPoint",
-        value: fEndCoords,
-      }
     );
+
+    if (this._sheetToWorldScale === undefined) {
```
a-gagnon, line 579: Is this needed? Alternatively, we can strip out some of those fields, or put values for both
like an entry for
`Delta Y`
and
`Delta Y (sheet)`

Might want to double check with product management what they expect.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -172,6 +223,14 @@ MeasureAreaToolModel
     return new MeasureAreaToolModel();
   }
 
+  public override decorate(context: DecorateContext): void {
+    super.decorate(context);
+
+    if (this._enableSheetMeasurements && this._currentMousePoint !== undefined && this.toolModel.drawingOrigin !== undefined && this.toolModel.drawingExtents !== undefined && !SheetMeasurementsHelper.checkIfInDrawing(this._currentMousePoint, this.toolModel.drawingOrigin, this.toolModel.drawingExtents)) {
```
a-gagnon, line 229: Feels like the drawing model boundary should be displayed regardless of where the mouse is. It would greatly simplify the 'if' logic.
Also, the `this._currentMousePoint` won't work for mobile devices.

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
bsy-nicholasw, line 54: I don't like such a generic term here called `ratio`  when the measurement calls it `sheetToWorldScale`. Also, why wouldn't the `sheetViewId` not be saved?

I think this information should be promoted to be a base class property (like an object property, that if defined, then has all the sheet metadata on it). Presumably other measurements may be supported in the sheet view, and there should be a standard way of handling that implementation.

change/@itwin-measure-tools-react-94a7f921-b7a3-4caf-b3e2-9ab99a2fafb6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Measurement in sheets",
+  "packageName": "@itwin/measure-tools-react",
+  "email": "maxime.brassard@bentley.com",
```
a-gagnon, line 5: You can use @bentley, but ideally use the no-reply email from GH to avoid being spammed with undesirable stuff since this is public source

bsy-nicholasw: RIP his email folder

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
Maxime-Brassard: We do not want to display the XY area if it's a sheet measurement since there's no point, it's in 3d so any area is XY anyway.
If we really wanted to show it, it would always display as 0 for cross sections and would always be equal to the regular area for plan views (since there is no Z depth).

packages/itwin/measure-tools/src/api/Polygon.ts
Maxime-Brassard: Good catch

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
Maxime-Brassard: I've made some modifications to have the sheet metadata in the base measurement class, please verify if that's sufficient.
It's not an object but I did not feel like there was enough to justify an object for now.

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -285,6 +304,66 @@ export abstract class Measurement {
...
+      this._drawingMetaData.extents = extents;
+    else
+      this._drawingMetaData = { extents };
+  }
+
+  public get drawingExtents(): Point2d | undefined {
+    return this._drawingMetaData?.extents;
+  }
+
+  public setWorldScale(scale: number | undefined) {
```
bsy-nicholasw, line 356: Having `setWorldScale` as a function while the others are getter/setter is off. Note a lot of the base properties have an `onXYZChanged` that subclasses can override, I would do the same for the drawing metadata properties (or maybe a single on changed?)

Also, I still think it's reasonable that at least worldScale/origin should be required properties on the metadata. Probably do some validation for the world scale, since usually scales cannot be zero (in this case, probably also not negative).

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -42,6 +43,7 @@ import { MeasureTools } from "../MeasureTools";
  */
 export interface AreaMeasurementProps extends MeasurementProps {
   polygonPoints: XYZProps[];
+  drawingMetaData?: DrawingMetaData;
```
bsy-nicholasw, line 46: Move `drawingMetaData` up to `MeasurementProps`

Maxime-Brassard: Opps, I had forgotten to remove it

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -0,0 +1,84 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { ColorDef, QueryBinder } from "@itwin/core-common";
+import type { DecorateContext, GraphicBuilder, IModelConnection } from "@itwin/core-frontend";
+import { GraphicType } from "@itwin/core-frontend";
+import type { Point3d } from "@itwin/core-geometry";
+import { Point2d } from "@itwin/core-geometry";
+
+interface IdAndLocation {
```
bsy-nicholasw, line 12: This looks very similar to the `DrawingMetadata` interface. I'd consolidate to just using the one defined in Measurement now.

bsy-nicholasw: (Not to mention, this should have been exported since I see it's used in an exported function below)

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -573,6 +596,7 @@ export class AreaMeasurement extends Measurement {
 
     const jsonArea = json as AreaMeasurementProps;
     jsonArea.polygonPoints = pts;
+    jsonArea.drawingMetaData = this.drawingMetaData;
```
bsy-nicholasw, line 599: Also remember to have the base json method read/write the drawing metadata.

You also have a slight problem here, the metadata has `Point2d`, but saving props it should be `XYProps`. The `Point2d` won't roundtrip correctly.

Maxime-Brassard: So I have to decompose the drawingMetaData for it to process properly?

bsy-nicholasw: Something like:

```
export interface DrawingMetaDataProps extends Omit<DrawingMetaData, "origin" | "extents"> {
  origin: XYProps;
  extents?: XYProps;
}
```

And the JSON methods converting it between the props-type and the "runtime-type".

Sometimes when I hit this Props problem, I'll turn the runtime-type into an actual class and not just have it as an interface, especially if there's lots of free functions that help support getting the data or doing something with it. I don't think your `DrawingMetaData` yet has a lot going on to warrant this, but just a general comment about the design philosophy.

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -495,6 +581,20 @@ export abstract class Measurement {
     return true;
   }
 
+  public drawingMetaDataToJSON(): DrawingMetaDataProps | undefined {
```
bsy-nicholasw, line 584: I wouldn't have this as public API on the measurement. Multiple to/from JSON methods are confusing / do not help with discoverability.

For interface types, I usually organize it with a corresponding namespace that has those as functions. E.g.

```
export namespace DrawingMetaData {
 export function toJSON(obj: DrawingMetaData): DrawingMetaDataProps { }

 export function fromJSON(props?: DrawingMetaDataProps): DrawingMetaData | undefined { }
}
```

Also useful for grouping related helper methods and type guards. Although as you can imagine, the more things you cram into a namespace associated with the interface, the more the case for turning it into a class type is made.

Maxime-Brassard: Should be good now

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -285,6 +329,67 @@ export abstract class Measurement {
...
+  public set sheetViewId(id: string | undefined) {
+    this._sheetViewId = id;
+  }
+
+  public get sheetViewId(): string | undefined {
+    return this._sheetViewId;
+  }
+
+  public set drawingId(id: string | undefined) {
+    if (this._drawingMetaData) {
```
a-gagnon, line 350: Having setters for individual properties seem like a slippery slope if they don't end up setting anything.
I'd be in favor of removing all the individual getters/setters except for the `get/set` for the whole drawing metadata object.
That still leaves the possibility of misusing the API by assigning directly inside the object, which would bypass the `onDrawingMetadataChanged` event. (eg. `this.drawingMetaData.sheetViewId = "0x1234").
A comment on the getter is probably sufficient for this, but you can also return Readonly<DrawingMetaData> to enforce it.

To make it easier to change individual properties on the object, you can implement a 'with' function in the DrawingMetaData namespace, eg:

```ts
// Returns a new DrawingMetaData object with one or more properties changed.
export function with(current: DrawingMetaData, overrides?: Partial<DrawingMetaData>): DrawingMetaData {
  return { ...current, ...overrides };
} 
```

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -120,6 +121,8 @@ export class AreaMeasurement extends Measurement {
   constructor(props?: AreaMeasurementProps) {
     super();
 
+    if (props?.drawingMetaData)
```
a-gagnon, line 124: I think the `if (props) this.readFromJSON(props)` line should take care of that in the base Measurement class?

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -285,6 +329,67 @@ export abstract class Measurement {
     this.onTransientIdChanged(prevId);
   }
 
+  public get drawingMetaData(): DrawingMetaData | undefined {
```
a-gagnon, line 332: Please add missing implementation in `Measurement.copyFrom`, `Measurement.readFromJSON`, `Measurement.writeToJSON`

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -120,6 +121,8 @@ export class AreaMeasurement extends Measurement {
...
+      this.drawingMetaData = DrawingMetaData.fromJSON(props.drawingMetaData);
```
a-gagnon, line 125: Please implement `onDrawingMetaDataChanged` and update polygon accordingly.

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -143,6 +144,8 @@ export class DistanceMeasurement extends Measurement {
     this._startPoint = Point3d.createZero();
     this._endPoint = Point3d.createZero();
     this._isDynamic = false;
+    if (props?.drawingMetaData)
```
a-gagnon, line 147: This code should be in measurement base class.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -24,13 +26,17 @@ import { MeasurementViewTarget } from "../api/MeasurementViewTarget";
...
+import type { DrawingMetaData } from "../api/Measurement";
 
 export class MeasureAreaTool extends MeasurementToolBase<
 AreaMeasurement,
 MeasureAreaToolModel
 > {
   public static override toolId = "MeasureTools.MeasureArea";
   public static override iconSpec = "icon-measure-2d";
+  private _enableSheetMeasurements;
+  private _currentMousePoint?: Point3d;
```
a-gagnon, line 39: Variable is only assigned but never read. Please remove.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -24,13 +26,17 @@ import { MeasurementViewTarget } from "../api/MeasurementViewTarget";
...
+import { SheetMeasurementsHelper } from "../api/SheetMeasurementHelper";
+import type { DrawingMetaData } from "../api/Measurement";
 
 export class MeasureAreaTool extends MeasurementToolBase<
 AreaMeasurement,
 MeasureAreaToolModel
 > {
   public static override toolId = "MeasureTools.MeasureArea";
   public static override iconSpec = "icon-measure-2d";
+  private _enableSheetMeasurements;
```
a-gagnon, line 38: This variable is marked as 'any' by default if you don't provide proper type information or default assignment. Same in MeasureDistanceTool.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -110,6 +127,30 @@ MeasureAreaToolModel
     return EventHandled.Yes;
   }
 
+  private async sheetMeasurementsDataButtonDown(ev: BeButtonEvent, initial: boolean) {
```
a-gagnon, line 130: Not sure why the  'first' or 'initial' is needed at all. This needs to be done once. Function can be called directly after the `setMeasurementViewport` which also needs to be done once.

packages/itwin/measure-tools/src/tools/MeasureDistanceTool.ts
```diff
@@ -51,19 +57,25 @@ MeasureDistanceToolModel
...
 
   public async onRestartTool(): Promise<void> {
-    const tool = new MeasureDistanceTool();
+    const tool = new MeasureDistanceTool(this._enableSheetMeasurements);
     if (await tool.run()) return;
 
     return this.exitTool();
   }
 
+  private resetSheetData(): void {
```
a-gagnon, line 72: Shouldn't `this.toolModel.reset(false)` take care of this?
(it's already called by super.onReinitialize)

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -232,6 +252,26 @@ export interface MeasurementEqualityOptions {
   angleTolerance?: number;
 }
 
+export interface DrawingMetaDataProps extends Omit<DrawingMetaData, "origin" | "extents"> {
+  origin: XYProps;
+  extents?: XYProps;
+}
+
+export interface DrawingMetaData {
```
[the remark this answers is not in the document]
bsy-nicholasw: I see `Metadata` all over our codebase too, funny enough the one example where that isn't the case comes from itwin core - `IModelDb::classMetaDataRegistry`

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -232,6 +252,26 @@ export interface MeasurementEqualityOptions {
...
+  /** Id of the drawing */
+  drawingId?: string;
```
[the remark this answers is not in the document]
Maxime-Brassard: This was a discussion with Nicholas where he mentionned the possibility that there could be other types of drawings using this in the future. For example, a runtime local drawing that does not have an ID and that extends infinitely.
The only thing that's always absolutely required (otherwise the concept would not make any sense) is the scale and origin.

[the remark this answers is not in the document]
a-gagnon: I guess that's fair, although that feels like an edge case similar to our profile/xsections. I'd consider making this required and deal with the edge case by supplying `Id64.invalid` or any other string the caller can make sense of. @bsy-nicholasw thoughts?

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -257,6 +297,10 @@ export abstract class Measurement {
   private _transientId?: Id64String; // Not serialized
   private _isVisible: boolean; // Not serialized
 
+  // Used for sheet measurements
+  private _drawingMetaData?: DrawingMetaData;
+  private _sheetViewId?: string;
```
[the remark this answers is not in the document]
Maxime-Brassard: Yes, the information is in drawing coordinates because it's to help locate the drawing in the sheet: 
The origin of the drawing in the sheet and the extent of the drawing in the sheet.
To me, it would not make sense to put it in the drawingMetadata since there can be multiple drawings in the sheet and we often compare multiple drawings that are in the same sheet.
Also, they serve a different purpose and are in different places in the code. The sheetId is mostly used to only display the measurement in the sheet it was created for and the DrawingMetadata is used for the drawing creation and transformation.

packages/itwin/measure-tools/src/api/Measurement.ts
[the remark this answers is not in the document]
bsy-nicholasw: I don't have a terribly strong opinion one way or another, but a lot of times when I've had an ID I usually expect it to not be `Id64.invalid` and undefined if it doesn't exist.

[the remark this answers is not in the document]
bsy-nicholasw: Also, anything that is an `Id64String` should be typed to that and not `string`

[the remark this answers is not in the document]
Maxime-Brassard: I would tend to go with the undefined also

packages/itwin/measure-tools/src/api/Measurement.ts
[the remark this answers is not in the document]
a-gagnon: > To me, it would not make sense to put it in the drawingMetadata since there can be multiple drawings in the sheet and we often compare multiple drawings that are in the same sheet.
The measurement will always be contained within a single drawing on the sheet. It makes sense to keep the sheetId alongside this data. If the same drawing were to be in a different sheet, the placement wouldn't be the same.

> So a `drawingId` is basically a child of a `sheetViewId`? Maybe then the metadata could have a `parentSheetViewId`?
Not a child, it's a reference. This is similar to the concept of geometry parts. A drawing is a container for 2d graphics, and you can reference the drawing as a whole inside a sheet by setting up proper placement (origin, scale). Thus, multiple sheets could reference the same drawing with different combinations of (origin, scale).
I haven't seen a case where a sheet references the SAME drawing more than once, but I guess that would be technically possible too ??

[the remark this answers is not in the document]
Maxime-Brassard: > The measurement will always be contained within a single drawing on the sheet. It makes sense to keep the sheetId alongside this data. If the same drawing were to be in a different sheet, the placement wouldn't be the same.

But there could be cases where we have drawings in sheets but not within any drawing boundaries and we still want to hide them when inside the wrong sheet.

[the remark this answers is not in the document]
Maxime-Brassard: Meaning we still need to have the sheetId even though the drawingMetadata has not been set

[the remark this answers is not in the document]
a-gagnon: That's a good point...
Back to the drawing board -- I wonder if we could augment the MeasurementViewTarget class to support targeting a specific model.
Up until now we only cared about the 'type' of view the measurement lies in. That makes sense for 3d views since they display [0..n] models, but 2d views only display a single model. That could even open the floor to measurements that are bound to particular model(s) in 3d where if you turn one of them off the measurement is hidden.
@bsy-nicholasw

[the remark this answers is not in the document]
bsy-nicholasw: You should be able to create a custom viewtype classifier that identifies based on the model with the current design. The typename may be a little flakey though, as you'd probably need to encode it like: `sheetdrawing::<drawing_model_id>`

packages/itwin/measure-tools/src/api/Measurement.ts
Maxime-Brassard: I removed the getters/setters (except for the worldScale getter) and added readonly to the drawingMetadata getter.

I'm not sure about the "with" function and would like to know more about it if you have time.

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -131,6 +131,10 @@ export class AreaMeasurement extends Measurement {
     if (props) this.readFromJSON(props);
   }
 
+  public setPolygonWorldScale(worldScale: number | undefined) {
```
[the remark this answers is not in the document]
Maxime-Brassard: I need to access it from the toolmodel so I'm thinking renaming it to updatePolygonScale with the proper documentation would suffice?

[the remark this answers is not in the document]
Maxime-Brassard: nevermind, just saw the next comment

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -495,6 +564,10 @@ export abstract class Measurement {
     return true;
   }
 
+  protected shouldDraw(context: DecorateContext): boolean {
```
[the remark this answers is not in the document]
Maxime-Brassard: I'm now calling the isViewportCompatible in the decorates and modified the function accordingly. Please verify if that's correct. Thank you!

packages/itwin/measure-tools/src/api/MeasurementViewTarget.ts
```diff
@@ -213,6 +232,11 @@ export class MeasurementViewTarget {
         return false;
     }
 
+    for (const viewId of this._viewIds) {
```
a-gagnon, line 235: Also need to check that they'Re the same size (see line 222)

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -298,10 +301,19 @@ export class AreaMeasurement extends Measurement {
...
+  public override onDrawingMetadataChanged(): void {
+    this.polygon.worldScale = this.worldScale;
+    this._polygon.recomputeFromPoints();
+  }
+
   public override decorate(context: DecorateContext): void {
     super.decorate(context);
 
-    if (this.polygonPoints.length === 0) return;
+    if (!this.viewTarget.isViewportCompatible(context.viewport))
```
a-gagnon, line 312: No need to this within any Measurement class. Caller will take care of it before calling Measurement.decorate()

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -250,6 +250,9 @@ export class DistanceMeasurement extends Measurement {
   public override decorate(context: DecorateContext): void {
     super.decorate(context);
 
+    if (!this.viewTarget.isViewportCompatible(context.viewport))
```
a-gagnon, line 253: Please remove.

packages/itwin/measure-tools/src/api/MeasurementViewTarget.ts
Maxime-Brassard: Opps! Nice catch

packages/itwin/measure-tools/src/measurements/DistanceMeasurement.ts
```diff
@@ -545,35 +545,41 @@ export class DistanceMeasurement extends Measurement {
...
-      {
-        label: MeasureTools.localization.getLocalizedString(
-          "MeasureTools:tools.MeasureDistance.endCoordinates"
-        ),
-        name: "DistanceMeasurement_EndPoint",
-        value: fEndCoords,
-      }
     );
+
+    if (this.drawingMetaData?.worldScale === undefined) {
```
bsy-nicholasw, line 550: Wouldn't the user expect to see converted start/end coordinates when there's drawing metadata?

Maxime-Brassard: Yes but, currently, we do not have the data to provide them. The conversion matrix to do so is in the works with John but, right now, we can only do scale.

bsy-nicholasw: Oh, I see. I assume then we have this in our backlog?

bsy-nicholasw (approved):

PR #842, opened 2024-05-24
Bump pnpm to v9 and update pipelines

saskliutas:
Bumped pnpm to use latest major version.
Added config to make sure that correct major pnpm version is used.
Updated pnpm version and actions used in pipelines
Updated azure pipelines pnpm setup based on  pnpm docs to enabled caching
Bumped node version in pipelines to 20


.pipelines/templates/change-files-steps.yaml
```diff
@@ -1,14 +1,20 @@
 steps:
-  - task: NodeTool@0
-    displayName: "Use Node 18.16.0"
+  - task: UseNode@1
+    displayName: "Use Node $(NodeVersion)"
     inputs:
-      versionSpec: 18.16.0
+      version: 18.16.0
```
grigasp, line 5: Why not `version: "$(NodeVersion)"` like in `build-steps.yaml`?

grigasp (approved):

PR #848, opened 2024-05-28
Add localization support to stateless tree-widget

jasdom:


grigasp (approved):
Not a fan of every word uppercase in the button tooltip (e.g. "Enable Instance Focus"), but I see that's consistent with other tooltips, so let's keep it that way. Otherwise, looks okay to me (only checked the strings, leaving implementation review for @saskliutas).

saskliutas (approved):

PR #858, opened 2024-05-29
TreeWidget: Stateless tree feature reporting

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/860.


packages/itwin/tree-widget/src/components/trees/stateless/common/components/VisibilityTree.tsx
```diff
@@ -135,11 +149,21 @@ function VisibilityTreeImpl({
           <VisibilityTreeRenderer
             rootNodes={rootNodes}
             {...treeProps}
+            expandNode={(node, isExpanded) => {
```
saskliutas, line 152: Looks like `expandNode`, `onFilterClick` and `onCheckboxClicked` should be memoized as they are used as dependencies in `useEffect` inside `VisibilityTreeRenderer`

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/trees/stateless/common/components/VisibilityTree.tsx
jasdom: Good catch, memoized the callbacks.

packages/itwin/tree-widget/src/components/trees/common/UseFeatureReporting.ts
```diff
@@ -48,3 +48,18 @@ export function useFeatureReporting(props: UseFeatureReportingProps): UseFeature
...
+  reportUsage?: (props: { featureId?: UsageTrackedFeatures; reportInteraction: true }) => void,
+): T {
+  const reportingAction = useCallback(
+    (...args: any[]) => {
+      reportUsage?.({ reportInteraction: true });
+      action(...args);
+    },
+    [action, reportUsage],
+  );
+  return reportingAction as T;
```
saskliutas, line 64: Does this work?
```suggestion
  return useCallback<T>(
    (...args: any[]) => {
      reportUsage?.({ reportInteraction: true });
      action(...args);
    },
    [action, reportUsage],
  );
```

jasdom: It does not because `T` can be a subtype of the constraint so the cast is used to force the result type.

saskliutas: Looks like this one works. Just need to remove explicit return type.
```suggestion
  return useCallback<(...args: Parameters<T>) => void>(
    (...args) => {
      reportUsage?.({ reportInteraction: true });
      action(...args);
    },
    [action, reportUsage],
  );
```

jasdom: Yes, this one seems to work, changed.

saskliutas (approved):

Issue #857, opened 2024-05-29
Stateless models tree: Limit increase button doesn't work


Yato333:
When loading a large iModel (more than 1000 root nodes) in the Test Viewer, this message appears.
Clicking "increase hierarchy level size limit to 10000" does nothing: no requests are being sent and no errors are shown.

![Screenshot 2024-05-29 092517](https://github.com/iTwin/viewer-components-react/assets/29233962/b18d3028-06ad-4303-8b7d-ce5ddf7a7818)

PR #859, opened 2024-05-29
Tree widget: reuse subjects cache

saskliutas:
Part of https://github.com/iTwin/viewer-components-react/issues/853
Fixes https://github.com/iTwin/viewer-components-react/issues/857

Added API for sharing SubjectsModels cache between different parts of ModelsTree.
Also updated SubjectModels cache to run queries without limit.


packages/itwin/tree-widget/src/components/trees/stateless/models-tree/ModelsTreeDefinition.ts
```diff
@@ -89,7 +85,7 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
         ],
       },
     });
-    this._subjectModelIdsCache = new SubjectModelIdsCache(props.imodelAccess);
+    this._subjectModelIdsCache = props.subjectModelIdsCache ?? new SubjectModelIdsCache(props.imodelAccess);
```
grigasp, line 88: As far as I can tell, `props.subjectModelIdsCache` is never `undefined`.

```suggestion
    this._subjectModelIdsCache = props.subjectModelIdsCache;
```

grigasp (approved):

PR #864, opened 2024-05-31
Julio/properties refactor UI

julioalvahash:
New UI design for group properties. 
Removes calculated property, custom calculation fieldset and tables. 
Group property handles all of the property logic.


change/@itwin-grouping-mapping-widget-e8e9e183-2074-4294-8bbe-84cd626f54c6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "removed export for calculated and custom property",
```
Guillar1, line 3: Can you be more specific about what changed? What replaced these two? This is also a `minor` change. We have not started using `major` bumps.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyTable.tsx
```diff
@@ -12,7 +25,6 @@ import { PropertyTable } from "../PropertyTable";
 import { useGroupingMappingApiConfig } from "../../context/GroupingApiConfigContext";
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { usePropertiesClient } from "../../context/PropertiesClientContext";
-
 export interface GroupPropertyTableProps {
```
Guillar1, line 28: Missing line

Guillar1:
Missing gap.
image

Guillar1 (changes requested):
image
Missing gap.

Guillar1 (changes requested):
Visualize Dimensions in calculated property hilites the entire group when toggled on. It does not clear after toggling it off or when you close out the component. It remains hilited.

Guillar1 (changes requested):
image

`No Calculated Property Type` option should be at the top.

Guillar1 (changes requested):
When modifying a custom calculation and clearing it out, it sends an empty string, which returns a 422.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyAction.tsx
```diff
@@ -217,36 +292,87 @@ export const GroupPropertyAction = ({ mappingId, group, groupProperty, onSaveSuc
...
+                    content={`${property.displayLabel}`}
+                    title={`${property.actualECClassName}`}
+                    description={property.categoryLabel}
+                  />
+                ))}
+            </div>
+          </InputGroup>
+        </ExpandableBlock>
+        <div ref={calculatedPropertyActionRef}>
+          <ExpandableBlock title={"Calculated Property"}
```
Guillar1, line 330: This can be made into its own component. 
Something like:
```
interface ScrollableExpandableBlockProps extends ExpandableBlockProps {
  ref: React.RefObject<HTMLDivElement>; <-- parent ref
}

const ScrollableExpandableBlock = forwardRef<HTMLDivElement, ScrollableExpandableBlockProps>(
...
    const handleToggle = useCallback((isExpanding) => {
      if (isExpanding) {
        ...
      }
      onToggle(isExpanding);
    }, [...])

    return (
      <div ref={ref} <-- forwardRef>
        <ExpandableBlock
          title={title}
          endIcon={endIcon}
          isExpanded={isExpanded}
          onToggle={handleToggle}>
          {children}
        </ExpandableBlock>
      </div>
    );
```
https://react.dev/reference/react/forwardRef

change/@itwin-grouping-mapping-widget-e8e9e183-2074-4294-8bbe-84cd626f54c6.json
julioalvahash: Done.

julioalvahash:
> When modifying a custom calculation and clearing it out, it sends an empty string, which returns a 422.

Did not realize empty would send an error, it's fixed.

julioalvahash:
> image
> 
> `No Calculated Property Type` option should be at the top.

Done.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyAction.tsx
julioalvahash: I got working with the fowardRef and created a new component, it could be refactor a bit more so I'll play around with it a little bit more.

packages/itwin/grouping-mapping-widget/src/components/Properties/PropertyExpandableBlock.tsx
```diff
@@ -0,0 +1,19 @@
...
+    return (
+        <div ref={ref}>
+            <ExpandableBlock {...props}>
+                {props.children}
+            </ExpandableBlock>
+        </div>
+    );
+});
+
+
```
Guillar1, line 19: Extra lines.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyTable.tsx
julioalvahash: Fixed with prettier.

packages/itwin/grouping-mapping-widget/src/components/Properties/PropertyExpandableBlock.tsx
julioalvahash: Fixed with prettier.

julioalvahash:
> image Missing gap.

Added gap.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyAction.tsx
julioalvahash: Added useFowardRef hook so that the forwarded ref can be use by the component.

julioalvahash:
> Visualize Dimensions in calculated property hilites the entire group when toggled on. It does not clear after toggling it off or when you close out the component. It remains hilited.

Clear of visualization when cancel and zoom to calculated property done.

packages/itwin/grouping-mapping-widget/src/formula/FormulaDataTypeResolver.ts
```diff
@@ -1,13 +1,15 @@
 /*---------------------------------------------------------------------------------------------
- * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
- * See LICENSE.md in the project root for license terms and full copyright notice.
- *--------------------------------------------------------------------------------------------*/
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import type { DataType } from "@itwin/insights-client";
```
Guillar1, line 5: Run prettier on this file.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyTable.tsx
Guillar1: Still no line between the interface and import statement.

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/ActionPanel.tsx
```diff
@@ -24,7 +25,7 @@ const ActionPanel = ({ onSave, onCancel, isCancelDisabled = false, isSavingDisab
           Save
         </Button>
         {onCancel && (
-          <Button styleType="default" type="button" id="cancel" onClick={onCancel} disabled={isCancelDisabled || isLoading}>
+          <Button styleType="default" type="button" id="cancel" onClick={onCancel} disabled={isCancelDisabled || isLoading} onClickCapture={onCancelCapture}>
```
Guillar1, line 28: On second thought, this might not be enough. When you turn on the Calculated Properties visualization, it does not clear out when you save or toggle it off.

Guillar1 (changes requested):
One more thing, there is a component called `CalculatedPropertyAction`. This one was suppose to be the component equivalent of `CalculatedPropertyActionWithVisuals` but without the visualization portion. Mark it as `@deprecated`. It no longer makes sense to individually use these components.

packages/itwin/grouping-mapping-widget/src/formula/FormulaDataTypeResolver.ts
julioalvahash: Done.

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyTable.tsx
julioalvahash: I though prettier did for me lol, extra line added now.

packages/itwin/grouping-mapping-widget/src/components/SharedComponents/ActionPanel.tsx
julioalvahash: Added onSaveCapture to clear the visualization on save, I've been messing around with visualization logic in CalculatedPropertyActionWithVisualst.tsx and it's not as simple as I though, changing the current implementation affects the behavior of the visualization, it seems like this needs to be a PR in its own, specially because `undefined` is now an option. 
Some thoughts on this: 
- I think the logic could be refactor a little bit, I think two of the useEffects that have the same falsely conditional could be merged together. 
- What would we want the visuals to be when `No Calculated Property` is selected?
The reason why it can't be in a useEffect now is because it would override the color group toggle (if on) from the property menu and clear the visuals, `calcualtedPropertyType` is `undefined` by default before any selection.
- If `No Calculated Property` is selected, would we want to clear out the group coloring all together or just the Spatial Data? If clearing it all, then it would override the previous toggle (if on) like mentioned. 
- What would be behavior be if the toggleSwitch is On but the expandable block is closed? 
- It's like the `undefined` selection and toggleSwitch are competing for control of the visualization logic so it makes it a bit tricky to reset the visuals, but the main concern is how it should behave.

Guillar1: Oh I see, why it's confusing. So I double checked the current behavior and it is incorrect (before this PR). I must changed the behavior accidently when I switched to react-query. I have an older build that I pulled up before this.
The behavior was like this:

1. The PropertyMenu should remember the GroupToggle setting. If we leave the PropertyMenu, it should clear out the Color Group.
2. Going into the Calculated Property Action would clear out the visuals.
3. But leaving Calculated Property Action and going back to the Property Menu would rehilite the group if toggled. 

Reason for this was because we only wanted to hilite one element to show its Spatial Data. Even if ColorGroup is toggled (on). 

I think the behavior now should be.

- When leaving the PropertyMenu it should clear out its Group hilite (when it unmounts). But remember the toggle (should already be implemented). 
- Visualize Calculated Properties should also clear out everything when toggled on and off, or when unmounting the component (cleaning itself up and going away either through cancel or save or some other way). Selecting `No Calculated Property` can either clear out everything or the Spatial Data, up to you, I'm fine either way. 
- Now when the the PropertyMenu comes back it should remember the toggle selection and rehilite right away or not.

julioalvahash: Behavior of visuals accomplished. 
- Leaving property menu clears our visuals but it remembers the toggle if on when back to menu. 
- Visuals appear with toggle on and clear out when off, visuals also clear out when block closes(but remembers the selection) and when `No calculated Property` is selected, if  `No Calculated Property` is selected, it will close the block and clear visuals, visual will also clear out on onSave and onCancel.

julioalvahash:
> One more thing, there is a component called `CalculatedPropertyAction`. This one was suppose to be the component equivalent of `CalculatedPropertyActionWithVisuals` but without the visualization portion. Mark it as `@deprecated`. It no longer makes sense to individually use these components.

Done

packages/itwin/grouping-mapping-widget/src/components/Properties/GroupProperties/GroupPropertyAction.tsx
```diff
@@ -73,6 +73,7 @@ export const GroupPropertyAction = ({ mappingId, group, groupProperty, onSaveSuc
   const [showPropertiesSelectionModal, setShowPropertiesSelectionModal] = useState<boolean>(false);
   const [showSaveConfirmationModal, setShowSaveConfirmationModal] = useState<boolean>(false);
   const [calculatedPropertyType, setCalculatedPropertyType] = useState<CalculatedPropertyType | undefined>(groupProperty?.calculatedPropertyType ?? undefined);
+  const [isCalculatedBlockExpanded, setIsCalculatedBlockExpanded] = useState<boolean>(calculatedPropertyType ? true : false);
```
Guillar1, line 76:
```suggestion
  const [isCalculatedBlockExpanded, setIsCalculatedBlockExpanded] = useState<boolean>(!!calculatedPropertyType);
```

julioalvahash: Changed.

Guillar1 (approved):

PR #865, opened 2024-05-31
Tree widget: Unify stateless tree rendering

saskliutas:
There was some code duplicated between `FilterableTree` and `VisibilityTree`. Introduced new `TreeRenderer` that can be used in both cases to make sure that both types of trees are renderer the same.


packages/itwin/tree-widget/src/components/trees/stateless/common/UseHierarchyVisibility.ts
```diff
@@ -25,19 +26,19 @@ interface UseHierarchyVisibilityProps {
...
 /** @internal */
 export function useHierarchyVisibility({ visibilityHandlerFactory }: UseHierarchyVisibilityProps): UseHierarchyVisibilityResult {
-  const statusMap = useRef(new Map<string, { node: PresentationHierarchyNode; status: VisibilityStatus; needsRefresh: boolean }>());
-  const [state, setState] = useState({
-    getCheckboxStatus: (_node: PresentationHierarchyNode): VisibilityStatus => ({ state: "hidden", isDisabled: true }),
+  const nodesStateMap = useRef(new Map<string, { node: PresentationHierarchyNode; state: TreeNodeCheckboxState; needsRefresh: boolean }>());
+  const [state, setState] = useState<UseHierarchyVisibilityResult>({
+    getCheckboxState: (_node: PresentationHierarchyNode): TreeNodeCheckboxState => ({ state: "off", isDisabled: true }),
     onCheckboxClicked: (_node: PresentationHierarchyNode, _checked: boolean): void => {},
   });
```
jasdom, line 38: Since you explicitly defined the type this should no longer be necessary.

```suggestion
  const [state, setState] = useState<UseHierarchyVisibilityResult>({
    getCheckboxState: () => ({ state: "off", isDisabled: true }),
    onCheckboxClicked: () => {},
  });
```

packages/itwin/tree-widget/src/components/trees/stateless/common/components/BaseTree.tsx
```diff
@@ -3,11 +3,13 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import type { IModelConnection } from "@itwin/core-frontend";
+import type { SchemaContext } from "@itwin/ecschema-metadata";
```
jasdom, line 7: Move this next to other `import type`.

packages/itwin/tree-widget/src/components/trees/stateless/common/components/BaseTree.tsx
```diff
@@ -64,98 +79,97 @@ export function FilterableTree({ imodel, getSchemaContext, hierarchyLevelSizeLim
...
+    return (
+      <Flex alignItems="center" justifyContent="center" flexDirection="column" style={{ width, height }}>
+        <Delayed show={true}>
+          <ProgressRadial size="large" />
+        </Delayed>
+      </Flex>
+    );
+  }
 
+  if (rootNodes.length === 0 && !isLoading) {
```
jasdom, line 146: You removed the condition that was used to display the `noDataMessage` when hierarchy level creation fails for root nodes. It was used for `StatelessExternalSourcesTree` to show the message when `ExternalSources` class is not found in the schema. Maybe there is some better way to handle the case?

saskliutas: I don't think previous approach was good. Changed `ExternalSourcesTreeDefinition` to check if schema has `ExternalSource` class

jasdom (approved):

PR #866, opened 2024-06-01
[ReportsConfig] Report mapping extractions state fix

Guillar1:
The iModel extraction states for the mappings after it finished kept appearing after each rerender. React Strict mode made this bug more obvious.

- Simplified the logic by removing the `initialLoad` and the `jobStarted` state which was no longer necessary.
- The disabled prop that took advantage of jobStarted was also unnecessary. The whole action group gets replaced by a loading spinner anyways. This might change when eventually the component will allow for single mapping runs instead of entire iModel scoped runs, due to how the bulk extractor only keeps states on the iModel level.

### DIFF
--- a/packages/itwin/reports-config-widget/src/widget/components/ReportMappingHorizontalTile.tsx
+++ b/packages/itwin/reports-config-widget/src/widget/components/ReportMappingHorizontalTile.tsx
@@ -26,15 +26,12 @@ export interface ReportMappingHorizontalTileProps {
 
 export const ReportMappingHorizontalTile = (props: ReportMappingHorizontalTileProps) => {
   const [extractionState, setExtractionState] = useState<ExtractionStates>(ExtractionStates.None);
-  const [jobStarted, setJobStarted] = useState<boolean>(true);
   const interval = useRef<number>();
-  const initialLoad = useRef<boolean>(true);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/itwin/reports-config-widget/src/widget/components/ReportMappingHorizontalTile.tsx
```diff
@@ -26,54 +26,56 @@ export interface ReportMappingHorizontalTileProps {
...
     };
     props.jobStartEvent.addListener(listener);
 
     return () => {
       props.jobStartEvent.removeListener(listener);
     };
-  }, [props.jobStartEvent, props.mapping]);
+  }, [props.jobStartEvent, props.mapping.imodelId]);
 
   const getExtractionState = useCallback(async () => {
```
MindaugasButkus, line 44: This callback needs renaming since `get` implies that it will return the extraction state but refreshes the react component state.

MindaugasButkus (approved):

packages/itwin/reports-config-widget/src/widget/components/ReportMappingHorizontalTile.tsx
Guillar1: Good call. Changed.

PR #871, opened 2024-06-04
Tree widget: Stateless models tree hierarchy integration tests

grigasp:
Prerequisite for https://github.com/iTwin/viewer-components-react/issues/853 - want to have the tests in place before making any changes.

- Unify iModel setup APIs with what we have in `presentation` repo (makes moving tests back and forth easier).
- Move Models tree tests from `presentation` repo.


packages/itwin/tree-widget/src/test/trees/models-tree/stateless/ModelsTreeDefinition.test.ts
saskliutas, line 1: I would suggest moving stateless models tree test files from `trees/models-tree/stateless/**` to `trees/stateless/models-tree/**`. This will match source files layout and will have same layout as tests in this PR https://github.com/iTwin/viewer-components-react/pull/829

grigasp: I'm going to do more file moves when all the pending PRs complete, so not sure it's worth changing anything here now.

The plan is to have a structure like this:
```
- src
  - components
    - trees
      - models-tree
        - ...code shared across stateless and old implementations
        - ...stateless implementation code
        - deprecated
          - ...code specific to the old implementation
      - ...other trees
```

saskliutas (approved):

PR #872, opened 2024-06-04
Tree widget: stateless tree e2e tests

jasdom:
Added e2e tests for stateless trees and fixed issue with horizontal scrollbar not always being shown.


packages/itwin/tree-widget/src/components/trees/stateless/categories-tree/CategoriesTreeComponent.tsx
```diff
@@ -79,19 +80,21 @@ function CategoriesTreeComponentImpl({
...
-              imodel={iModel}
-              categories={categories}
-              activeView={viewport}
-              width={width}
-              height={height}
-              filter={filter}
-            />
-          )}
-        </AutoSizer>
+        <div className={classNames("tree-widget-tree-content", density === "enlarged" && "enlarge")}>
```
saskliutas, line 83: I see this uses css classes from https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/tree-widget/src/components/trees/common/TreeRenderer.scss which includes quite a lot of css used for old tree components. Can you add a new cleaned up scss file under `stateless` folder that could be shared between stateless trees with header? This should make everything simpler when we will be removing old tree implementations

saskliutas (changes requested):

grigasp (commented):
Reviewed only the screenshots, filed a couple of issues:
- https://github.com/iTwin/viewer-components-react/issues/873
- https://github.com/iTwin/viewer-components-react/issues/874

packages/itwin/tree-widget/src/components/trees/stateless/Tree.scss
```diff
@@ -0,0 +1,23 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+.tw-tree-with-header {
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+
+  .tw-tree-header {
```
saskliutas, line 12: Is this needed?

packages/itwin/tree-widget/src/components/trees/stateless/categories-tree/CategoriesTree.tsx
```diff
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import "../Tree.scss";
```
saskliutas, line 6: I think this should be imported in `CategoriesTreeComponent` and `ModelsTreeComponent`

saskliutas (changes requested):

packages/itwin/tree-widget/package.json
```diff
@@ -52,15 +52,15 @@
     "@itwin/core-frontend": "^4.0.0",
     "@itwin/core-react": "^4.10.0",
     "@itwin/ecschema-metadata": "^4.0.0",
-    "@itwin/presentation-components": "^5.0.0",
+    "@itwin/presentation-components": "^5.3.2",
```
saskliutas, line 55: Why do we need to bump `presentation-components` peer dep?

saskliutas: I think it should be enough to bump dev dep. No need to bump peer.

packages/itwin/tree-widget/src/components/trees/stateless/Tree.scss
jasdom: No, removed.

packages/itwin/tree-widget/src/components/trees/stateless/categories-tree/CategoriesTree.tsx
jasdom: Fixed.

packages/itwin/tree-widget/package.json
jasdom: Changed.

jasdom:
> Reviewed only the screenshots, filed a couple of issues:
> 
>     * [Tree widget: Stateless Categories tree has too much whitespace in place of the icon #873](https://github.com/iTwin/viewer-components-react/issues/873)
> 
>     * [Tree widget: Bad layout in Categories tree when hierarchy level is filtered with no matches #874](https://github.com/iTwin/viewer-components-react/issues/874)

The whitespace in categories tree was there because I included an empty icon to fix node alignment issues. I filed an issue in `iTwinUI` and it was recently fixed, so after removing the empty icon this issue is resolved.

saskliutas (approved):

Issue #811, opened 2024-04-11
Tree & Property widgets: Clean up internal APIs exposed through barrel file


grigasp:
- Do not export top-level `@internal` APIs through the barrel.
- Use [internal symbols approach](https://github.com/iTwin/itwinjs-core/pull/6783) for `@internal` members of public APIs.

As part of the change, also add a `check-internal` script to the pipeline (see [here](https://github.com/iTwin/presentation/blob/master/scripts/checkInternal.js))

PR #885, opened 2024-06-05
Tree and Property Widget: Cleanup internal APIs

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/811.


scripts/checkInternal.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// only look for non-indented tags which target root level APIs - we still want to allow things
+// like @internal class functions and interface attributes
+const re = /^\/\/ @internal/gm;
+const content = fs.readFileSync(apiSummaryPath, { encoding: "utf8" });
+if (re.test(content)) {
+  console.error(`Fail! Detected exposed @internal APIs - please make sure they're not exported through barrel file and re-generate API summary.`);
+  process.exit(1);
+}
```
grigasp, line 39: I'd like us to hide `@internal` members of `@public` API too. Can you please try to apply the approach described [here](https://github.com/iTwin/itwinjs-core/blob/bced2ae03ac2cc75f671ca7b03a7807acb88944a/core/backend/src/internal/Internal.ts#L6-L60) and if that works - update this script to detect ALL internal APIs rather than only top level ones.

grigasp (changes requested):

scripts/checkInternal.js
grigasp: @saskliutas points out that some of the currently `@internal` props in property-grid package shouldn't be internal at all.

jasdom: Switched to symbols for `@internal` members and updated the script.

grigasp (approved):

saskliutas (approved):

aruniverse (approved):

Issue #867, opened 2024-06-03
Tree Widget: Support zoom to element in stateless models tree


jasdom:
Previous `ModelsTree` supported zooming to an element when a node was double clicked. Add support for such feature in `StatelessModelsTree`.

PR #892, opened 2024-06-06
Tree Widget: Zoom to element in stateless models tree

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/867.


packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeRenderer.tsx
```diff
@@ -45,8 +47,21 @@ export function TreeRenderer({
...
+            onNodeDoubleClick && event.detail !== 2 && onNodeClick?.(nodeId, isSelected, event);
+          }}
+          nodeProps={{
+            onDoubleClick: (event) => {
+              if (nodeProps.isDisabled || "type" in nodeProps.node || !isPresentationHierarchyNode(nodeProps.node)) {
+                return;
+              }
+              onNodeDoubleClick?.(nodeProps.node, !!nodeProps.isSelected);
+              // Select node to not lose highlight
+              !nodeProps.isSelected && onNodeClick?.(nodeProps.node, true, event);
```
grigasp, line 61: does it only loose selection in the tree or does this cause the node to be deselected and afterwards selected in unified selection?

packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeRenderer.tsx
```diff
@@ -45,8 +47,21 @@ export function TreeRenderer({
       return (
         <TreeNodeRenderer
           {...nodeProps}
+          onNodeClick={(nodeId, isSelected, event) => {
```
saskliutas, line 50: Maybe we should handle double clicks in `onNodeClick` as now we have code related to double clicks in both events?

jasdom: Moved into `onNodeClick`.

packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeRenderer.tsx
jasdom: The first click is handled as normal so it does deselect in unified selection if double clicked slowly enough. Seems to work the same as in the old models tree.

grigasp: As discussed, in "extended" selection mode a click on a selected node should not deselect it. I think that solves the problem and we won't need to select it back.

In other selection modes let's keep the behavior you have now.

packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeNodeRenderer.tsx
```diff
@@ -19,11 +19,11 @@ type CoreTreeNodeRendererProps = ComponentPropsWithoutRef<typeof CoreTreeNodeRen
 type TreeNodeRendererProps = CoreTreeNodeRendererProps & TreeNodeRendererOwnProps;
 
 /** @internal */
-export function TreeNodeRenderer({ checkboxProps, ...props }: TreeNodeRendererProps) {
+export function TreeNodeRenderer({ checkboxProps, nodeProps, ...props }: TreeNodeRendererProps) {
   return (
     <CoreTreeNodeRenderer
       {...props}
-      nodeProps={{ className: cx("tw-tree-node", props.isSelected && "selected", props.className) }}
+      nodeProps={{ ...nodeProps, className: cx("tw-tree-node", props.isSelected && "selected", props.className) }}
```
saskliutas, line 26: I guess `...nodeProps` can be removed? Or if you expect that we will need to pass custom nodeProps in the future same need to be done with `contentProps`

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeNodeRenderer.tsx
jasdom: Removed, we can add it when we need it.

packages/itwin/tree-widget/src/components/trees/stateless/common/components/TreeRenderer.tsx
jasdom: Done.

grigasp: For tracking - the change was made in https://github.com/iTwin/presentation/pull/627 and consumed into tree widget with https://github.com/iTwin/viewer-components-react/pull/898.

saskliutas (approved):

Issue #906, opened 2024-06-10
Tree widget: `_` in Categories tree search matches any character


grigasp:
http://localhost:3000/?iTwinId=b391ba44-add7-47a0-8375-f2889a3540e8&iModelId=ba504f88-a479-4156-9d81-658ee169588e

`_` should match the underscore, but not other characters.

![Image](https://github.com/iTwin/viewer-components-react/assets/35135765/59d76ad4-8c9e-4c53-9b67-f1fcab0e3224)

PR #952, opened 2024-06-14
Tree widget: Escape special symbols when searching in CategoriesTree

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/906


grigasp (changes requested):
Please add a test similar to this: https://github.com/iTwin/viewer-components-react/blob/3afe3faad85e58eac1cd583206d54426d139b254/packages/itwin/tree-widget/src/test/trees/models-tree/stateless/ModelsTreeFiltering.test.ts#L1008

grigasp (approved):

Issue #846, opened 2024-05-28
Tree widget: README & CHANGELOG update


grigasp:
Update screenshots, docs, examples. Ensure new features and breaking changes are documented.

PR #956, opened 2024-06-17
Tree widget: Readme 3.0

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/846


saskliutas (approved):

Yato333 (approved):

packages/itwin/tree-widget/README.md
```diff
@@ -169,69 +209,66 @@ In general, the component is expected to be used by advanced users to inspect co
...
+
+### Usage tracking
 
 Components from this package allows consumers to track the usage of specific features.
 
 This can be achieved by passing `onFeatureUsed` function to `CategoriesTreeComponent`, `ModelsTreeComponent`, `IModelContentTreeComponent` or `TreeWidgetUiItemsProvider`. The function is invoked with feature id as the component is being used. List of tracked features:
 
 - `"choose-{tree}"` - when a tree is selected in the tree selector.
-- `"use-{tree}"` - when an interaction with a tree hierarchy happens.
+- `"use-{tree}"` - when an interaction with a tree hierarchy happens (a node is expanded or collapsed).
```
jasdom, line 271: This is now not entirely correct as this feature is also reported when other interactions happen (selecting nodes, changing visibility, filtering, etc.)

grigasp: Hm, I must have missed the change. Was it on purpose?

jasdom: No, it was that way from the start. I asked you whether it should be reported on header button actions and we decided that it is better to only report interactions with the hierarchy, which I assumed includes all interactions.

grigasp: This definitely shouldn't include interactions with header buttons. Since we already have other feature IDs for visibility and filtering, I assumed this one is used only for expand / collapse. But I guess it's okay to include all interactions, as long as we're consistent and don't change the meaning of what we report.

I'll update the readme.

jasdom (approved):

PR #958, opened 2024-06-17
Test-viewer: Add hot reloading for injected packages

saskliutas:
Setup `vite` to watch for injected packages changes inside `node_modules`


apps/test-viewer/vite.config.mts
```diff
@@ -58,3 +70,55 @@ export default defineConfig(({ mode }) => {
...
+          dedupe: ["@itwin/imodels-access-common", "@itwin/imodels-access-frontend"],
+        },
+      };
+    },
+  };
+}
+
+// list of dependencies that should be force optimized. This is needed to avoid errors like this:
+// Uncaught SyntaxError: The requested module '/node_modules/.pnpm/someDep*' does not provide an export named 'someExport'
+const forceIncludes = [
```
Yato333, line 102: Does this have to be edited by hand if dependencies change? Is it possible to extract this from "package.json" or is this a random issue for different packages?

saskliutas: My first attempt was to read dependencies from `package.json` of each injected package but it was not working, I needed to exclude some of those. Eventually I ended up with two list (deps that should be excluded from include 😆  and deps that are not listed as deps in package.json but needed to be included). So it seemed more easier just to add dependencies one by one that need to be force optimized. I guess this list does not need to be updated constantly unless new deps are added.

aruniverse (approved):

Issue #904, opened 2024-06-10
Tree widget: Expanding a Subject node in IModelContent tree takes a long time


grigasp:
http://localhost:3000/?iTwinId=892aa2c9-5be8-4865-9f37-7d4c7e75ebbf&iModelId=3268d9fb-465d-45b3-93d2-1289d010edd0

![Image](https://github.com/iTwin/viewer-components-react/assets/35135765/ca377536-ab33-417c-89fe-7082f7f77d9a)

1. Open the iModel in test viewer
2. Select the stateless iModel Content tree
3. Expand the Subject node starting with "H"

PR #961, opened 2024-06-18
Tree widget: Improve performance of creating Subjects' and Models' hierarchies in iModel content tree

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/904.

The query to get categories under specific models is still pretty slow due to specifics of the iModel - nothing we can do there besides advertising that categories SHOULD NOT be displayed under models...


packages/itwin/tree-widget/src/components/trees/stateless/imodel-content-tree/internal/IModelContentTreeIdsCache.ts
```diff
@@ -0,0 +1,234 @@
...
+    return this._subjectInfos;
+  }
+
+  /** Returns ECInstanceIDs of Subjects that either have direct Model or at least one child Subject with a Model. */
+  public async getParentSubjectIds(): Promise<Id64String[]> {
+    this._parentSubjectIds ??= (async () => {
+      const subjectInfos = await this.getSubjectInfos();
+      const parentSubjectIds = new Set<Id64String>();
+      subjectInfos.forEach((subjectInfo, subjectId) => {
+        if (subjectInfo.childModels.size === 0 && subjectInfo.childSubjects.size === 0) {
```
jasdom, line 124: You inverted the condition and added the extra check `childSubjects.size === 0`, should the same be done in `ModelsTreeIdsCache`?

jasdom (approved):

packages/itwin/tree-widget/src/components/trees/stateless/imodel-content-tree/internal/IModelContentTreeIdsCache.ts
Yato333, line 1: What is the point of copying the exact same file from `models-tree`? Can't you just move it to the common folder and reuse from both components?

packages/itwin/tree-widget/src/components/trees/stateless/imodel-content-tree/internal/IModelContentTreeIdsCache.ts
grigasp: > should the same be done in `ModelsTreeIdsCache`?

No. 

Models tree shows Subject nodes only if they have a non-empty Model down in the hierarchy. We use this function to tell if a Subject node has children, so we have to check child modes.

IModel content tree, on the other hand, always shows Subject nodes, so to tell if a Subject has children we check if they have child Subjects or child Models.

PR #972, opened 2024-06-24
Added callback props for EC3 export

tejassutar2024:
- Added success and failure callback props for EC3 export
- Moved all prop types to new files


packages/itwin/ec3-widget/src/components/EC3/EC3Config.ts
```diff
@@ -96,11 +96,27 @@ export type EC3ConfigPropsWithDefaultReport = EC3ConfigCommonProps & {
...
+  /**
+   * The callback function for EC3 export success.
+   */
+  onExportSucceeded?: (status: EC3JobStatus) => void;
+
+  /**
+   * The callback function for EC3 export failure.
+   */
+  onExportFailed?: (status: EC3JobStatus) => void;
+};
```
Guillar1, line 109: This looks like it can be combined and be more generic. e.g. `onExportResult?: (status:EC3JobStatus) => void`

MindaugasButkus (approved):

Issue #954, opened 2024-06-14
Tree widget: Scroll position is not persisted when switching between widgets


grigasp:
![Image](https://github.com/iTwin/viewer-components-react/assets/35135765/d4f52b5b-34e0-4b4c-a3a4-ceafe2ab99b9)

PR #973, opened 2024-06-25
Tree widget: persist scroll position

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/954.

### DIFF
--- a/packages/itwin/tree-widget/src/components/trees/common/components/BaseTree.tsx
+++ b/packages/itwin/tree-widget/src/components/trees/common/components/BaseTree.tsx
@@ -168,7 +168,7 @@ function BaseTreeRenderer({
 
   return (
     <div style={{ position: "relative", height, overflow: "hidden" }}>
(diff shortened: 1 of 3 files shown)


packages/itwin/tree-widget/src/components/trees/common/components/BaseTree.tsx
```diff
@@ -168,7 +168,7 @@ function BaseTreeRenderer({
 
   return (
     <div style={{ position: "relative", height, overflow: "hidden" }}>
-      <div style={{ overflow: "auto", height: "100%" }}>
+      <div className="tw-tree-renderer-container" style={{ overflow: "auto", height: "100%" }}>
```
saskliutas, line 171: Maybe use `id` instead of css class that does not have any styling?

jasdom: Switched to `id`.

saskliutas (approved):

PR #974, opened 2024-06-26
measure-tools-react: using transforms for location measurements

Maxime-Brassard:
This PR covers a few things:
- Main part is the addition of the transform for measurement tools in sheets that allows for a point in sheet coordinates to be transformed to it's 3d counterpart. This is a 3 part operation which is why is is stored in 3 parts.
- The Location measurement uses this new transform (when activated and used in sheets) to display the accurate 3d coordinates.
- When hovering over a drawing that's not suited for the measurement you are using, prevent clicks and display the stopping sign cursor.
- Using a different location for origin and extents of drawings that is pre-transformed. This fixes the app being unable to locate profile drawings as they had transformed origins.
- Fixed a few typos


packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -6,17 +6,36 @@
...
 
 export namespace SheetMeasurementsHelper {
 
+  export enum DrawingType {
+    CrossSection = 3,
+    Profile = 5,
+    Plan = 6
+  }
+
+  export interface SheetTransformProps {
```
a-gagnon, line 22: By convention, I would not expect instances of classes in an object with the name `Props`. Perhaps rename this to `SheetTransform` or `SheetTransformParams` ?

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -6,17 +6,36 @@
 import { ColorDef, QueryBinder } from "@itwin/core-common";
 import type { DecorateContext, GraphicBuilder, IModelConnection } from "@itwin/core-frontend";
 import { GraphicType } from "@itwin/core-frontend";
-import type { Point3d } from "@itwin/core-geometry";
+import { Point3d} from "@itwin/core-geometry";
```
a-gagnon, line 9: lint: Missing space

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -6,17 +6,36 @@
...
+import { Transform } from "@itwin/core-geometry";
 import { Point2d } from "@itwin/core-geometry";
 import type { DrawingMetadata } from "./Measurement";
 
 export namespace SheetMeasurementsHelper {
 
+  export enum DrawingType {
```
a-gagnon, line 16: I guess this enum maps to an existing enum somewhere else? Please add some information so we know where it comes from.

packages/itwin/measure-tools/src/measurements/LocationMeasurement.ts
```diff
@@ -233,24 +234,27 @@ export class LocationMeasurement extends Measurement {
 
   private async createTextMarker(): Promise<void> {
     const adjustedLocation = this.adjustPointForGlobalOrigin(this._location);
+    let convertedSheetPoint;
+    if (this.drawingMetadata?.transform)
```
a-gagnon, line 238: Instead of copying this code over and over, consider introducing something similar to `adjustPointForGlobalOrigin`. We could also have both adjustments inside a single function.

a-gagnon: Adding to this, please use a more meaningful name for `this.drawingMetadata.transform`, like `drawingToWorld`

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -61,13 +63,34 @@ MeasureAreaToolModel
     this._enableSheetMeasurements = enableSheetMeasurements;
   }
 
+  private async updateDrawingTypeCache() {
+    this._sheetChangeListener();
+    const sheetIds = [];
+    if (this._enableSheetMeasurements) {
+      for (const viewport of IModelApp.viewManager) {
+        if (viewport.view.classFullName === "BisCore:SheetViewDefinition") {
```
a-gagnon, line 71: This will not work for subclasses of SheetViewDefinition. Check for `instanceof SheetViewDefinition` instead

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -61,13 +63,34 @@ MeasureAreaToolModel
...
+
   public async onRestartTool(): Promise<void> {
     const tool = new MeasureAreaTool(this._enableSheetMeasurements);
     if (await tool.run()) return;
 
     return this.exitTool();
   }
 
+  public override async onPostInstall(): Promise<void> {
+    await this.updateDrawingTypeCache();
```
a-gagnon, line 91: Not sure if base class does anything but I'd add `await super.onPostInstall()` for safety.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -35,6 +35,8 @@ MeasureAreaToolModel
   public static override toolId = "MeasureTools.MeasureArea";
   public static override iconSpec = "icon-measure-2d";
   private _enableSheetMeasurements: boolean;
+  private _drawingTypeCache: SheetMeasurementsHelper.DrawingTypeData[] = [];
```
a-gagnon, line 38: Please create a dedicated cache object and reuse it across tools instead of copying the cache code multiple times.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
jason-crow: or `view.isSheetView()`

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
Maxime-Brassard: @a-gagnon Meaning adding a property to the measurementToolBase class instead?

a-gagnon: I was thinking of a separate class implementation -- having this in the base class might not be desirable if someone writes specialized measurements like the section anchored measurements. We wouldn't need/care for sheet information in those cases.

Maxime-Brassard: Sounds good!

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -0,0 +1,39 @@
...
+  public constructor() {
+    this._drawingTypeCache = [];
+  }
+
+  public async updateDrawingTypeCache(iModel: IModelConnection) {
+    this._sheetChangeListener();
+    const sheetIds = [];
+
+    for (const viewport of IModelApp.viewManager) {
+      if (viewport.view.isSheetView()) {
```
a-gagnon, line 28: There could be more than one sheet view open at any given time. I would suggest changing to an array of listeners, namely `_sheetChangeListener: VoidFunction[] = []`

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -57,14 +59,17 @@ export namespace DrawingMetadata {
...
     if (origin !== undefined)
-      return { origin, extents, worldScale: obj.worldScale, drawingId: obj.drawingId };
+      return { origin, extents, worldScale: obj.worldScale, drawingId: obj.drawingId, sheetToWorldTransform: (masterOrigin !== undefined && sheetTov8Drawing !== undefined && v8DrawingToDesign !== undefined) ? {masterOrigin, sheetTov8Drawing, v8DrawingToDesign}: undefined };
     return undefined;
   }
 
   export function fromJSON(json: DrawingMetadataProps): DrawingMetadata {
 
-    return { origin: Point2d.fromJSON(json.origin), worldScale: json.worldScale, drawingId: json.drawingId, extents: Point2d.fromJSON(json.extents)};
+    return { origin: Point2d.fromJSON(json.origin), worldScale: json.worldScale, drawingId: json.drawingId, extents: Point2d.fromJSON(json.extents), sheetToWorldTransform: { masterOrigin: Point3d.fromJSON(json.sheetToWorldTransform?.masterOrigin), sheetTov8Drawing: Transform.fromJSON(json.sheetToWorldTransform?.sheetTov8Drawing), v8DrawingToDesign: Transform.fromJSON(json.sheetToWorldTransform?.v8DrawingToDesign)}};
```
a-gagnon, line 72: Since the `sheetToWorldTransform` in `DrawingMetadata` is optional, I would make sure that `json.sheetToWorldTransform` is defined before definitely assigning it using `fromJSON` methods (which will return an identity transform).

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -688,8 +710,8 @@ export abstract class Measurement {
     this.lockStyle = other.lockStyle;
     this.viewTarget.copyFrom(other.viewTarget);
     this.displayLabels = other.displayLabels;
-    if (other.drawingMetaData)
-      this._drawingMetaData = { origin: other.drawingMetaData.origin.clone(), worldScale: other.drawingMetaData.worldScale, drawingId: other.drawingMetaData.drawingId, extents: other.drawingMetaData.extents};
+    if (other.drawingMetadata)
+      this._drawingMetadata = { origin: other.drawingMetadata.origin.clone(), worldScale: other.drawingMetadata.worldScale, drawingId: other.drawingMetadata.drawingId, extents: other.drawingMetadata.extents, sheetToWorldTransform: other.drawingMetadata.sheetToWorldTransform};
```
a-gagnon, line 714: I think `extents` should be cloned too

packages/itwin/measure-tools/src/measurements/LocationMeasurement.ts
```diff
@@ -287,7 +288,9 @@ export class LocationMeasurement extends Measurement {
   protected override async getDataForMeasurementWidgetInternal(): Promise<
   MeasurementWidgetData | undefined
   > {
-    const adjustedLocation = this.adjustPointForGlobalOrigin(this._location);
+    let adjustedLocation = this.adjustPointForGlobalOrigin(this._location);
```
a-gagnon, line 291: Can use the new method introduced like line 236

packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
```diff
@@ -122,6 +154,23 @@ MeasureLocationToolModel
     return new MeasureLocationToolModel();
   }
 
+  public override isValidLocation(ev: BeButtonEvent, _isButtonEvent: boolean): boolean {
```
a-gagnon, line 157: Shouldn't there be a check to see if we're targeting a drawing in the first place?
like:
```ts
if (true !== ev.viewport?.view.is2d()) {
  return true;
}
```

Maxime-Brassard: True, isSheetView seems even more appropriate

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public constructor() {
+    this._drawingTypeCache = [];
+  }
+
+  public async updateDrawingTypeCache(iModel: IModelConnection) {
+    this._sheetChangeListener.forEach((func) => {
+      func();
+    });
```
a-gagnon, line 26: Need to clear `this._sheetChangesListener` after the `forEach`

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    for (const viewport of IModelApp.viewManager) {
+      if (viewport.view.isSheetView()) {
+        this._sheetChangeListener.push(viewport.onViewedModelsChanged.addListener(async () => this.updateDrawingTypeCache(iModel)));
+        sheetIds.push(viewport.view.id);
+      }
+    }
+
+    for (const id of sheetIds) {
+      this._drawingTypeCache = await SheetMeasurementsHelper.getSheetTypes(iModel, id);
```
a-gagnon, line 37: Should probably `concat` the results, or have a set in case we load the same sheetId twice

a-gagnon (approved):

Issue #955, opened 2024-06-17
Property widget: Need informational message when all properties are filtered out


grigasp:
When all properties are filtered out, the content area of the property widget is empty - need some kind of informational message to tell user that none of the properties match given filter.

![Image](https://github.com/iTwin/viewer-components-react/assets/35135765/4e3162dc-ad1c-49c9-af4c-462a308fe47a)

PR #982, opened 2024-07-04
Property Widget: display message when no properties match filter

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/955.


packages/itwin/property-grid/src/components/FilteringPropertyGrid.tsx
```diff
@@ -46,6 +51,13 @@ export function FilteringPropertyGrid({ filterer, dataProvider, autoExpandChildC
     };
   }, [filterer, dataProvider, autoExpandChildCategories]);
 
+  const { value, inProgress } = useDebouncedAsyncValue(
```
Yato333, line 54: I think `value` should be renamed to `matchesCount`

jasdom: Renamed both `value` and `inProgress`.

packages/itwin/property-grid/public/locales/en/PropertyGrid.json
```diff
@@ -20,6 +20,10 @@
     "too-many-elements-selected": "Too many elements selected",
     "no-elements-selected": "Select element(s) to view properties"
   },
+  "filtering": {
+    "no-matching-properties": "There are no properties matching filter \"{{filter}}\"",
+    "no-non-null-values": "There are no properties containing non-null values"
```
saskliutas, line 25:
```suggestion
    "no-non-null-values": "There are no properties containing non-empty values"
```

saskliutas (approved):

PR #986, opened 2024-07-08
Measure-tools-react: Adds profile transform to DrawingMetadata

Maxime-Brassard:
Adds the profile transform to the drawingMetaData so that we can use it to make measurements in profile sheets.


packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -298,6 +298,9 @@ export interface DrawingMetadata {
   /** Represents the transform from sheet points to 3d points */
   sheetToWorldTransform?: SheetMeasurementsHelper.SheetTransformParams;
 
+  /** Represents the transform from sheet points to distance along alignment (X) and vertical position related to alignment (Y) */
+  profileTransform?: Transform;
```
a-gagnon, line 302: I think I'd stick to the same name as the connector -- `sheetToProfileTransform`

a-gagnon (approved):

PR #992, opened 2024-07-10
Measure-tools-react: drawingTypeCache refactor

Maxime-Brassard:
Modified the drawingTypeCache so that it is easier to use and keeps the drawing types so that it does not run again when it goes back to a previously checked sheet.
Also made it a singleton so that it persists between tools.

Per the comment made on this PR: https://dev.azure.com/bentleycs/Civil-iTwin/_git/Civil-iTwin/pullrequest/441863


packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -7,38 +7,71 @@ import type { IModelConnection } from "@itwin/core-frontend";
...
-  public get drawingtypes(): ReadonlyArray<SheetMeasurementsHelper.DrawingTypeData> {
-    return this._drawingTypeCache;
+  public getDrawingtypes(viewedModelID: string): Readonly<SheetMeasurementsHelper.DrawingTypeData[]> {
+    const result = this._drawingTypeCache.get(viewedModelID);
+    return result ?? [];
+  }
+
+  public constructor(imodel: IModelConnection) {
+    this._drawingTypeCache = new Map<string, SheetMeasurementsHelper.DrawingTypeData[]>();
+    this.updateDrawingTypeCache(imodel);
```
a-gagnon, line 24: should probably be marked as void, or add a catch statement.

What is the expected behavior if this fails ?

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -9,35 +9,47 @@ import { SheetMeasurementsHelper } from "./SheetMeasurementHelper";
 
 export class DrawingDataCache {
```
bsy-nicholasw: There's a number of issues I see with this:

1. While rare, viewports can view different imodels, so it's possible that you're updating the cache with an imodel that is not the same imodel the viewport currently knows about.

2. Database ID's are only unique within the imodel, so it's possible to have the same ID for two sheets coming from different imodels.

3. Viewports can be dropped, and you'll have the listener disposers dangling.

I propose:

```ts
export class DrawingDataCache {
  private _modelToSheets: Map<IModelConnection, Map<Id64String, SheetMeasurementsHelper.DrawingTypeData>>;
  private _viewportModelChangedListeners: Map<Viewport, () => void>;

  public constructor() {
    this._modelToSheets = new Map<IModelConnection, Map<Id64String, SheetMeasurementsHelper.DrawingTypeData>>();
    this._viewportModelChangedListeners = new Map<Viewport, () => void>();

    this.setupEvents();
  }

  public getSheetDrawingDataForViewport(vp: Viewport): SheetMeasurementsHelper.DrawingTypeData | undefined {
    if (!vp.view.isSheetView())
      return undefined;

    const cache = this._modelToSheets.get(vp.iModel);
    if (cache)
      return cache.get(vp.view.id);

    return undefined;
  }

  public async querySheetDrawingData(imodel: IModelConnection, sheetId: Id64String): Promise<SheetMeasurementsHelper.DrawingTypeData> {
    let cache = this._modelToSheets.get(imodel);
    if (!cache) {
      cache = new Map<Id64String, SheetMeasurementsHelper.DrawingTypeData>();
      this._modelToSheets.set(imodel, cache);
    }

    let sheetData = cache.get(sheetId);
    if (!sheetData) {
      sheetData = await SheetMeasurementsHelper.getSheetTypes(imodel, sheetId);
      cache.set(sheetId, sheetData);
    }

    return sheetData;
  }

  private setupEvents() {
    // If an imodel closes, clear the cache for it
    IModelConnection.onClose.addListener((imodel) => {
      this._modelToSheets.delete(imodel);
    });

    // Listen for new viewports opening
    IModelApp.viewManager.onViewOpen.addListener((vp) => {
      this.addViewport(vp);
    });

    // Listen for viewports closing, this also is called when IModelApp shuts down
    IModelApp.viewManager.onViewClose.addListener((vp) => {
      this.dropViewport(vp);
    });
  }

  private addViewport(vp: Viewport) {
    vp.onViewedModelsChanged.addListener((viewport) =>{
      if (!viewport.view.isSheetView())
        return;

      void this.querySheetDrawingData(viewport.iModel, viewport.view.id);
    });
  }

  private dropViewport(vp: Viewport) {
    const listener = this._viewportModelChangedListeners.get(vp);
    if (listener) {
      listener();
      this._viewportModelChangedListeners.delete(vp);
    }
  }
}
```

Then you can add this as a property of the `MeasurementManager` for there to be only one instance, that will also drop cached data as imodels close. Something will have to be done about this being publicly constructable, so it may be a good idea to only expose this via a similarly named functions as my get/query methods:

```ts
export class MeasurementManager {
  public async querySheetDrawingData(imodel: IModelConnection, sheetId: Id64String): Promise<SheetMeasurementsHelper.DrawingTypeData>;

  public getSheetDrawingDataForViewport(vp: Viewport): SheetMeasurementsHelper.DrawingTypeData | undefined;
}
```

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -7,38 +7,71 @@ import type { IModelConnection } from "@itwin/core-frontend";
...
+  }
+
+  public constructor(imodel: IModelConnection) {
+    this._drawingTypeCache = new Map<string, SheetMeasurementsHelper.DrawingTypeData[]>();
+    this.updateDrawingTypeCache(imodel);
   }
 
-  public constructor() {
-    this._drawingTypeCache = [];
+  public destructor() {
```
a-gagnon, line 27: There's already such thing as implementing the `IDisposable` interface, maybe you'll want to rename this to `public dispose(): void`

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
bsy-nicholasw: Added initial populate to the code proposal above.

Also I noticed your current code has an array of data per sheet ID, I missed that, but the same idea would still apply.

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
Maxime-Brassard: I added both.
The desired behavior, if this fails, would be to have the listeners and to have the drawing type cache initiated to []

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
Maxime-Brassard: Alright, I've modified to implement almost everything you mentioned.
I think the differences are the drawingTypeData being arrays for the most part and added some singleton logic.

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -3,42 +3,104 @@
...
+
+  private dropViewport(vp: Viewport) {
+    const listener = this._viewportModelChangedListeners.get(vp);
+    if (listener) {
+      listener();
+      this._viewportModelChangedListeners.delete(vp);
+    }
+  }
+
+  public getSheetDrawingDataForViewport(vp: Viewport): SheetMeasurementsHelper.DrawingTypeData[] | [] {
```
bsy-nicholasw, line 62: The return type should just be `ReadonlyArray<SheetMeasurementsHelper.DrawingTypeData>`,

Likewise for the query method, I think there's a difference between `Readonly<Array<T>>` and `ReadonlyArray<T>`.

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
```diff
@@ -3,42 +3,104 @@
...
+      sheetData = await SheetMeasurementsHelper.getSheetTypes(imodel, viewedModelID);
+      cache.set(viewedModelID, sheetData);
+    }
+
+    return sheetData;
+  }
+
+}
+
+export class DrawingDataCacheSingleton {
```
bsy-nicholasw, line 91: IMO as a matter of style, I wouldn't have a separate type with the word "singleton" in it. Just export `DrawingDataCache` and have it's constructor private, with a `public static get instance(): DrawingDataCache` getter that creates the instance if undefined.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -27,7 +27,7 @@ import { MeasureAreaToolModel } from "../toolmodels/MeasureAreaToolModel";
 import { MeasureTools } from "../MeasureTools";
 import { SheetMeasurementsHelper } from "../api/SheetMeasurementHelper";
 import type { DrawingMetadata } from "../api/Measurement";
-import { DrawingDataCache } from "../api/DrawingTypeDataCache";
+import { DrawingDataCache } from "../measure-tools-react";
```
a-gagnon, line 30: This import is targeting the barrel file -- please update.

packages/itwin/measure-tools/src/tools/MeasureAreaTool.ts
```diff
@@ -155,12 +150,10 @@ MeasureAreaToolModel
...
 
-    if (this._drawingTypeCache) {
-      for (const drawing of this._drawingTypeCache.drawingtypes) {
-        if (SheetMeasurementsHelper.checkIfInDrawing(ev.point, drawing.origin, drawing.extents)) {
-          if (drawing.type !== SheetMeasurementsHelper.DrawingType.CrossSection && drawing.type !== SheetMeasurementsHelper.DrawingType.Plan) {
-            return false;
-          }
+    for (const drawing of DrawingDataCache.getInstance().getSheetDrawingDataForViewport(ev.viewport)) {
+      if (SheetMeasurementsHelper.checkIfInDrawing(ev.point, drawing.origin, drawing.extents)) {
+        if (drawing.type !== SheetMeasurementsHelper.DrawingType.CrossSection && drawing.type !== SheetMeasurementsHelper.DrawingType.Plan) {
```
a-gagnon, line 155: minor: If the 'checkIfInDrawing' is more expensive than the second test (which I think it is), I'd consider reversing the order.

packages/itwin/measure-tools/src/tools/MeasureDistanceTool.ts
```diff
@@ -26,7 +26,7 @@ import { MeasureTools } from "../MeasureTools";
 import { MeasureDistanceToolModel } from "../toolmodels/MeasureDistanceToolModel";
 import { SheetMeasurementsHelper } from "../api/SheetMeasurementHelper";
 import type { DrawingMetadata } from "../api/Measurement";
-import { DrawingDataCache } from "../api/DrawingTypeDataCache";
+import { DrawingDataCache } from "../measure-tools-react";
```
a-gagnon, line 29: Bad import

packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
```diff
@@ -36,7 +36,7 @@ import type { DialogItem, DialogItemValue, DialogPropertySyncItem } from "@itwin
 import { PropertyDescriptionHelper } from "@itwin/appui-abstract";
 import { SheetMeasurementsHelper } from "../api/SheetMeasurementHelper";
 import type { DrawingMetadata, DrawingMetadataProps } from "../api/Measurement";
-import { DrawingDataCache } from "../api/DrawingTypeDataCache";
+import { DrawingDataCache } from "../measure-tools-react";
```
a-gagnon, line 39: Bad import

a-gagnon (commented):
Tools look ok to me, I'll let Nicholas comment on the data cache.

packages/itwin/measure-tools/src/api/DrawingTypeDataCache.ts
bsy-nicholasw: @Maxime-Brassard This was resolved without a code change, just have the return type `ReadonlyArray<SheetMeasurementsHelper.DrawingTypeData>`, best to convey that this is not a copy to callers so they don't mutate the internal data.

Maxime-Brassard: @bsy-nicholasw Damn, thanks for spotting that, not sure how it hapenned... Maybe I missclosed it thinking it was another one.

bsy-nicholasw: @Maxime-Brassard So I played with `Readonly<T[]>` and `ReadonlyArray<T>` in TS playground and it does seem like the type checker treats about as the same... however that empty `[]` you still have in the return type really doesn't make much sense to me.

bsy-nicholasw (approved):

Issue #989, opened 2024-07-09
[measure-tools] Resolve no-internal warnings


DanRod1999:
Address all problems flagged by eslint no-internal.
Current status:

```[tasklist]
### internal - occurrences - solution
- [x] method ToolAdmin.fillEventFromLastDataButton - 2 - update itwin packages to 4.7
- [x] method AccuSnap.getCurrSnapDetail - 2 - update itwin packages to 4.7
- [x] method AccuDrawHintBuilder.- 1 - use AccuDrawHintBuilder.setMatrix
- [x] method IModelConnection.requestSnap - 1 - use  AccuSnap.doSnapRequest
- [x] interface SnapResponseProps - 12 - resolved after switching to AccuSnap.doSnapRequest
- [x] class ToolSettingsState - 2 - update itwin packages to 4.7
- [x] property ToolAdmin.toolSettingsState - 2 - update itwin packages to 4.7
- [x] class TentativeOrAccuSnap - 2 - update itwin packages to 4.7
```
```[tasklist]
### beta - occurrences - solution
- [ ] method ViewManager.invalidateCachedDecorationsAllViews - 1
- [x] method IModelConnection.createQueryReader - 2 - update itwin packages to 4.7
- [ ] function Toolbar - 1
```
Total occurrences across functions
+----------+------------------+
| Tag      | # of Occurrences |
+----------+------------------+
| internal | 24               |
+----------+------------------+
| beta     | 4                |
+----------+------------------+

DanRod1999:
Most off the methods flagged were resolved just by updating to a newer version of itwinjs. This issue goes over most of these changes
https://github.com/iTwin/itwinjs-backlog/issues/782

PR #993, opened 2024-07-10
Resolve Methods flagged by no-internal in measure-tools

DanRod1999:
resolves: https://github.com/iTwin/viewer-components-react/issues/989

utilized work from this issue and related PRs
https://github.com/iTwin/itwinjs-backlog/issues/782

fillEventFromLastDataButton and fillEventFromLastDataButton promoted in: https://github.com/iTwin/itwinjs-core/pull/5720
doSnapRequest api created by: https://github.com/iTwin/itwinjs-core/pull/5720
toolSettingsState promoted in: https://github.com/iTwin/itwinjs-core/pull/5672
createQueryReader promoted from beta in: https://github.com/iTwin/itwinjs-core/pull/5788


DanRod1999:
Most errors were resolved just by updating itwinjs-core packages to a version after the above issue was resolved. 
AccuDrawHintBuilder.setRotation uses setMatrix instead, and could remove the call to .inverse()
IModelConnection.requestSnap uses AccuSnap.doSnapRequest instead, and and needed a few updates to its params and related properties.

packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
```diff
@@ -204,8 +205,9 @@ MeasureLocationToolModel
       snapAperture: hit.viewport.pixelsFromInches(0.1),
     };
 
-    const result = await this.iModel.requestSnap(requestProps);
-    if (result.status !== SnapStatus.Success) return undefined;
+    const accuSnapInst = new AccuSnap();
+    const result = await accuSnapInst.doSnapRequest(hit);
```
bsy-nicholasw, line 209: Wouldn't this be satisfied by using `IModelApp.accuSnap`, instead of creating a new instance of `AccuSnap`? Also looks like you should remove the requestProps variable.

change/@itwin-measure-tools-react-a682d053-c4ed-4aec-8a4c-4e33bbe76d16.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
aruniverse, line 2: since youre bumping the peers, this is technically a major version bump, however since this pkg is 0.x, we can get away with a minor bump

change/@itwin-measure-tools-react-a682d053-c4ed-4aec-8a4c-4e33bbe76d16.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "resolve no-internal warnings",
```
aruniverse, line 3:
```suggestion
  "comment": "replace usage of internal apis",
```

change/@itwin-measure-tools-react-a682d053-c4ed-4aec-8a4c-4e33bbe76d16.json
aruniverse: im not 100% sure why we needed to update deps, is it because some of the replacement apis are only available in newer versions?

DanRod1999: a lot of the functions were promoted to public in versions >4.0, and AccuSnap.doSnapRequest was a new api created in the same pr that promoted a few of the apis
https://github.com/iTwin/itwinjs-core/pull/5720

aruniverse: then youll need to update the changelog entry, and would be good to call out when the apis youre using were added in the description

aruniverse: why is the type here set to `none` ?

packages/itwin/measure-tools/package.json
```diff
@@ -23,6 +23,7 @@
     "rebuild": "npm run clean && npm run build",
     "lint": "eslint -f visualstudio \"./src/**/*.{ts,tsx}\" 1>&2",
     "lint:fix": "npm run lint -- --fix",
+    "lint:table": "eslint ./**/*.{ts,tsx} 1>&2 -f ./node_modules/@itwin/eslint-plugin/dist/formatters/no-internal-summary-with-table.js",
```
aruniverse, line 26: this currently creates a csv, by default when you run this, which should be git ignored 
theres this pr which should fix that https://github.com/iTwin/eslint-plugin/pull/78

packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
```diff
@@ -223,7 +210,7 @@ MeasureLocationToolModel
       result.snapPoint
     );
     snap.setCurvePrimitive(
-      parseCurve(result.curve),
+      parseCurve(result.primitive),
       undefined,
```
aruniverse, line 214: why do we still need `parseCurve` ?

DanRod1999: misinterpreted result.primitive. Thought it was the replacement of result.curve. But I see now that result.primitive is what should be the outcome of parseCurve(result.curve).

change/@itwin-measure-tools-react-a682d053-c4ed-4aec-8a4c-4e33bbe76d16.json
DanRod1999: changing to minor

packages/itwin/measure-tools/package.json
DanRod1999: added to git ignore

aruniverse (approved):

packages/itwin/measure-tools/package.json
anmolshres98: https://www.npmjs.com/package/@itwin/eslint-plugin/v/4.1.1, published

aruniverse: i guess dan, plz update to use this version then, but keep the gitignore change

PR #1008, opened 2024-07-24
Tree Widget 2.x: Clear models tree cache when briefcase changes

jasdom:
Fixes for https://github.com/iTwin/itwinjs-core/issues/6955.


packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeComponent.tsx
```diff
@@ -124,13 +125,20 @@ function ModelsTreeComponentImpl(props: ModelTreeComponentProps & { iModel: IMod
...
+    const queryModels = () => {
+      queryModelsForHeaderActions(iModel)
+        .then((modelInfos: ModelInfo[]) => {
+          setAvailableModels(modelInfos);
+        })
+        .catch((_e) => {
+          setAvailableModels([]);
+        });
+    };
+    queryModels();
```
saskliutas, line 137: Probably need to do something to avoid race conditions?

saskliutas (changes requested):

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeComponent.tsx
jasdom: You are right, updated.

saskliutas (approved):

PR #1009, opened 2024-07-29
Added new property `shouldShow` to `PropertyGridUiItemsProvider`'s props

mdastous-bentley:
When only transient Ids are in the imodel selection set, the grid does not show up.

### DIFF
--- a/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
+++ b/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Visibility of PropertyGridWidget can now be controlled using a custom predicate.",
+  "comment": "Visibility of the widget, when creating it using `PropertyGridUiItemsProvider`, can now be controlled using a `propertyGridProps.shouldShow` predicate, which is invoked on every unified selection change. The default behavior stays the same - show the widget when a node or a non-transient element is selected.",
   "packageName": "@itwin/property-grid-react",
   "email": "mdastous-bentley@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
+++ b/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Visibility of PropertyGridWidget can now be controlled using a custom predicate.",
+  "packageName": "@itwin/property-grid-react",
+  "email": "mdastous-bentley@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
+++ b/change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Visibility of the widget, when creating it using `PropertyGridUiItemsProvider`, can now be controlled using a `propertyGridProps.shouldShow` predicate, which is invoked on every unified selection change. The default behavior stays the same - show the widget when a node or a non-transient element is selected.",
+  "comment": "Visibility of the widget, when creating it using `PropertyGridUiItemsProvider`, can now be controlled using a `propertyGridProps.shouldShow` predicate, which is invoked on every unified selection change. The default behavior stays the same - show the widget when a node or a non-transient element is selected. ",
   "packageName": "@itwin/property-grid-react",
   "email": "mdastous-bentley@users.noreply.github.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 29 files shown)


mdastous-bentley:
@saskliutas Would like to get your feedback about this

grigasp:
This raises two questions to me:
- Is there a reason to ever not show the property grid when a transient element is selected? Maybe we can always do that instead of controlling the behavior through a prop?
- If the answer to the above is "no", then - is it really enough to control this at the application level through a flag, rather than per-element basis? Maybe the check should be based on the element itself?..

saskliutas:
> This raises two questions to me:
> 
> * Is there a reason to ever not show the property grid when a transient element is selected? Maybe we can always do that instead of controlling the behavior through a prop?
> * If the answer to the above is "no", then - is it really enough to control this at the application level through a flag, rather than per-element basis? Maybe the check should be based on the element itself?..

I think when only transient elements are selected property grid with default data provider will be empty. So I think it should be shown for transient elements only when used with custom data provider that returns some data for transient elements.

mdastous-bentley:
> > This raises two questions to me:
> > 
> > * Is there a reason to ever not show the property grid when a transient element is selected? Maybe we can always do that instead of controlling the behavior through a prop?
> > * If the answer to the above is "no", then - is it really enough to control this at the application level through a flag, rather than per-element basis? Maybe the check should be based on the element itself?..
> 
> I think when only transient elements are selected property grid with default data provider will be empty. So I think it should be shown for transient elements only when used with custom data provider that returns some data for transient elements.

How about this?

saskliutas:
> > > This raises two questions to me:
> > > 
> > > * Is there a reason to ever not show the property grid when a transient element is selected? Maybe we can always do that instead of controlling the behavior through a prop?
> > > * If the answer to the above is "no", then - is it really enough to control this at the application level through a flag, rather than per-element basis? Maybe the check should be based on the element itself?..
> > 
> > 
> > I think when only transient elements are selected property grid with default data provider will be empty. So I think it should be shown for transient elements only when used with custom data provider that returns some data for transient elements.
> 
> How about this?

Checking if `createDataProvider` is not `undefined` is not a correct approach because not all custom data providers will work with transient elements. I think your initial approach was ok but it can be future proofed by changing from `boolean` flag to a predicate. Something like `shouldShow?: (selection: KeySet) => boolean`.

Also I think the new property should be added only to `PropertyGridWidget` component and should not be available to `PropertyGridComponent`.

packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -73,8 +75,17 @@ export class PropertyGridUiItemsProvider implements UiItemsProvider {
...
+ * @beta
+ */
+export interface PropertyGridWidgetProps extends PropertyGridComponentProps {
+  /** Predicate indicating if the widget should be shown for the current selection set. */
+  shouldShow?: (selection: Readonly<KeySet>) => boolean
+}
+
 /** Component that renders `PropertyGridComponent` an hides/shows widget based on `UnifiedSelection`. */
-function PropertyGridWidget(props: PropertyGridComponentProps) {
+function PropertyGridWidget(props: PropertyGridWidgetProps) {
```
saskliutas, line 88: Need to destructure props and take out `shouldShow` prop to avoid passing it down to `PropertyGridComponent`
```suggestion
function PropertyGridWidget({ shouldShow, ...props }: PropertyGridWidgetProps) {
```

packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -85,10 +96,16 @@ function PropertyGridWidget(props: PropertyGridComponentProps) {
...
-      if (!hasSelectedElements) {
+      let showWidget = false;
+      if (props.shouldShow) {
+        showWidget = props.shouldShow(selection);
+      } else {
+        // Default behavior:  hide grid widget if there are no nodes or valid instances selected.
+        showWidget = selection.nodeKeysCount !== 0 || selection.some((key) => Key.isInstanceKey(key) && (!Id64.isTransient(key.id)));
+      }
+
+      if (!showWidget) {
```
saskliutas, line 108:
```suggestion
      const predicate = shouldShow
        ? shouldShow
        : // Default behavior:  hide grid widget if there are no nodes or valid instances selected.
          (keys: Readonly<KeySet>) => keys.nodeKeysCount !== 0 || keys.some((key) => Key.isInstanceKey(key) && !Id64.isTransient(key.id));

      if (!predicate(selection)) {
```

saskliutas (approved):
Looks good.

mdastous-bentley:
@saskliutas I was forced to mark `PropertyGridWidgetProps` as public because it's being used in `PropertyGridUiItemsProviderProps`, which is already public.  Let me know if its a problem.

change/@itwin-property-grid-react-f6c86575-936f-497c-a1f3-12d9f0d0f4b2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Visibility of PropertyGridWidget can now be controlled using a custom predicate.",
```
grigasp, line 3: Let's be more explicit about what exactly was added.

```suggestion
  "comment": "Visibility of the widget, when creating it using `PropertyGridUiItemsProvider`, can now be controlled using a `propertyGridProps.shouldShow` predicate, which is invoked on every unified selection change. The default behavior stays the same - show the widget when a node or a non-transient element is selected.",
```

packages/itwin/property-grid/src/PropertyGridUiItemsProvider.tsx
```diff
@@ -97,7 +111,7 @@ function PropertyGridWidget(props: PropertyGridComponentProps) {
         widgetDef.setWidgetState(WidgetState.Open);
       }
     });
-  }, [widgetDef]);
+  }, [props, shouldShow, widgetDef]);
```
grigasp, line 114: is `props` really needed in this list?

mdastous-bentley: good catch, not needed.

Issue #947, opened 2024-06-14
Tree widget: Grouping nodes should not auto-expand when being focused


grigasp:
At the moment, then a grouping node is selected in focus mode, it gets auto-expanded. It should not.

PR #1010, opened 2024-07-31
Tree Widget: Prevent auto-expansion of grouping nodes when focusing on them

jasdom:
Closes https://github.com/iTwin/viewer-components-react/issues/947.

With my very limited testing, it seems to work as intended, however I was not able to add any tests or properly cleanup the code. I ran into some issues with correctly selecting the "closest" instance node that should be used to target a grouping node, and the queries for it ended up quite messy and probably not performant. A lot more testing needs to be done with label and property grouping nodes.


packages/itwin/tree-widget/src/components/trees/models-tree/UseModelsTree.tsx
```diff
@@ -285,46 +286,21 @@ function SvgClassGrouping() {
...
-    parentType: "element" | "category";
-    className: string;
-    groupingNode: GroupingHierarchyNode;
-    modelIds: Id64String[];
-  }> = [];
-  for await (const key of loadFocusedInstancesKeys()) {
-    if ("id" in key) {
-      targetKeys.push(key);
+async function collectTargetKeys(loadFocusedItems: () => AsyncIterableIterator<InstanceKey | ClassGroupingHierarchyNode>) {
+  const targetKeysOrGroupingNodes: Array<InstanceKey | ClassGroupingHierarchyNode> = [];
```
grigasp, line 290: maybe rename `targetKeysOrGroupingNodes` to `focusedItems`?

packages/itwin/tree-widget/src/components/trees/models-tree/UseModelsTree.tsx
```diff
@@ -285,46 +286,21 @@ function SvgClassGrouping() {
...
-    parentKey: InstancesNodeKey;
-    parentType: "element" | "category";
-    className: string;
-    groupingNode: GroupingHierarchyNode;
-    modelIds: Id64String[];
-  }> = [];
-  for await (const key of loadFocusedInstancesKeys()) {
-    if ("id" in key) {
-      targetKeys.push(key);
+async function collectTargetKeys(loadFocusedItems: () => AsyncIterableIterator<InstanceKey | ClassGroupingHierarchyNode>) {
```
grigasp, line 289: I'd rename to `collectFocusedItems`

packages/itwin/tree-widget/src/components/trees/models-tree/UseModelsTree.tsx
```diff
@@ -45,7 +44,9 @@ interface UseModelsTreeProps {
   hierarchyConfig?: Partial<ModelsTreeHierarchyConfiguration>;
   visibilityHandlerOverrides?: ModelsTreeVisibilityHandlerOverrides;
   getFilteredPaths?: (props: {
-    createInstanceKeyPaths: (props: { keys: Array<InstanceKey | ElementsGroupInfo> } | { label: string }) => Promise<HierarchyFilteringPaths>;
+    createInstanceKeyPaths: (
+      props: { keysOrGroupingNodes: Array<InstanceKey | ClassGroupingHierarchyNode> } | { label: string },
```
grigasp, line 48: what do you think about renaming `keysOrGroupingNodes` to `targets`?

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -605,17 +589,12 @@ function createGeometricElementInstanceKeyPaths(
   imodelAccess: ECClassHierarchyInspector & LimitingECSqlQueryExecutor,
   idsCache: ModelsTreeIdsCache,
   hierarchyConfig: ModelsTreeHierarchyConfiguration,
-  elementInfos: Array<Id64String | ElementsGroupInfo>,
+  elementInfos: Array<Id64String | ClassGroupingHierarchyNode>,
```
grigasp, line 592: `elementInfos` -> `targets`

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -676,48 +652,22 @@ function createGeometricElementInstanceKeyPaths(
...
       {
         ctes,
         ecsql,
-        bindings: [
-          ...elementIds.map(createIdBinding),
-          ...groupInfos.flatMap((info) =>
-            info.parent.type === "element" ? info.parent.ids.map(createIdBinding) : createCategoryElementsBindings(info.parent),
-          ),
-        ],
+        bindings: [...elementIds.map(createIdBinding), ...groupingNodes.map((info) => createIdBinding(info.groupedInstanceKeys[0].id))],
```
grigasp, line 662: did you try focusing multiple elements under a grouping node? are all of them loaded with the change?

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -77,14 +77,13 @@ export interface ElementsGroupInfo {
...
-  className: string;
-  groupingNode: GroupingHierarchyNode;
+  groupingNode: ClassGroupingHierarchyNode;
 }
 
 interface ModelsTreeInstanceKeyPathsFromInstanceKeysProps {
   imodelAccess: ECClassHierarchyInspector & LimitingECSqlQueryExecutor;
   idsCache: ModelsTreeIdsCache;
-  keys: Array<InstanceKey | ElementsGroupInfo>;
+  focusedItems: Array<InstanceKey | ElementsGroupInfo>;
```
grigasp, line 86: I don't think this is a good rename. Paths lookup based on target items will probably be most often done due to focusing, but not necessarily. So we shouldn't use the notion of "focus" in this context. I'd change this to `targetItems` or `targets`. Also, I think the interface name should also be changed accordingly, e.g. `ModelsTreeInstanceKeyPathsFromTargetItemsProps`.

Yato333: Renamed to `targetItems` and interface to `ModelsTreeInstanceKeyPathsFromTargetItemsProps`

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
Yato333: Fixed this by creating paths from all grouped elements

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
Yato333: Renamed

packages/itwin/tree-widget/src/components/trees/common/FocusedInstancesContext.ts
```diff
@@ -7,9 +7,12 @@ import { createContext, useContext } from "react";
 
 import type { InstanceKey } from "@itwin/presentation-shared";
 import type { GroupingHierarchyNode } from "@itwin/presentation-hierarchies";
+import type { ClassGroupingNodeKey } from "@itwin/presentation-hierarchies/lib/cjs/hierarchies/HierarchyNodeKey";
+
+export type ClassGroupingHierarchyNode = GroupingHierarchyNode & { key: ClassGroupingNodeKey };
 
 export interface FocusedInstancesContext {
-  loadInstanceKeys?: () => AsyncIterableIterator<InstanceKey | GroupingHierarchyNode>;
+  loadFocusedItems?: () => AsyncIterableIterator<InstanceKey | ClassGroupingHierarchyNode>;
```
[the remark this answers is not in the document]
Yato333: I wanted to move this to `models-tree` folder but this would require additional changes in other trees. In the end, I've decided to revert this change

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -35,6 +37,9 @@ import type {
   InstanceKey,
 } from "@itwin/presentation-shared";
 import type { ModelsTreeIdsCache } from "./internal/ModelsTreeIdsCache";
+import type { ClassGroupingNodeKey } from "@itwin/presentation-hierarchies/lib/cjs/hierarchies/HierarchyNodeKey";
```
grigasp, line 40: should import from barrel

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -599,47 +607,63 @@ function createGeometricElementInstanceKeyPaths(
...
+          AND ${parent.type === "element" ? bind("e.Parent.Id", parent.ids) : `e.Parent IS NULL AND ${bind("e.Category.Id", parent.ids)} AND ${bind("e.Model.Id", parent.modelIds)}`}
+        `,
     );
 
-    const whereClause = [...(elementsClause ? [elementsClause] : []), ...(classClause ?? [])].join(" OR ");
-
     const ctes = [
-      `ModelsCategoriesElementsHierarchy(ECInstanceId, ParentId, ModelId, Path) AS (
+      `InstanceElementsWithClassGroupingNodes(ECInstanceId, ECClassId, ParentId, ModelId, CategoryId, GroupingNodeIndex) AS (
+        ${[...(targetElementsInfoQuery ? [targetElementsInfoQuery] : []), ...targetGroupingNodesElementInfoQueries].join("\n\nUNION ALL\n\n")}
```
grigasp, line 643:
```suggestion
        ${[...(targetElementsInfoQuery ? [targetElementsInfoQuery] : []), ...targetGroupingNodesElementInfoQueries].join(" UNION ALL ")}
```

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -994,13 +1051,21 @@ describe("Models tree", () => {
...
+            }),
+          };
         },
-        (x) => [
-          [x.rootSubject, x.model, x.category, x.testElement1],
-          [x.rootSubject, x.model, x.category, x.testElement2],
-        ],
-        (x) => [{ parent: { type: "category", ids: [x.category.id], modelIds: [x.model.id] }, classes: [x.testElement1.className] }],
+        (x) => x.filterPaths,
+        (x) => [x.filterTargetItem],
```
grigasp, line 1068: Not a big fan of this. Now I have to scroll up to `createFilteredGroupingNodeTestInfo` to see what `filterPaths` and `filterTargetItem` means. Could we only have a function that creates a grouping node and keep the functions as they were - creating the results themselves?

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -1019,17 +1084,8 @@ describe("Models tree", () => {
...
-                            instanceKeys: [x.testElement1],
-                            label: /^test element 1/,
-                          }),
-                          NodeValidators.createForInstanceNode({
-                            instanceKeys: [x.testElement2],
-                            label: /^test element 2/,
-                          }),
-                        ],
+                        autoExpand: false,
+                        groupedInstanceKeys: x.groupingNode.groupedInstanceKeys,
```
grigasp, line 1088: why was the children check removed?

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -1271,29 +1341,119 @@ describe("Models tree", () => {
...
+                    ],
+                  }),
+                ],
+              }),
+            ],
+          }),
+        ],
+      ),
+      TreeFilteringTestCaseDefinition.create(
+        "deeply nested grouping nodes",
```
grigasp, line 1365: We're not filtering grouping nodes, so let's make the test name more specific. Maybe "grouped elements at different hierarchy levels"? Please review names of the other new tests as well.

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
Yato333: Hmm, apparently it's not exported thru barrel. There seems no other way to get this type from `hierarchies`. For the time being, I can copy this type, but we should think about whether or not hierarchy node key types should be exported.

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
Yato333: There were already 4 test cases added by @saskliutas that have names like that, e.g. "grouping element node under category". I just followed the example. Should I rename all of those as well?

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
grigasp: Agree. Can you please add those exports in [the associated PR](https://github.com/iTwin/presentation/pull/688)?

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
grigasp: I don't agree the names are "like that", but I agree that their names could use an improvement.

Saulius' test names mention a "grouping element" it's unclear what that really is. If I didn't see the test source, I'd probably think it's the source of `bis.ElementGroupsMembers` relationship. I think what Saulius meant was "grouped element" - can you please make the change?

Your test names mention "grouping node", which can't be a filter target. I suggest these renames:
- "deeply nested grouping nodes" -> "elements' hierarchy, when auto-expanding their grouping nodes",
- "grouping node and element" -> "element, when auto-expanding up to element and its grouping node",
- "grouping nodes under different categories" -> "elements under different categories, when auto-expanding up to their grouping nodes"

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -1051,21 +1007,27 @@ describe("Models tree", () => {
...
+            parentKeys: pathUntilTargetElement.map((key) => ({ type: "instances", instanceKeys: [key] })),
+          });
+          return { rootSubject, model, category, pathUntilTargetElement, groupingNode };
         },
-        (x) => x.filterPaths,
-        (x) => [x.filterTargetItem],
+        (x) =>
+          x.groupingNode.groupedInstanceKeys.map((elementKey) => ({
+            path: [...x.pathUntilTargetElement, elementKey],
+            options: { autoExpand: { key: x.groupingNode.key, parentKeysCount: x.groupingNode.parentKeys.length } },
```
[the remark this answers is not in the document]
Yato333: These tests were already testing cases for focusing on a grouping node. See previous focused item definition for this test:
```(x) => [{ parent: { type: "category", ids: [x.category.id], modelIds: [x.model.id] }, classes: [x.testElement1.className] }],```

You can see that expectation for `autoExpand === true` for parent nodes was already there before my changes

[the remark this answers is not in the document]
Comment (grigasp) [reply packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts:1023]: Yes, we were checking that the hierarchy is correctly auto-expanded up to the target node. What I question is whether we should change the auto-expand option from `true` to the grouping node. That essentially changes the test, so if you think there's value in testing that, then I believe that should be a separate test.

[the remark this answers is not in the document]
Comment (Yato333) [reply packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts:1023]: I don't see, how this changes the test. The test was already targeting a grouping node. See focused item before the changes:
```ts
 (x) => [{ parent: { type: "category", ids: [x.category.id], modelIds: [x.model.id] }, classes: [x.testElement1.className] }],
 ```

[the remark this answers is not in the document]
Comment (grigasp) [reply packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts:1023]: OK, sorry about the confusion. I now see these tests were specifically added for testing grouping nodes' focusing. In that case, it would be great to rename the tests similarly to how I described [here](https://github.com/iTwin/viewer-components-react/pull/1010#discussion_r1714711254):
- "grouping element node under category" I think can be removed as it's covered with the test below
- "grouping element node under category and correct model" -> "grouped root element, when expanding up to its grouping node"
- "grouping element node under parent element" -> "grouped child element, when expanding up to its grouping node"
- "multiple grouping element nodes under parent element" -> "grouped child elements of different classes, when expanding up to their grouping nodes"

I guess, if we move all these tests under a `describe("when expanding up to element class grouping nodes", ...)`, then the suffix can be removed from the test names.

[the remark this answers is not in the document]
Comment (Yato333) [reply packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts:1023]: Gotcha

File: packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts [hunk 14]
```diff
@@ -1271,29 +1347,156 @@ describe("Models tree", () => {
...
-                                  }),
-                                  NodeValidators.createForInstanceNode({
-                                    instanceKeys: [x.physicalElement2],
-                                    label: /^element 2/,
-                                  }),
-                                ],
+                                autoExpand: false,
+                                children: x.physicalElementGroupingNode.groupedInstanceKeys.map((key) =>
+                                  NodeValidators.createForInstanceNode({ instanceKeys: [key] }),
+                                ),
```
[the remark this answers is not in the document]
Yato333: Fixed

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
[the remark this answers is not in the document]
Yato333: Moved these test cases under a separate `describe` and renamed accordingly

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
Yato333: Renamed the tests

grigasp (approved):

PR #1013, opened 2024-08-08
Tree widget: Fix `itwinui-react` peer dependency version

grigasp:
https://github.com/iTwin/viewer-components-react/pull/890 started using `TreeNode` attributes that were only added to `@itwin/itwinui-react` in version `3.11.0`. That requires a peer dependency bump...


Yato333 (approved):

aruniverse (approved):

grigasp:
As far as I can tell, E2E tests found 2 problems in the tree widget:
- `tw-tree-renderer` needs `{ height: "100%" }`.
- Node widths are incorrect - that's clearly seen if you make the widget narrower and scroll to the right. IMO, to solve this we need some intermediate container inside the virtualized `tw-tree-renderer` one, because we need width of all nodes to be as large as the largest node needs. Will need to wait for @saskliutas to have a look at this.

PR #1016, opened 2024-08-13
Hide map layer info tool when map layers is off or no arcgis feature info layers exist

wilmaier:
Show Map Layer Info Tool only when ArcGIS Feature Connection exists

FeatureInfoToolNotShown
FeatureInfoToolShown


packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,6 +34,26 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
       await IModelApp.tools.run(MapFeatureInfoTool.toolId);
     },
     badgeType: BadgeType.TechnicalPreview,
```
jjbeckman13, line 36: @mdastous-bentley 
Is this still considered Tech Preview? Just curious if we should remove that with this PR as well. I think was the agreement we came to that we could GA this if we could limit the exposure of the tool to only showing up when it could actually be used.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,6 +34,26 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+        }
+      }
+      const overlayLayers = vp.displayStyle.settings.mapImagery.overlayLayers.map((value) => value.toJSON());
+      for (const layer of overlayLayers) {
+        if (layer.formatId === "ArcGISFeature") {
+          return false;
+        }
+      }
+      return true;
+    }, [SyncUiEventId.ViewStateChanged]),
```
jason-crow, line 56: this event is very chatty, it is triggered from the viewport's onViewChanged
```typescript
/** Event called whenever this viewport is synchronized with its [[ViewState]].
     * @note This event is invoked *very* frequently. To avoid negatively impacting performance, consider using one of the more specific Viewport events;
     * otherwise, avoid performing excessive computations in response to this event.
     * @see [[onViewportChanged]] for receiving events at more regular intervals with more specific information about what changed.
     * @see [[onChangeView]] for an event raised specifically when a different [[ViewState]] becomes associated with the viewport.
     */
    readonly onViewChanged: BeEvent<(vp: Viewport) => void>;
```
Probably not the best thing to do given the `toJSON()` operation here, it could contribute to poor performance, 
we really would want a syncuievent associated with the `onDisplayStyleChanged` but I'm not seeing one that exists yet. Perhaps though a unique syncuievent for map layers when a layer is attached in conjunction with the `SyncUiEventId.ActiveViewportChanged` would cover this case for the isHidden to work as expected

wilmaier: Switched from using SyncUiEventId.ViewStateChanged to custom event MapFeatureInfoTool.toolSyncUiEventId to trigger isHidden for MapFeatureInfoTool

jason-crow: Wanted to write out what we discussed in a call earlier so @mdastous-bentley can weigh in when he gets back

- regarding the `MapFeatureInfoTool.toolSyncUiEventId`, we want the `SyncUiEvent` we use here to be more generalized to the event we care about, which technically is the [onMapImageryChanged](https://github.com/iTwin/itwinjs-core/blob/master/core/common/src/DisplayStyleSettings.ts#L473). Unfortunately, there isn't a `SyncUiEvent` for this yet. It probably should live next to other view related events in appui's `SyncUiEventDispatcher` you can find [here](https://github.com/iTwin/appui/blob/release/4.16.x/ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts#L25), but perhaps since this package is the general 'owner' of these settings, it wouldn't be completely inappropriate for this package to export its own `MapLayersSyncUiEventId` enum with `MapImageryChanged` that it dispatches itself from a useEffect in [MapLayersWidget](https://github.com/iTwin/viewer-components-react/blob/3cf4d181df078b62ea0484624fe2737f22243392/packages/itwin/map-layers/src/ui/widget/MapLayersWidget.tsx). Then the tool here could use that `SyncUiEvent` in addition to `SyncUiEventId.ActiveViewportChanged` to recheck whether to hide this tool
- regarding the `MapFeatureInfoTool`, I'm unsure how appropriate hardcoding `"ArcGIS"` here is. Isn't this tool generalized and not intended to be coupled to specific formatIds? If so would it be better to check against the registered formatId's? Is there an additional attribute of any registered format that makes them relevant to this tool?

eringram: This is a side note, but I was looking into this and `MapLayerImageryProvider.supportsMapFeatureInfo` is alpha right now. Considering the recent effort to no longer use internal/alpha APIs outside of the core repo, I would suggest we promote it to beta first or alongside this PR. This seems like a good use case of having access to that property

wilmaier: I need to get supportsMapFeatureInfo upgraded to beta to clear this error:

D:\a\1\s\packages\itwin\map-layers\src\ui\FeatureInfoUiItemsProvider.tsx(47,13): error @itwin/no-internal : getter "MapLayerImageryProvider.supportsMapFeatureInfo" is alpha.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,14 +37,70 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+    isHidden: new ConditionalBooleanValue(() => {
+      // Hide the MapFeatureInfoTool if the Map Layers toggle is off or no ArcGISFeature layers are active
+      let isHidden = true;
+      if (!FeatureInfoUiItemsProvider.mapLayerProviders) {
+        FeatureInfoUiItemsProvider.loadMapLayerProviders();
+      }
+      FeatureInfoUiItemsProvider.mapLayerProviders?.forEach((provider) => {
+        if (provider?.supportsMapFeatureInfo) {
+          isHidden = false;
+          return;
```
mdastous-bentley, line 49: Even though there is return here, `forEach ` will continue to iterate on the remaining imagery providers, which I think is useless.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,14 +37,70 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+    });
+  }
 
-  public constructor(private _featureInfoOpts: MapFeatureInfoOptions) {}
+  /** Gets a list of currently active map layer imagery providers. */
+  public static get mapLayerProviders(): MapLayerImageryProvider[] | undefined {
+    return this._mapLayerProviders;
+  }
+
+  public static loadMapLayerProviders() {
```
mdastous-bentley, line 80: Why do we keep in memory the list of all providers being used in the view while the only information we need is a Boolean indicating if there is are least one provider supporting 'MapFeatureInfo' ?

Also we should stop iterating as soon as we encountered a first provider fulfilling our criteria.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,14 +37,70 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+
+  public constructor(private _featureInfoOpts: MapFeatureInfoOptions) {
+    IModelApp.viewManager.onViewOpen.addOnce((vp: ScreenViewport) => {
+      const handleMapImageryChanged = (_args: Readonly<MapImagerySettings>) => {
+        FeatureInfoUiItemsProvider.loadMapLayerProviders();
+      };
+      const handleDisplayStyleChanged = () => {
+        FeatureInfoUiItemsProvider.loadMapLayerProviders();
+      };
+      vp.displayStyle.settings.onMapImageryChanged.addListener(handleMapImageryChanged);
```
mdastous-bentley, line 70: We never unregister those handlers, are we sure this is ok?

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,14 +37,70 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+      });
+      return isHidden;
+    }, [MapLayersSyncUiEventId.MapImageryChanged, SyncUiEventId.ActiveViewportChanged, SyncUiEventId.ViewStateChanged]),
   });
 
 export class FeatureInfoUiItemsProvider implements UiItemsProvider {
   // eslint-disable-line deprecation/deprecation
   public readonly id = "FeatureInfoUiItemsProvider";
   public static readonly widgetId = "map-layers:mapFeatureInfoWidget";
+  private static _mapLayerProviders: MapLayerImageryProvider[] | undefined = undefined;
```
mdastous-bentley, line 60: Can we make it work without a static?

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
```diff
@@ -32,14 +37,70 @@ export const getMapFeatureInfoToolItemDef = (): ToolItemDef =>
...
+  public static get mapLayerProviders(): MapLayerImageryProvider[] | undefined {
+    return this._mapLayerProviders;
+  }
+
+  public static loadMapLayerProviders() {
+    this._mapLayerProviders = [];
+    const vp = IModelApp.viewManager.selectedView;
+    if (vp?.viewFlags.backgroundMap) {
+      const backgroundLayers = vp.displayStyle.settings.mapImagery.backgroundLayers.map((value) => value.toJSON());
+      for (let bgLayerIndex = 0; bgLayerIndex < backgroundLayers.length; bgLayerIndex++) {
```
mdastous-bentley, line 85: Lot of duplicated code here

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
mdastous-bentley: @jjbeckman13 OK for GA.

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
wilmaier: Resolved by eliminating handlers

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
wilmaier: Resolved by removing static

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
wilmaier: Resolved by refactoring to eliminate duplicate code

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
wilmaier: Resolved by refactoring code to not maintain list of providers

packages/itwin/map-layers/src/ui/FeatureInfoUiItemsProvider.tsx
wilmaier: Will remove tech preview badges in followup PR.

PR #1017, opened 2024-08-14
Added properties for group deletion callback and deletion confirmation message

tejassutar2024:


packages/itwin/grouping-mapping-widget/src/components/Groups/GroupsVisualization.tsx
```diff
@@ -33,6 +33,8 @@ export interface GroupsVisualizationProps extends GroupsProps {
   emphasizeElements?: boolean;
   hideVisualizationToggle?: boolean;
   hideRefreshIcon?: boolean;
+  deleteConfirmationMessage?: (group: GroupMinimal) => JSX.Element;
```
MindaugasButkus, line 36: Seems a bit weird that `deleteConfirmationMessage` is a function that returns an element. I would name it `deleteConfirmationContentFactory`.

Guillar1 (approved):

PR #1022, opened 2024-08-22
[EC3Widget] Template component now handles 404s

Guillar1:


packages/itwin/ec3-widget/src/components/Templates.tsx
```diff
@@ -35,32 +35,37 @@ export const Templates = ({ onClickCreate, onClickTemplateTitle, onExportResult
...
+            displayName: x.displayName,
+            description: x.description ?? "",
+            id: x.id,
+            labels: [],
+            reportId: x._links.report.href.split("/reports/")[1],
+          };
+        });
+        setTemplates(configurations);
+      } else {
+        toaster.negative("Invalid iTwinId.");
```
MindaugasButkus, line 53: I wonder if this should also be translated, since toasts are user facing.

MindaugasButkus (approved):

packages/itwin/ec3-widget/src/components/Templates.tsx
Guillar1: To be honest, this is a useless check. I removed it.

PR #1024, opened 2024-08-26
Add volume measurement tool

saaaaaally:
Add Volume measurement tool. Tool is a special case since it is only applicable to meshes so it is off by default and can be enabled with the `showVolumeTool` flag.

volumeTool

### DIFF
--- a/change/@itwin-measure-tools-react-d4cd4a91-77b9-405d-af6d-9d910bdd494d.json
+++ b/change/@itwin-measure-tools-react-d4cd4a91-77b9-405d-af6d-9d910bdd494d.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add volume measurement tool",
+  "packageName": "@itwin/measure-tools-react",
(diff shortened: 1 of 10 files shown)


a-gagnon (changes requested):
Before I review any further, let me start by saying this:
-I don't think this code should go in viewer-components-react.
-Unlike the name suggests, this is **not** some generic volume calculation tool. It uses the viewport and the current level of detail to compute an approximate volume. This is relevant for stockpiles, but would fail badly for anything that is hollow.

bsy-nicholasw:
@saaaaaally I agree with Alex. Civil has a lot of custom tools that we do not add to this repo, it's OK and in the spirit of the original intentions of the library for the app to have it's own measurement types and leave super generic types to the base library.

We may need to know if multiple apps are requesting this type of tool, which in that case yes it makes sense to have it here to lessen dependency burdens, but it probably should be renamed to not convey that it's a generic volume measurement -or- something that gets volumes from the real elements instead of the LOD display.

Issue #980, opened 2024-07-02
[measure-tools] Allow apps to hook into feature tracking


aruniverse:
It came from our discussion on internal API usage over at https://github.com/iTwin/itwinjs-backlog/issues/782#event-9866856242

My understanding after seeing these changes and pineapple's wiki on the matter is that the role is reversed, pineapple will explicitly hook into feature tracking events from the packages they consume?

_Originally posted by @bsy-nicholasw in https://github.com/iTwin/viewer-components-react/pull/552#discussion_r1268645855_

DanRod1999:
should be able to follow how tree-widget is currently handling feature tracking

PR #1026, opened 2024-08-27
Add Feature Tracking for Measure-Tools

MikeNBentley:
Resolve [Issue #980](https://github.com/iTwin/viewer-components-react/issues/980)

Based off solutions on [tree-widget](https://www.npmjs.com/package/@itwin/tree-widget-react) package, add feature tracking to measure-tools. Now, users can pass their callback function `onFeatureUsed` to `MeasureToolsUiItemsProvider` to handle features being used. There is an usage example in the [updated README](https://github.com/iTwin/viewer-components-react/tree/mikeN/measure-tools-features-track/packages/itwin/measure-tools).

_(This information is also included in README but just put here for reference)_ The feature tracking message is now in format `feature-[feature being used]-event-[event related to the feature]`. This format should be easy enough for users to process and paste the message.

Current tracking features and respective are (will **update more** later!) (for detailed message, please again refer to [updated README](https://github.com/iTwin/viewer-components-react/tree/mikeN/measure-tools-features-track/packages/itwin/measure-tools)):
 
- Distance Measurement: Trigger, Cancel (for example we change to select mode or clear all measurements), Click on Start/End point of Distance

- Clear Measurements: Trigger

- Property Widget: When results are updated, when widget is empty

- Selection Set: While this is not essentially a feature. All measurements seem to base off a global selection set. Hence, changes to this selection set might prove useful for users, so feature tracking for Selection Set is added.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,61 +1,61 @@
-// {
-//   "editor.insertSpaces": true,
-//   "editor.tabSize": 2,
-//   "editor.trimAutoWhitespace": true,
-//   "editor.defaultFormatter": "esbenp.prettier-vscode",
-//   "editor.formatOnSave": false,
-//   "editor.codeActionsOnSave": [
-//     "source.fixAll"
-//   ],
+{
+  "editor.insertSpaces": true,
+  "editor.tabSize": 2,
+  "editor.trimAutoWhitespace": true,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": [
+    "source.fixAll"
+  ],
 
-//   "files.associations": {
-//     ".nycrc": "json",
-//   },
-//   "files.exclude": {
-//     "**/coverage": true,
-//     "**/node_modules": true,
-//     "**/lib": true,
-//     "**/build": true
-//   },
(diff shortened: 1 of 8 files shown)


bsy-nicholasw:
We already have feature tracking that all the tools call if subclasses provide `Feature` and it's called in the `postInstall` of the base tool. At some point in the past, the implementation did change a bit so the app would have to set it up by adding an event listener to `FeatureTracking.onFeature`.

At most this should just be providing a listener in the props to the UI provider that hooks up this event for the app, but it's also arguable that could also happen in `MeasureToolsFrontend.startup`.

For the extra granularity in features, they should be added to our list of features in `FeatureTracking`, but who requested adding those for every point / property widget update?

MikeNBentley:
> We already have feature tracking that all the tools call if subclasses provide `Feature` and it's called in the `postInstall` of the base tool. At some point in the past, the implementation did change a bit so the app would have to set it up by adding an event listener to `FeatureTracking.onFeature`.
> 
> At most this should just be providing a listener in the props to the UI provider that hooks up this event for the app, but it's also arguable that could also happen in `MeasureTools.startup`.
> 
> For the extra granularity in features, they should be added to our list of features in `FeatureTracking`, but who requested adding those for every point / property widget update?

PR #1029, opened 2024-08-30
Allow Users to Pass Function to Measure-Tools Feature Tracking

MikeNBentley:
Resolve [Issue #980](https://github.com/iTwin/viewer-components-react/issues/980)

- **Description:** Based off solutions on [tree-widget](https://www.npmjs.com/package/@itwin/tree-widget-react) package and idea mentioned in [this PR](https://github.com/iTwin/viewer-components-react/pull/1026).
Now, users can pass their callback function `onFeatureUsed` to `MeasureToolsUiItemsProvider` to handle features being used. There is an usage example in the [updated README](https://github.com/iTwin/viewer-components-react/tree/mikeN/feature-track-uievent/packages/itwin/measure-tools).

- **Affected File(s):** 
_(Modified)_ packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx, 
_(Modified)_ packages/itwin/measure-tools/README.md

### DIFF
--- a/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
+++ b/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
@@ -17,6 +17,7 @@ import { MeasureToolDefinitions } from "../tools/MeasureToolDefinitions";
 import type { RecursiveRequired } from "../utils/types";
 import { MeasurementPropertyWidget, MeasurementPropertyWidgetId } from "./MeasurementPropertyWidget";
 import { IModelApp } from "@itwin/core-frontend";
+import { Feature, FeatureTracking } from "../measure-tools-react";
 
 // Note: measure tools cannot pick geometry when a sheet view is active to snap to and therefore must be hidden
 //  to avoid giving the user the impression they should work
@@ -32,11 +33,15 @@ export interface MeasureToolsUiProviderOptions {
   // If we check for sheet to 3d transformation when measuring in sheets
   enableSheetMeasurement?: boolean;
   stageUsageList?: string[];
+  // Callback that is invoked when a tracked feature is used.
+  onFeatureUsed?: (feature: Feature) => void;
 }
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
-  private _props: RecursiveRequired<MeasureToolsUiProviderOptions>;
+  private _props: Omit<RecursiveRequired<MeasureToolsUiProviderOptions>, 'onFeatureUsed'> & {
+    onFeatureUsed?: (feature: Feature) => void;
+  };
 
   constructor(props?: MeasureToolsUiProviderOptions) {
     this._props = {
@@ -48,7 +53,10 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
       },
       enableSheetMeasurement: props?.enableSheetMeasurement ?? false,
       stageUsageList: props?.stageUsageList ?? [StageUsage.General],
+      onFeatureUsed: props?.onFeatureUsed,
     };
+    if (!FeatureTracking.onFeature.numberOfListeners && this._props.onFeatureUsed)
+      FeatureTracking.onFeature.addListener(this._props.onFeatureUsed);
   }
 
   public provideToolbarItems(
--- a/packages/itwin/measure-tools/README.md
+++ b/packages/itwin/measure-tools/README.md
@@ -99,3 +99,60 @@ An application can further customize UI event behavior by registering override h
 
 A concrete example of this customization is an application that has measurements organized into multiple groups. One group may be "frozen" due to some application state (state that the measure-tools library may be unaware of) and should not be cleared by the clear measurements tool.
 So the application would register a custom UI event handler that would cause those measurements to be ignored when the clear measurement tool is invoked.
+
+### Usage Tracking
+
+This package allows consumers to track the usage of specific features.
+
+This can be achieved by passing `onFeatureUsed` function to `MeasureToolsUiProvider`. The function is invoked with the information about the feature being used. The feature information is based off Feature interface (that can be imported from `"@itwin/measure-tools-react"`)
+
+```
+export interface Feature {
+  name: string;
+  guid: GuidString;
+  metaData?: Map<string, any>;
+}
+```
+
+As an example, for Measure Distance, we will have
+
+```
+{
+  name: "CRT_Tools_MeasureDistance",
+  guid: "10e474ee-9af8-4262-a505-77c9d896b065",
+}
+```
+
+### Example for Usage Tracking
+
+In this case, we create a sample [Itwin Viewer](https://www.npmjs.com/package/@itwin/web-viewer-react) and configure `MeasureToolsUiProvider`
+
+```ts
+import { Feature, MeasureToolsUiItemsProvider } from "@itwin/measure-tools-react";
+
+const App: React.FC = () => {
+  // Viewer Setup here...
+  return (
+    <div className="viewer-container">
+      <Viewer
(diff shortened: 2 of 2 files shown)


tcobbs-bentley:
I guess I'm OK with this, but I'm not entirely sure why it's even needed. Couldn't you just update the readme to instruct people to use `FeatureTracking.onFeature.addListener` directly? I'm pretty sure that's what they have been doing up until now, and it's not clear to me that this solution is a substantial improvement over that.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -32,11 +33,15 @@ export interface MeasureToolsUiProviderOptions {
...
+  // Callback that is invoked when a tracked feature is used.
+  onFeatureUsed?: (feature: Feature) => void;
 }
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
-  private _props: RecursiveRequired<MeasureToolsUiProviderOptions>;
+  private _props: Omit<RecursiveRequired<MeasureToolsUiProviderOptions>, 'onFeatureUsed'> & {
+    onFeatureUsed?: (feature: Feature) => void;
+  };
```
ben-polinsky, line 44: This is getting a bit confusing, especially since it doesn't seem like `this._props.onFeatureUsed` is used outside of the constructor: why should it be added to `this._props` at all? One idea is to each of the `_props` members into its own private member. Then we wouldn't have this unnecessary coupling between internal state and arguments.

ben-polinsky:
> I guess I'm OK with this, but I'm not entirely sure why it's even needed. Couldn't you just update the readme to instruct people to use `FeatureTracking.onFeature.addListener` directly? I'm pretty sure that's what they have been doing up until now, and it's not clear to me that this solution is a substantial improvement over that.

I see that, but on the other hand it's sort of reaching into the library's internals? I know FeatureTracking is currently public, but if we do want to change the implementation it's better to have consumers use this. Consumers shouldn't be able to fire off feature tracking events, either, so we really shouldn't be exposing the BeEvent itself.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
MikeNBentley: Thank you, onFeatureUsed is indeed used in constructor only. I will remove `& {
    onFeatureUsed?: (feature: Feature) => void;
  };` part and directly use `props?.onFeatureUsed` instead. This is also reasonable as it does not make a lot of sense to initialize onFeatureUsed with any default value if it is a member of `this._props`

```
this._props = {
      itemPriority: props?.itemPriority ?? 20,
      groupPriority: props?.groupPriority ?? 10,
      widgetPlacement: {
        location: props?.widgetPlacement?.location ?? StagePanelLocation.Right,
        section: props?.widgetPlacement?.section ?? StagePanelSection.Start,
      },
      enableSheetMeasurement: props?.enableSheetMeasurement ?? false,
      stageUsageList: props?.stageUsageList ?? [StageUsage.General],
      onFeatureUsed: props?.onFeatureUsed, // will also remove this line
    };
```
About the idea to each of the `_props` members into its own private member, I think we can do that in another PR where we can replace `_props` usage with those private properties. But does that mean we do not need `RecursiveRequired` introduced in this [commit](https://github.com/iTwin/viewer-components-react/commit/5dde769047e695c11eb1e91e930de28a267ea7ab) (to make sure all `_props` members initialized with at least some value) anymore?

bsy-nicholasw:
@aruniverse Is there a push for packages to adopt this pattern? It doesn't really add much other than better discoverability, and beginning to move away from package statics. I just want to understand what/who is driving this.

@ben-polinsky IMO the same can then be said for every public event on any public API, but it's a common pattern to expose it in the style of a C# event.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -48,7 +53,10 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
       },
       enableSheetMeasurement: props?.enableSheetMeasurement ?? false,
       stageUsageList: props?.stageUsageList ?? [StageUsage.General],
+      onFeatureUsed: props?.onFeatureUsed,
     };
+    if (!FeatureTracking.onFeature.numberOfListeners && this._props.onFeatureUsed)
```
bsy-nicholasw, line 58: I'd nix the condition checking the number of listenes.

@aruniverse Do UI providers ever get cleaned up? Or only ever created once per app lifetime?

MikeNBentley: The reason for `!FeatureTracking.onFeature.numberOfListeners` is as I test on a sample viewer this listener is added twice. I guess the UI provider might get cleaned up and we have not removed this listener before re-creating the UI provider (but I am not really clear on when it gets cleaned up or why it might be created twice).

tcobbs-bentley: With FeatureTracking.onFeature being public, I don't think it's appropriate to assume that nobody else is listening for that event. I doesn't seem acceptable to have this not work because the event is already being listened to by somebody else.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
ben-polinsky: Sure, I was just thinking that perhaps someone would come along and do something similar for another arg that wasn't used outside of the constructor, but whichever you feel more comfortable doing. Just an observation. This repo/file isn't my code so don't want to impose.

ben-polinsky:
> IMO the same can then be said for every public event on any public API, but it's a common pattern to expose it in the style of a C# event.

Perhaps, but doesn't make it great for situations where there's really no reason for consumers to emit. Just my two cents, I get it's already public, so maybe more trouble than it is worth.

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
bsy-nicholasw: Looks like the interface has a cleanup mechanism: https://www.itwinjs.org/reference/appui-react/uiprovider/uiitemsprovider/onunregister/

TBH something like this belongs better in the `MeasureTools.startup` which the app has to call anyways. It also has an associated terminate, which should be called alongside the IModelApp startup/shutdown by the app.

MikeNBentley: Sure, I can implement it as a startup option instead. However, its usage will be different from [tree-widget](https://www.npmjs.com/package/@itwin/tree-widget-react) package (where you can provide onFeatureUsed callback function to UI Provider). If you are fine with that, I will do it right away. Thank you!

aruniverse:
> @aruniverse Is there a push for packages to adopt this pattern? It doesn't really add much other than better discoverability, and beginning to move away from package statics. I just want to understand what/who is driving this.

There is no immediate push for packages to adopt this pattern. If packages need to provide feature tracking this is probably the most preferred way to do it so apps can inject that themselves, instead of every package having to create their own static feature tracker and event emitter. This PR mainly stems from removing the internal Feature tracking apis from core, and making it easier in case people need a replacement. I have no guarantees Apps are even hooking into the existing feature tracking.

aruniverse:
> Perhaps, but doesn't make it great for situations where there's really no reason for consumers to emit. Just my two cents, I get it's already public, so maybe more trouble than it is worth.

Although it's marked "public" but technically its a `0.x` pkg so it could happen

PR #1054, opened 2024-09-19
[G&M] Preserving group visualization for group properties pane

tejassutar2024:
As per current behavior, when groups visualization is turned on for all groups, and when individual group is selected, the visualization goes away. With this fix, visualization will remain in tact when user enters group properties pane.

**When groups visualization is turned on:**

image

**When individual group properties pane is entered:**

image

### DIFF
--- a/packages/itwin/grouping-mapping-widget/src/components/context/GroupHilitedElementsContext.ts
+++ b/packages/itwin/grouping-mapping-widget/src/components/context/GroupHilitedElementsContext.ts
@@ -18,6 +18,12 @@ export interface OverlappedElementsMetadata {
   groupElementsInfo: Map<string, number>;
   overlappedElementGroupPairs: OverlappedElementGroupPairs[];
 }
+
+export interface GroupVisualizationColor {
+  groupId: string;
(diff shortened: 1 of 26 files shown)


packages/itwin/grouping-mapping-widget/src/components/context/GroupHilitedElementsContext.ts
```diff
@@ -50,6 +58,8 @@ export const GroupHilitedElementsContext = React.createContext<GroupHilitedEleme
   setOverlappedElementsMetadata: () => {},
   setIsOverlappedColored: () => {},
   setIsVisualizationsEnabled: () => {},
+  setGroupColors: () => {},
+  groupColors: [],
```
Guillar1, line 62: A `Map` would make this quicker and cleaner.
` backgroundColor={groupColors.find((c) => c.groupId === props.group.id)?.color ?? ""}`
vs
`backgroundColor={groupColors.get(props.group.id) ?? ""}`

Guillar1 (approved):

PR #1065, opened 2024-10-14
Fix measurements widget looping from setState in useEffect

wilmaier:
Switching frontstages while measurement widget is open causes window to freeze due to infinite recursion loop triggered by setState inside of useEffect.  

The simple solution is to switch to using useLayoutEffect which only gets called once.


wilmaier:
This fix is urgently needed as it is blocking a pineapple release build.

a-gagnon (approved):

PR #1070, opened 2024-10-17
Fix z-index of map layers settings popup.

wilmaier:
**Before:**
image

**After:**
image

**Modal dialog still shown above settings popup**
image


packages/itwin/map-layers/src/ui/widget/MapLayerSettingsPopupButton.scss
```diff
@@ -7,7 +7,7 @@
 $hovered-icon-color: $buic-foreground-primary;
 
 .maplayers-settings-popup {
-  z-index: 900; // Used to be higher, but we want this be less than modal dialog z-index
+  z-index: 6001; /* one more than floating widget */
```
aruniverse, line 10: what happens when they change the z index of floating widget?
is there no scss var for floating widget z indexes that you can use then add 1 to that?

aruniverse (commented):
i think we need to really re-think the popup features in map layers, its really strange ux compared to everything else where we use modals

packages/itwin/map-layers/src/ui/widget/MapLayerSettingsPopupButton.scss
jason-crow: im not seeing any z-index variables that itwinui is using, they seem to use 0, 1, 99, 999, 9999 consistently though

wilmaier: I guess I can use:      z-index: uicore-get-z-index(dragged-widget) + 1;

jason-crow: that mixin is ancient, and I was under the impression we were moving away from using mixins. Perhaps we need to check with Gerardas/Mayank on the best practise here

jason-crow: @jzbuchalski do you have any thoughts here?

jzbuchalski: My $0.02....

For the new Saved Views widget, I went with a new approach for displaying settings.   Some settings, for example, need a description (certainly when it comes to the complexity of saved views).  Instead of rendering yet another "..." menu, I decided to overlay a sliding panel (btw... a sliding component is under development by the itwinui team).   After I implemented the solution (below), passed it by a few of us, the consensus was a better solution, and we should consider using it in other widgets.  Mantvydas specifically pointed out map layers :-)

The solutions below:

- Avoids z-index issues and yet another popup/dialog.
- Scales better (easier to add additional settings).
- Provide extra detail for complex settings (some users have complainted Pineapple has poor documentation).

image

image

jason-crow: thanks @jzbuchalski that seems like a nice enhancement down the road (fyi @mdastous-bentley)

For now, since this z-index issue is really a workaround for the deprecated Popup component from appui that map-layers is still using, @wilmaier, my suggestion is to just keep the 6001 and add a TODO comment explaining that this is a temporary workaround for a deprecated component

The next step would be to migrate away from the deprecated component, but lets just get this fixed first then coordinate on that

@aruniverse  agreeable?

mdastous-bentley: Guys, we have no plan for re-designing the map-layers widget as of now... we are working on a new "unified" widget that will cover map-layers but also features layers, reality data, etc.

fyi @MarcBedard8

mdastous-bentley (approved):

PR #1072, opened 2024-10-23
[measure-tools] Prevent drawing measure tools in non-sheet views when sheet measurements are enabled

Maxime-Brassard:
This makes it so that, when sheet measurements are enabled, they can only be used in sheet views.
It does not affect measurements when sheet measurements are disabled.
Related to https://dev.azure.com/bentleycs/Civil-iTwin/_workitems/edit/1546695


a-gagnon:
I don't think adding restrictive logic on the tool to tailor it to our specific workflow is a good idea.

If we're the ones exposing this tool in a different context as portrayed in the linked work item, it might be best to subclass the tools and override the isValidLocation instead.

PR #1074, opened 2024-10-28
Add support for additional Toolbar items to MeasureToolsUiItemsProvider

saaaaaally:
Added addAdditionalToolbarItems to MeasureToolsUiItemsProvider so that applications can pass in additional measure tools to be displayed at the end of the measure tools list.

### DIFF
--- a/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
+++ b/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
@@ -37,8 +37,7 @@ export interface MeasureToolsUiProviderOptions {
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
-  private _props: RecursiveRequired<MeasureToolsUiProviderOptions>;
-  private _additionalToolbarItems: ToolItemDef[] = [];
+  private _props: Omit<RecursiveRequired<MeasureToolsUiProviderOptions>, 'additionalToolbarItems'> & { additionalToolbarItems?: ToolItemDef[] };
 
   constructor(props?: MeasureToolsUiProviderOptions) {
     this._props = {
@@ -50,9 +49,8 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
       },
       enableSheetMeasurement: props?.enableSheetMeasurement ?? false,
       stageUsageList: props?.stageUsageList ?? [StageUsage.General],
-      additionalToolbarItems: []
+      additionalToolbarItems: props?.additionalToolbarItems
     };
-    this._additionalToolbarItems = props?.additionalToolbarItems ?? [];
   }
 
   public provideToolbarItems(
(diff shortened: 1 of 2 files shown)


wilmaier (approved):

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -51,6 +52,10 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
     };
   }
 
+  public addAdditionalToolbarItems(toolbarItems: ToolItemDef[]): void {
```
aruniverse, line 55: why wouldnt you just add additional tool bar items to MeasureToolsUiProviderOptions?

bsy-nicholasw: Yeah, I'd put it as an option to the ctor. "Add" also here is really a replace, which I think consumers would not expect they'd be overwriting someone else's addition.

saaaaaally: Moved additionalToolbarItems to MeasureToolsUiProviderOptions

bsy-nicholasw (approved):

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -32,11 +32,13 @@ export interface MeasureToolsUiProviderOptions {
   // If we check for sheet to 3d transformation when measuring in sheets
   enableSheetMeasurement?: boolean;
   stageUsageList?: string[];
+  additionalToolbarItems?: ToolItemDef[];
 }
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
   private _props: RecursiveRequired<MeasureToolsUiProviderOptions>;
+  private _additionalToolbarItems: ToolItemDef[] = [];
```
aruniverse, line 41: do we need this? we're gonna have 2 "places of truth"
_props would have additionalToolbarItems itself, but you set that to always be an empty array

saaaaaally: Moved additionalToolbarItems into _props only

packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -32,11 +32,12 @@ export interface MeasureToolsUiProviderOptions {
   // If we check for sheet to 3d transformation when measuring in sheets
   enableSheetMeasurement?: boolean;
   stageUsageList?: string[];
+  additionalToolbarItems?: ToolItemDef[];
 }
 
 export class MeasureToolsUiItemsProvider implements UiItemsProvider {
   public readonly id = "MeasureToolsUiItemsProvider";
-  private _props: RecursiveRequired<MeasureToolsUiProviderOptions>;
+  private _props: Omit<RecursiveRequired<MeasureToolsUiProviderOptions>, 'additionalToolbarItems'> & { additionalToolbarItems?: ToolItemDef[] };
```
aruniverse, line 40: i was thinking `<RecursiveRequired<Omit<T, x>,> & { x: X },
but this should work too

aruniverse (approved):

PR #1075, opened 2024-10-28
[measure-tools] Show tooltip when using sheet measurement on wrong drawing type

Maxime-Brassard:
Added tooltip to convey to user which drawings can take the measurement he's trying to use when using sheet measurements.
TooltipSheetMeas

### DIFF
--- a/packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
+++ b/packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
@@ -192,9 +192,6 @@ export namespace SheetMeasurementsHelper {
     if (allowedDrawingTypesList.length < 1) {
       return "";
     }
-    if (allowedDrawingTypesList.length === 1) {
-      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.OneMeasurementInvalid", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[0])});
-    }
     if (allowedDrawingTypesList.length > 1){
       let result = IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MoreMeasurementsInvalidHead", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[0])});
       for (let i = 1; i < allowedDrawingTypesList.length - 1; i++) {
@@ -203,6 +200,6 @@ export namespace SheetMeasurementsHelper {
       result = result + IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MoreMeasurementsInvalidLast", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[allowedDrawingTypesList.length - 1])});
       return result;
     }
-    return "";
+    return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.OneMeasurementInvalid", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[0])});
   }
 }
--- a/packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
+++ b/packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
@@ -174,11 +174,11 @@ export namespace SheetMeasurementsHelper {
 
   function getNameFromDrawingType(type: SheetMeasurementsHelper.DrawingType): string  {
     if (type === SheetMeasurementsHelper.DrawingType.CrossSection) {
-      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MeasurementNames.CrossSection");
+      return IModelApp.localization.getLocalizedString("MeasureTools:SheetMeasurementTooltip.MeasurementNames.CrossSection");
     } else if (type === SheetMeasurementsHelper.DrawingType.Plan) {
-      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MeasurementNames.Plan");
+      return IModelApp.localization.getLocalizedString("MeasureTools:SheetMeasurementTooltip.MeasurementNames.Plan");
     } else if (type === SheetMeasurementsHelper.DrawingType.Profile) {
-      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MeasurementNames.Profile");
+      return IModelApp.localization.getLocalizedString("MeasureTools:SheetMeasurementTooltip.MeasurementNames.Profile");
     } else {
       return "";
     }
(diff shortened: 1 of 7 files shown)


packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
```diff
@@ -152,12 +155,20 @@ MeasureLocationToolModel
...
       return false;
 
     return true;
   }
 
+  public override async getToolTip(hit: HitDetail): Promise<HTMLElement | string> {
+    if (!this._enableSheetMeasurements || SheetMeasurementsHelper.checkIfAllowedDrawingType(hit.viewport, hit.hitPoint, MeasureLocationTool._allowedDrawingTypes)) {
+      return super.getToolTip(hit);
+    } else {
+      return SheetMeasurementsHelper.getToolTipText(MeasureLocationTool._allowedDrawingTypes);
```
bsy-nicholasw, line 168: `SheetMeasurementsHelper.getToolTipText` should be named to reflect it's specifically getting an error text. Otherwise, to reduce the boilerplate code these tool getToolTips can become:

```
if (!this._enableSheetMeasurements)
  return super.getToolTip(hit);

return SheetMeasurementsHelper.getToolTipText(hit, Tool._allowedDrawingTypes, async () => super.getToolTip(hit));
```

e.g. The helper checks if it's allowed, calls the function that is passed in, otherwise returns the error text.

Also, this feels like it could be in the base measurement tool class. And the list of allowed drawing types either like how subclass tools override static properties, or it be an instance-based method that is overridden that returns the array.

Maxime-Brassard: @bsy-nicholasw I think I've implemented your suggestion, please verify when you have time.

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -159,16 +159,50 @@ export namespace SheetMeasurementsHelper {
...
+    } else if (type === SheetMeasurementsHelper.DrawingType.Plan) {
+      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MeasurementNames.Plan");
+    } else if (type === SheetMeasurementsHelper.DrawingType.Profile) {
+      return IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MeasurementNames.Profile");
+    } else {
+      return "";
+    }
+  }
+
+  export async function getSheetErrorToolTipText(hit: HitDetail, allowedDrawingTypesList: SheetMeasurementsHelper.DrawingType[], defaultToolTip:(hit: HitDetail) => Promise<HTMLElement | string>): Promise<string | HTMLElement> {
```
bsy-nicholasw, line 187: The suggestion for changing the name was an "or" with the suggestion of bumping the usage logic to the base measurement tool. Since you did the latter (and it all looks good), I wouldn't have changed the helper name, since if the types are allowed, it's not returning an error. It's a nitpick. I'd just keep the name "getSheetToolTipText" or "getSheetToolTipTextOrError".

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -159,16 +159,50 @@ export namespace SheetMeasurementsHelper {
...
+    if (SheetMeasurementsHelper.checkIfAllowedDrawingType(hit.viewport, hit.hitPoint, allowedDrawingTypesList)) {
+      return defaultToolTip(hit);
+    }
+
+    if (allowedDrawingTypesList.length < 1) {
+      return "";
```
bsy-nicholasw, line 193: Since a zero allowed drawing type list won't be allowed, it feels like then we need an error message for this case. I suppose realistically, it's a case that shouldn't happen to the end user (if no allowed types, then the sheet measurement would be disabled in the tool?). So this may actually be more of an error message to inform the developer that they did something wrong.

"Hey Jerk, you didn't add allow any drawing types" 😃

packages/itwin/measure-tools/src/tools/MeasureLocationTool.ts
bsy-nicholasw: Yeah, the tool changes look good. I made some other comments about the helper

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -159,16 +159,50 @@ export namespace SheetMeasurementsHelper {
...
+    }
+    if (allowedDrawingTypesList.length > 1){
+      let result = IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MoreMeasurementsInvalidHead", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[0])});
+      for (let i = 1; i < allowedDrawingTypesList.length - 1; i++) {
+        result = result + IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MoreMeasurementsInvalidMiddle", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[i])});
+      }
+      result = result + IModelApp.localization.getLocalizedString("CivilReviewTools:SheetMeasurementTooltip.MoreMeasurementsInvalidLast", { drawingName: getNameFromDrawingType(allowedDrawingTypesList[allowedDrawingTypesList.length - 1])});
+      return result;
+    }
+    return "";
```
bsy-nicholasw, line 206: For code clarity, I'd have the final return be your `length===1` case, since you had already checked if zero or if greater than one.

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
Maxime-Brassard: So we're thinking more like throwing an error than a message here?

bsy-nicholasw (approved):

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
```diff
@@ -159,16 +159,47 @@ export namespace SheetMeasurementsHelper {
...
-        if (SheetMeasurementsHelper.checkIfInDrawing(ev.point, drawing.origin, drawing.extents)) {
+        if (SheetMeasurementsHelper.checkIfInDrawing(point, drawing.origin, drawing.extents)) {
           return true;
         }
       }
     }
     return false;
   }
+
+  function getNameFromDrawingType(type: SheetMeasurementsHelper.DrawingType): string  {
```
aruniverse, line 175: you should NOT be using strings from other namespaces
if you need these strings, copy them into this pkg

aruniverse (changes requested):

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
Maxime-Brassard: Oops, my bad, I originally made this change in the other package and I forgot this change! It should be good now

aruniverse (approved):

packages/itwin/measure-tools/src/api/SheetMeasurementHelper.ts
bsy-nicholasw: Arun for the save...I completely glossed over CivilReviewTools...

PR #1079, opened 2024-10-30
[measure-tools] Added a callback that allows filtering viewports where the measurements can be drawn

Maxime-Brassard:


bsy-nicholasw:
Hmm, this feels invasive and brittle, and a retread of how view targets work.

For example, a measurement's view target is checked by the [manager ](https://github.com/iTwin/viewer-components-react/blob/cfe9b0401f13948986e0602b729218cedf7604bb/packages/itwin/measure-tools/src/api/MeasurementManager.ts#L401 )before invoking decorate on each.

It sounds like a variation on that theme would be for this case, where the tool model also has acceptable viewTargets that the tool then checks during the `isValidLocation` check. Or that the tool model has a new method that the tool checks, since forcing tool models to declare view targets sounds like a breaking change.

bsy-nicholasw (approved):

Issue #1012, opened 2024-08-08
Property widget: Need access to data provider in action button renderers


grigasp:
In some cases consumers may need a data provider to render their custom action button. 

At the moment that is not possible, because consumer has no access to the data provider, so it can't be captured by the `ActionButtonRenderer`.

We should consider exposing the data provider and, possibly, the field that matches the property record, through the action button renderer props, similar to what we do with context menu (see `ContextMenuItemProps`). At the moment property widget simply exposes the `actionButtonRenderers` prop from the virtualized property grid component. We don't want a change there. Instead, we should override the prop with a new type, where the renderer receives the extended props.

grigasp:
@morganadrales, I have you listed as the requestor for this. The change released with [`@itwin/property-grid-react@1.14.0`](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/property-grid/CHANGELOG.md#1140).

PR #1082, opened 2024-11-05
[Property-grid]: Add access to data provider in action button renderers

MartynasStrazdas:
closes https://github.com/iTwin/viewer-components-react/issues/1012


saskliutas (changes requested):

packages/itwin/property-grid/src/test/components/PropertyGridContent.test.tsx
```diff
@@ -108,6 +108,23 @@ describe("<PropertyGridContent />", () => {
...
   });
 
+  it("renders with extended props action buttons", async () => {
+    const imodel = {} as IModelConnection;
+    const spy = sinon.spy();
+
+    const renderTestActionButton = (dataProvider: IPresentationPropertyDataProvider) => {
+      spy(dataProvider);
+      return <div>Test action button</div>;
+    };
```
grigasp, line 118: seems like `sinon.fake` would be better suited for this

packages/itwin/property-grid/src/hooks/UseActionButtons.ts
```diff
@@ -0,0 +1,44 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ReactNode } from "react";
+import { useMemo } from "react";
+
+import type { IPresentationPropertyDataProvider } from "@itwin/presentation-components";
+import type { ActionButtonRendererProps } from "@itwin/components-react";
+
+interface PresentationActionButtonRendererProps extends ActionButtonRendererProps {
```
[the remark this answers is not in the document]
MartynasStrazdas: changed

packages/itwin/property-grid/src/test/components/PropertyGridContent.test.tsx
```diff
@@ -108,6 +108,25 @@ describe("<PropertyGridContent />", () => {
...
+    const imodel = {} as IModelConnection;
+    const fake = sinon.fake();
+
+    const renderTestActionButton = (dataProvider: IPresentationPropertyDataProvider) => {
+      fake(dataProvider);
+      return <div>Test action button</div>;
+    };
+
+    const { getAllByText } = renderWithContext(
+      <PropertyGridContent dataProvider={provider} imodel={imodel} actionButtonRenderers={[({ dataProvider }) => renderTestActionButton(dataProvider)]} />,
```
saskliutas, line 121: Why not `actionButtonRenderers={[fake]}`?

MartynasStrazdas: Could do that, though we have no test that check if actionButtons adds element to the DOM. Thought it might be useful to have one.

saskliutas: In that case you could do:

```tsx
const fake = sinon.fake().returns(<div>Test action button</div>);
// or if fake does not support returns
const stub = sinon.stub().returns(<div>Test action button</div>);
```

grigasp: yeah, I think it's either 
```ts
const fake = sinon.fake(() => {
  return <div>Test action button</div>;
});
```

or
```ts
const stub = sinon.stub().returns(<div>Test action button</div>);
```

I also think this stub/fake should be passed straight as an action button renderer, rather than called from one, e.g. `actionButtonRenderers={[fake]}`. Then you can check for correct call with `expect(fake).to.be.calledWith(sinon.match((arg) => arg.dataProvider === provider))`

MartynasStrazdas: adjusted.

saskliutas (approved):

Issue #1076, opened 2024-10-29
Tree widget: Finding Models tree node paths by target instance keys is too slow


grigasp:
Num of elements (instance keys) | Time for initial load of tree
-- | --
4883 | 39.18 seconds
10000 | 85 seconds

It's been already determined that the slow part is paths' lookup. Need to investigate further.

grigasp:
Released with version [3.0.2](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/tree-widget/CHANGELOG.md#302)

PR #1084, opened 2024-11-07
Increase filtering speed of filtering paths creation

JonasDov:
Related to: https://github.com/iTwin/viewer-components-react/issues/1076.
These changes were tested in https://github.com/iTwin/presentation test app. 

2 implementations were tested (that performed worse than final implementation) with these results:
Number of target items | 1st implementation | 2nd implementation
-- | -- | --
4883 | 560 ms | 510 ms
10000 | 1.2 seconds | 1 second
87246 | 17.7 seconds | 12.2 seconds

Changes from this suggestion were added: https://github.com/iTwin/presentation/pull/770#discussion_r1833753477. 
Now the changes create perform like so:
Number of target items | Original implementation (time for initial load of tree) | New implementation (time for initial load of tree)
-- | -- | --
4883 | 26.18 seconds | 450 ms
10000 | 1.12 seconds | 900 ms
87246 | 18.2 seconds | 7.5 seconds

These how all of the implementations work:
- Original implementation: paths are created during query execution. It  creates a `json_array` filled with `json_object`. This later has to be parsed using `JSON.parse()` (query returns this object as a string).
- 1st implementation: paths are created after query execution. Query returns all nodes (that are needed to create paths) with relevant information.
- 2nd implementation: Paths are created during query execution just like in original implementation. During query execution Models, Categories and GeometricElements tables need to be used to create the paths. Original implementation used `JOIN` to get relevant information from the tables. This implementation changed `JOIN` to `LEFT JOIN`, and adjusted the `ON` condition, to only join those tables when it is necessary.
- New implementation: Paths are created during query execution just like in 2nd implementation. This implementation does not use `json_array` filled with `json_object`, instead it saves paths into a string. This string is later parsed to create the paths. Additionally, query is executed multiple times, when there are more than 5k elements. All target elements are split into groups of 5k elements (new query is executed for each group), this reduces the time by about 1 second (tried splitting into groups of: 2k, 5k, 10k, 15k, 20k, 25k, 30k, 45k, and 5k performance was the best).


packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -577,8 +577,8 @@ describe("Models tree", () => {
           return { rootSubject, model1, model2, category1, category2, element11, element12, element21, element22 };
         },
         (x) => [
-          [x.rootSubject, x.model1, x.category1, x.element11],
-          [x.rootSubject, x.model2, x.category2, x.element22],
+          [x.rootSubject, { ...x.model1, className: "BisCore.GeometricModel3d" }, x.category1, { ...x.element11, className: "BisCore.GeometricElement3d" }],
```
grigasp, line 580: what do you think about adding a couple of functions like this:
```ts
const modelKey = (source: InstanceKey) => ({ className: "BisCore.GeometricModel3d", id: source.id });
const elementKey = (source: InstanceKey) => ({ className: "BisCore.GeometricElement3d", id: source.id });
```

and then using them when definining the paths?

JonasDov: Addded two functions:
```ts
const adjustedModelKey = (source: InstanceKey) => ({ className: "BisCore.GeometricModel3d", id: source.id });
const adjustedElementKey = (source: InstanceKey) => ({ className: "BisCore.GeometricElement3d", id: source.id });
```

grigasp (approved):

PR #1087, opened 2024-11-08
[measurement-tools] (RE) expose a callback that allows the application to determine whether to show or hide the measurement tools.

raplemie:
The measuring tools have supported measurement on sheet views since PR [837](https://github.com/iTwin/viewer-components-react/pull/837). As a result, the changes made in PR [336](https://github.com/iTwin/viewer-components-react/pull/336), which disabled the measurement tools when a sheet view is active, need to be revised accordingly.

This PR introduces a callback that allows the application to determine whether to show or hide the measurement tools. Additionally, the tools now listen to the `SyncUiEventId.ActiveViewportChanged` event to update their visibility when the user changes the active view in a multi-view frontstage.

**This is an attempt at doing @thanhbinh-d #1069 to validate if the issue is with org membership.**

### DIFF
--- a/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
+++ b/packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
@@ -19,9 +19,6 @@ import { MeasurementPropertyWidget, MeasurementPropertyWidgetId } from "./Measur
 import { IModelApp } from "@itwin/core-frontend";
 import type { ScreenViewport } from "@itwin/core-frontend";
 
-// Note: measure tools cannot pick geometry when a sheet view is active to snap to and therefore must be hidden
-//  to avoid giving the user the impression they should work
-const isSheetViewActive = () => !!IModelApp.viewManager.selectedView?.view?.isSheetView();
 
 export interface MeasureToolsUiProviderOptions {
   itemPriority?: number;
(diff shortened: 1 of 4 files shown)


packages/itwin/measure-tools/src/ui-2.0/MeasureToolsUiProvider.tsx
```diff
@@ -117,8 +118,9 @@ export class MeasureToolsUiItemsProvider implements UiItemsProvider {
             MeasureToolDefinitions.clearMeasurementsToolCommand,
             {
               isHidden: new ConditionalBooleanValue(
-                () => isSheetViewActive() || !MeasurementUIEvents.isClearMeasurementButtonVisible,
+                () => !MeasurementUIEvents.isClearMeasurementButtonVisible,
```
a-gagnon, line 121: Is this something that can wait until appui 5.x is released?
With the refactor that deprecates `provideToolbarItems` in favor of `getToolbarItems`, the consuming application will have the freedom to add or override the `isHidden` at will, making this change useless.
If it's needed right now, I suppose we can squeeze it in and remove it later on.

raplemie: Well, let's comment on the original PR, I was trying to validate if this was an issue with organisation. I'm closing this PR now.

raplemie:
#1069 is the real PR for this.

PR #1091, opened 2024-11-14
[grouping-mapping-widget] Fixed reset button to clear selected element and simpleSelectionQuery

AniruddhaDhavalikar:
grouping-mapping-widget: fixed reset button to clear selected element and simpleSelectionQuery.


packages/itwin/grouping-mapping-widget/src/components/customUI/GroupQueryBuilderCustomUI.tsx
```diff
@@ -74,7 +74,11 @@ export const GroupQueryBuilderCustomUI = ({ updateQuery, isUpdating, resetView }
   const onClickResetButton = async () => {
```
MindaugasButkus, line 74: This should be changed to be `useCallback` with dependencies on `iModelConnection` and `currentPropertyList`.

Guillar1: I can somewhat see that `resetView` being wrapped in a `useMutation` with the `onError` being used to console log.

Guillar1: Please attempt to do this. It will make the code a bit cleaner. 
```
const { mutate: resetMutation} = useMutation(resetView, {
  onError: (error) => console.error(error),
});
```
Is what I had in mind. Let me know if it works.

AniruddhaDhavalikar: I've wrapped `resetView` in `useMutation` as suggested, using `onError` to log any errors. The `onClickResetButton` function now calls `resetMutation.mutate()` directly, simplifying the error handling and making the code cleaner.

Guillar1: Callback is no longer `async`.

AniruddhaDhavalikar: I've removed async from onClickResetButton since useMutation handles the async behavior, and the callback is no longer asynchronous. Thanks for catching that!

Guillar1 (approved):

MindaugasButkus (approved):

PR #1094, opened 2024-11-15
[tree-widget]: Reduce tree-widget UI blocking

saskliutas:
Added caching to class checker to speed up visibility status lookup.
Reduce main thread blocking when checking node visibility statuses.

### DIFF
--- a/packages/itwin/tree-widget/src/components/trees/common/UseHierarchyVisibility.ts
+++ b/packages/itwin/tree-widget/src/components/trees/common/UseHierarchyVisibility.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { useEffect, useRef, useState } from "react";
-import { defer, distinct, from, lastValueFrom, mergeMap, Subject, takeUntil } from "rxjs";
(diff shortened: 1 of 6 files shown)


grigasp (approved):

packages/itwin/tree-widget/src/components/trees/common/UseHierarchyVisibility.ts
```diff
@@ -79,15 +79,21 @@ export function useHierarchyVisibility({ visibilityHandlerFactory }: UseHierarch
...
+        throttleTime(100, undefined, { leading: true, trailing: true }),
       )
       .subscribe({
-        next: ({ node, status }) => {
-          visibilityStatusMap.current.set(node.id, {
-            node,
-            status,
-            needsRefresh: false,
-          });
+        next: () => {
```
MartynasStrazdas, line 96: Do we still need next here?

saskliutas: This observable never completes and waits for new nodes that needs visibility determined (when those nodes are rendered) so complete is never called.

MartynasStrazdas (approved):

PR #1097, opened 2024-11-22
[measure-tools] update to itwinui v3

aruniverse:
- update to itwinui v3
	- only being used for 1 icon
- need to update appui peer to minimum of 4.10 to use itwinuiv3 in appui so updated to the latest
	- this allows us to drop appui-layout entirely
- updated core, presentation, appui deps to latest minor to make sure no apis were deprecated in devDeps
- updated to support react 18


packages/itwin/measure-tools/package.json
```diff
@@ -44,42 +44,40 @@
...
-    "@itwin/components-react": "^4.4.0",
-    "@itwin/core-bentley": "^4.7.0",
-    "@itwin/core-common": "^4.7.0",
-    "@itwin/core-frontend": "^4.7.0",
-    "@itwin/core-geometry": "^4.7.0",
-    "@itwin/core-react": "^4.4.0",
+    "@itwin/appui-abstract": "^4.10.0",
+    "@itwin/appui-react": "^4.17.0",
+    "@itwin/components-react": "^4.17.0",
+    "@itwin/core-bentley": "^4.10.0",
```
raplemie, line 50: @aruniverse  Since these are peerDependencies, shouldnt that cause this to be a major change ? Basically it means that if we want to use this new version, we need to bump our dependencies to consume it and make sure that we are to the latest dependencies, which if no code change is required, means that there is no real reasons to change the iTwin.js dependencies. (on top of that, I think it shouldnt be hidden under the iTwinUI V3 changes)

It would at least make this 1.0.0, and make this package more easily manageable with versions, rather than 0.x version which is ignored by package management.

Maxime-Brassard:
I'm unsure here why there was a need to change the peerDeps at all. Since there's no new api that are being used. 
Are all of these package updates (peerDeps and devDeps) really needed in that case?
I would not want every consuming app for this package to have to update all of this right away.

packages/itwin/measure-tools/package.json
Maxime-Brassard: Per my other comment, if there are no new APIs being used, the peer deps should probably just remain the same.

aruniverse: Thats very fair, while these are "breaking" since this pkg is 0.x its fair imo.
I do not want to bump this to 1.0 until we have clear idea of ownership of this pkg, also with both appui and core majors around the corner we'd have to do a 2.0 and 3.0 release shortly back to back. 

I reverted bumping the core peer deps. The AppUI you need to bump to a minimum of 4.10 to pull in iTwinUI@3. And at that point if youre bumping it, might as well go to the latest. This also helps us verify we're not using any deprecated APIs that will be removed in AppUI@5.

DevDeps have been updated to latest minors and that should be fine

aruniverse:
> I'm unsure here why there was a need to change the peerDeps at all. Since there's no new api that are being used. Are all of these package updates (peerDeps and devDeps) really needed in that case? I would not want every consuming app for this package to have to update all of this right away.

please see [comment](https://github.com/iTwin/viewer-components-react/pull/1097#discussion_r1854634704)

@raplemie @Maxime-Brassard  would like to merge this asap

raplemie:
@jason-crow I see you are in the "Viewer Components Reviewers" components reviewer for this package, as you are the one dealing mostly with peer deps, is that the solution we need for these packages or that will cause more problem than it will solve ? (The change from iTwinUI v2 direct dependency to a iTwinUI V3 peer dependency ?)

jason-crow:
> @jason-crow I see you are in the "Viewer Components Reviewers" components reviewer for this package, as you are the one dealing mostly with peer deps, is that the solution we need for these packages or that will cause more problem than it will solve ? (The change from iTwinUI v2 direct dependency to a iTwinUI V3 peer dependency ?)

The solution is to set is a peer in v3 and this is because the ThemeManager is a singleton, but while we have a mix in our pineapple repo of v2 and v3 we have to use a pnpfile.cjs workaround to switch it back to a dependency from a peerDependency so we can ensure its paired to v3 as we alias the v3 so the peer otherwise matches with v2. Soon we should be all on v3 so hoping we can undo our workarounds in that pnpfile.cjs but at this layer it should just be a peer if that makes sense

raplemie (approved):

jason-crow (approved):

Issue #1093, opened 2024-11-15
Tree widget: Determining Models tree nodes' visibility is slow on filtered hierarchies


grigasp:
When Models tree is filtered using large sets of target instances (tens of thousands), determining visibility status of each node:
- is very slow,
- blocks the main thread.

PR #1098, opened 2024-11-26
Tree widget: visibility status performance improvements

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/1093

Improvements made to visibility status calculation:
* For filtered tree built internal filtered instance tree structure to speed up filter targets lookup
* Removed tooltips creation when looking up ancestor visibility status as they are not used.
* Reduced main thread blocking when handling large amount of element to determine parent node visibility status.

Improvements for visibility change performance:
* Added logic to disable visibility change events handling while changing visibility to avoid computing visibility status of tree nodes while change is in progress.
* Added logic to disable neverDrawn/alwaysDrawn elements cache update while changing visibility.

Breaking change to beta API:
* Replaced `ModelsTreeVisibilityHandlerOverrides.changeElementState` with `ModelsTreeVisibilityHandlerOverrides.changeElementsState`. Instead of changing visibility for one element at the time the new function is invoked with a list of elements so update for all of them could be made in one go. 

Tried to measure performance on local machine. Used devTools profiler to check for long running tasks that blocks main thread (UI).

Measured time each getVisibilityStatus call takes on initial tree load to determine visibility status for nodes with most filtered paths. Root node - 87k, and two nodes directly under root with 43k and 34k paths:

| Get Visibility Status  | Before | after |
|------------------------|--------|-------|
| with 87k paths         | 1.2s   | 0.2s  |
| with 43k paths         | 1.4s   | 0.4s  |
| with 34k paths         | 1.1s   | 0.2s  |
| Main thread blocks (long running tasks) |
| > 10s                  |      1 |     0 |
| > 1s                   |      2 |     0 |
| > 100ms                |      1 |     0 |
| > 50ms                 |      1 |     0 |

Measured time it takes for `changeVisibilityStatus` to complete on root node:

| Visibility Change (87k paths) | before | After |
|-------------------------------|--------|-------|
| Turn All On from mixed state  |    63s |  1.2s |
| Main thread blocks (long running tasks)        |
| > 10s                         |      5 |     0 |
| > 1s                          |      9 |     0 |
| > 100ms                       |      5 |     1 |
| > 50ms                        |      0 |     2 | 
| Turn All Off from All On      |    79s |  1.1s |
| Main thread blocks (long running tasks)        |
| > 1min                        |      1 |     0 |
| > 10s                         |      1 |     0 |
| > 1s                          |      1 |     0 |
| > 100ms                       |      3 |     2 |
| > 50ms                        |      0 |     2 |
| Turn All On from All Off      |    84s |  1.2s |
| Main thread blocks (long running tasks)        |
| > 1min                        |      1 |     0 |
| > 10s                         |      1 |     0 |
| > 1s                          |      2 |     0 |
| > 100ms                       |      1 |     0 |
| > 50ms                        |      0 |     3 |


packages/itwin/tree-widget/api/tree-widget-react.api.md
```diff
@@ -276,7 +283,7 @@ export interface ModelsTreeVisibilityHandlerOverrides {
         on: boolean;
     }) => Promise<void>>;
     // (undocumented)
-    changeElementState?: HierarchyVisibilityHandlerOverridableMethod<(props: ChangeGeometricElementDisplayStateProps) => Promise<void>>;
+    changeElementsState?: HierarchyVisibilityHandlerOverridableMethod<(props: ChangeGeometricElementsDisplayStateProps) => Promise<void>>;
```
MartynasStrazdas, line 286: Wouldn't renaming a property for beta interface still be a breaking change?

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -5,7 +5,10 @@
 
 import { defer, from, map, merge, mergeAll, mergeMap } from "rxjs";
 import {
-  createNodesQueryClauseFactory, createPredicateBasedHierarchyDefinition, NodeSelectClauseColumnNames, ProcessedHierarchyNode,
+  createNodesQueryClauseFactory,
+  createPredicateBasedHierarchyDefinition,
+  NodeSelectClauseColumnNames,
+  ProcessedHierarchyNode,
 } from "@itwin/presentation-hierarchies";
```
MartynasStrazdas, line 12: Why are there cases of imports becoming multiline without adding or changing the imports themselves?

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -0,0 +1,255 @@
...
+  for (const child of node.children.values()) {
+    collectFilterTargets(filterTargets, child);
+  }
+}
+
+function addTarget(filterTargets: FilterTargets, node: FilteredTreeNode) {
+  switch (node.type) {
+    case "subject":
+      (filterTargets.subjects ??= new Set()).add(node.id);
+      return;
```
MartynasStrazdas, line 173: Could assignment and return be in one line.

packages/itwin/tree-widget/api/tree-widget-react.api.md
MartynasStrazdas: Missed it being mentioned in description :D

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -154,24 +157,56 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
   }
 
   public async getVisibilityStatus(node: HierarchyNode): Promise<VisibilityStatus> {
-    return firstValueFrom(this.getVisibilityStatusObs(node));
+    return firstValueFrom(
+      this.getVisibilityStatusObs(node).pipe(
+        // unsubscribe from the observable if the change request for this or ancestor node is received
+        takeUntil(this._changeRequest.pipe(filter(({ key, depth }) => depth <= node.parentKeys.length || HierarchyNodeKey.equals(node.key, key)))),
```
grigasp, line 163: Not sure the `depth <= node.parentKeys.length` is enough to say that it's an ancestor. E.g., we could have this:

```
- A
  - B
- C
  - D
```

We may be checking status of `B` and changing status of `C`. Depth of `C` is smaller than of `B`, but it's not an ancestor.

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -241,9 +276,13 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
-                return this.getElementDisplayStatus({ modelId, categoryId, elementId });
+              releaseMainThreadOnItemsCount(50),
+              mergeMap(([categoryKey, elementIds]) => {
+                const { modelId, categoryId } = parseCategoryKey(categoryKey);
+                return from(elementIds).pipe(
+                  releaseMainThreadOnItemsCount(1000),
+                  mergeMap((elementId) => this.getElementDisplayStatus({ modelId, categoryId, elementId, ignoreTooltip: true })),
+                );
               }),
             ),
```
grigasp, line 287: Do we need both `releaseMainThreadOnItemsCount` calls here? It seems the first one ensures we don't attempt to call `parseCategoryKey` for every element node in a single task. Then we call `getElementDisplayStatus` in batches of `1000` elements. Did you try changing the top-level batch size to `1000` and removing the inner `releaseMainThreadOnItemsCount`?

Also, maybe it makes sense to add a limit to how many tasks in parallel are launched my `mergeMap`?

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -465,77 +520,17 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
       return EMPTY;
     }
 
-    return this.changeElementState({
-      elementId: node.key.instanceKeys[0].id,
+    return this.changeElementsState({
+      elementIds: new Set([node.key.instanceKeys[0].id]),
```
grigasp, line 524: Let's try to be more correct, since we're now capable of handling multiple IDs at once.


```suggestion
      elementIds: new Set([...node.key.instanceKeys.map(({ id }) => id)]),
```

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -698,29 +693,68 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
-          return of(elementId).pipe(this.changeElementStateNoChildrenOperator({ on, isDisplayedByDefault }));
+          return this.queueElementChange(elementIds, on, isDisplayedByDefault);
         }),
       );
     });
-    return createVisibilityHandlerResult(this, props, result, this._props.overrides?.changeElementState);
+    return createVisibilityHandlerResult(this, props, result, this._props.overrides?.changeElementsState);
+  }
+
+  private queueElementChange(elementIds: Id64Set, on: boolean, visibleByDefault: boolean) {
```
grigasp, line 712: I'd rename this to `queueElementsVisibilityChange`

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -698,29 +693,68 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
+    const finishedSubject = new Subject<boolean>();
+    // observable to track if visibility change is finished/cancelled
+    const changeFinished = finishedSubject.pipe(startWith(false), shareReplay(1));
+
+    const changeObservable = from(elementIds).pipe(
+      // check if visibility change is not finished (cancelled) due to change overall change request being cancelled
+      takeUntil(changeFinished.pipe(filter((finished) => finished))),
```
grigasp, line 719: maybe `filter((finished) => finished)` should be moved to where `changeFinished` is created?

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -698,29 +693,68 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
+      this.changeElementStateNoChildrenOperator({ on, isDisplayedByDefault: visibleByDefault }),
+      tap({
+        next: () => {
+          // notify that visibility change is finished
+          finishedSubject.next(true);
+        },
+      }),
```
grigasp, line 726: Why do we care if it's finished or not?

change/@itwin-tree-widget-react-d9ea273e-f06f-4880-bb8b-a70991486fd9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Improve performance of visibility status calculation and visibility changes. Also reduce main thread blocking to avoid UI freezing.",
```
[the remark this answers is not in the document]
grigasp: Agree, it would be great if the PR description mentioned what improvement to expect. I understand it's not trivial to get the numbers for this kind of problem, but something approximate could be retrieved from the devtools?.. E.g., take one specific test case and provide something like this:

|                                                                                                         | Before the changes | After the changes |
| ------------------------------------------------------------------------------------------------------- | ------------------ | ----------------- |
| Time it takes to fully determine nodes' visibility for initial view, filtered by X target elements      | ?                  | ?                 |
| Time it takes to fully handle root node visibility change, with hierarchy filtered by X target elements | ?                  | ?                 |
| Number of main thread blocks above 1s                                                                   | ?                  | ?                 |
| Number of main thread blocks above 500ms                                                                | ?                  | ?                 |
| Number of main thread blocks above 100ms                                                                | ?                  | ?                 |
| Number of main thread blocks above 50ms                                                                 | ?                  | ?                 |

packages/itwin/tree-widget/src/components/trees/models-tree/internal/VisibilityChangeEventListener.ts
```diff
@@ -46,5 +49,14 @@ export function createVisibilityChangeEventListener(viewport: Viewport): IVisibi
       listeners.forEach((x) => x());
       listeners.length = 0;
     },
+    suppressChangeEvents: () => {
+      suppressChangeEvents++;
+    },
+    unSuppressChangeEvents: () => {
+      suppressChangeEvents--;
+      if (suppressChangeEvents === 0) {
+        handleVisibilityChange();
```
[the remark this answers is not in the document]
saskliutas: I don't think there is need to add additional logic to track if visibility change events happened while event listener was suppressed because we suppress and resume event listening only when tree visibility handler is changing visibility so I think we can assume that it will make visibility changes.

packages/itwin/tree-widget/src/components/trees/common/Tooltip.ts
```diff
@@ -13,14 +13,19 @@ export type Visibility = "visible" | "hidden" | "partial";
...
+export function createVisibilityStatus(status: "visible" | "hidden", options?: VisibilityStatusOptions): NonPartialVisibilityStatus;
+export function createVisibilityStatus(status: "visible" | "hidden" | "partial" | "disabled", options?: VisibilityStatusOptions): VisibilityStatus;
+export function createVisibilityStatus(status: Visibility | "disabled", { ignoreTooltip, tooltipStringId }: VisibilityStatusOptions = {}): VisibilityStatus {
   return {
     state: status === "disabled" ? "hidden" : status,
     isDisabled: status === "disabled",
-    tooltip: createTooltip(status, tooltipStringId),
+    tooltip: ignoreTooltip ? undefined : createTooltip(status, tooltipStringId),
   };
 }
```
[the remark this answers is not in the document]
saskliutas: Tooltip is built from `status` and `tooltipStringId`. There might be cases where we do not have `tooltipStringId` and tooltip is built only from `status`

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -0,0 +1,255 @@
...
+    addTarget(filterTargets, node);
+  }
+
+  if (!node.children) {
+    return;
+  }
+
+  for (const child of node.children.values()) {
+    collectFilterTargets(filterTargets, child);
+  }
```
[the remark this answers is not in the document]
saskliutas: In previous implementation if filter target is `element` it was adding it's `element` ancestors also https://github.com/iTwin/viewer-components-react/blob/485ec9d4bd818fb9ba6aab08af56287c18f41cb7/packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts#L523

So this does the same because we reach an `element` that is not a filter target it means that it's children is filter target.

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
saskliutas: I guess it could but `add` returns set object with added value so function return type would change from `void` to `Set`. Also I think current implementation is easier to read.

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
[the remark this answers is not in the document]
grigasp: Okay, seems like I was confused by the function name `collectFilterTargets`. Can we rename it to something more meaningful, e.g. `collectVisibilityChangeTargets` or something like that?

packages/itwin/tree-widget/src/components/trees/common/Tooltip.ts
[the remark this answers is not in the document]
grigasp: Maybe that could be handled by adding a distinction between `tooltipStringId` being `undefined` vs `""`?

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
saskliutas: Renamed

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
saskliutas: Moved

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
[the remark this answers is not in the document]
saskliutas: Renamed

packages/itwin/tree-widget/src/components/trees/models-tree/internal/VisibilityChangeEventListener.ts
```diff
@@ -11,14 +11,17 @@ import type { Viewport } from "@itwin/core-frontend";
 /** @internal */
 export interface IVisibilityChangeEventListener extends IDisposable {
   onVisibilityChange: BeEvent<() => void>;
+  suppressChangeEvents(): void;
+  unSuppressChangeEvents(): void;
```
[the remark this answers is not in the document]
saskliutas: Renamed

packages/itwin/tree-widget/src/components/trees/common/Tooltip.ts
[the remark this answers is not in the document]
saskliutas: Renamed `tooltipStringId` to `useTooltip` and changed type to `string | false` where `false` indicates that we don't need tooltip at all.

packages/itwin/tree-widget/src/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -0,0 +1,255 @@
...
+    };
+  }
+
+  if ("type" in parent && parent.type === "element") {
+    return {
+      id,
+      isFilterTarget,
+      type,
+      modelId: parent.modelId,
+      categoryId: parent.categoryId,
```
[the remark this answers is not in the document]
saskliutas: Looks like this is already existing issue. Filled a new issue to avoid increasing scope of this PR: https://github.com/iTwin/viewer-components-react/issues/1100

change/@itwin-tree-widget-react-d9ea273e-f06f-4880-bb8b-a70991486fd9.json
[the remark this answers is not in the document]
saskliutas: Updated description

packages/itwin/tree-widget/src/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
saskliutas: Tried using `mergeMap` second parameter instead of `releaseMainThreadOnItemsCount(50)` before but I saw the we get at least 1 longer running microtask >50ms (I guess it's for the root subject visibility as it has 87k filtered elements underneath)

packages/itwin/tree-widget/src/components/trees/models-tree/ModelsTreeDefinition.ts
saskliutas: Prettier updated imports

grigasp (approved):

PR #1104, opened 2024-12-02
[tree-widget]: repository tree investigation

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/778

Repository tree:

https://github.com/user-attachments/assets/36acb51b-1eab-429d-b0b8-3bce8641a156


packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
```diff
@@ -0,0 +1,78 @@
...
+  const result = await fetchData<Repository[]>(url, accessToken);
+  return result;
+}
+
+async function fetchData<T>(url: string, accessToken: string): Promise<T> {
+  const headers = {
+    Authorization: accessToken,
+    "Content-Type": "application/json",
+    Prefer: "return=representation",
+  };
```
grigasp, line 51: Based on `https://developer.bentley.com/apis/iTwins/operations/get-repositories-by-itwin-id/`, it seems we should also set the `Accept` header.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
```diff
@@ -0,0 +1,78 @@
...
+export async function getItwinRepositories(itwinId: string, accessToken: string, environment?: "PROD" | "QA" | "DEV"): Promise<ITwinRepository[]> {
+  const url = getRepositoriesUrl(itwinId, environment);
+  const result = await fetchData<ITwinRepository[]>(url, accessToken);
+  return result;
+}
+
+/**
+ * @internal
+ */
+export async function getRepositoryData(accessToken: string, url?: string): Promise<Repository[]> {
```
grigasp, line 39: I believe this function returns not repositories, but their content, doesn't it? Return type should be named accordingly.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
```diff
@@ -0,0 +1,78 @@
...
+  type?: string;
+  name?: string;
+}
+
+interface ITwinRepository {
+  id?: string;
+  class: string;
+  subClass?: string;
+  uri?: string;
+}
```
grigasp, line 25: I think we need to be more strict about these types.

For `ITwinRepository`, I think we should make `class` and `uri` required. `id` can probably be removed altogether.

For `Repository` (needs a rename), I think `id` and `displayName` should be required. Not sure what `name` means or how it's different from `displayName` - do we need it?

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+  accessToken: string;
+  itwinId: string;
+  setIsLoading: (isLoading: boolean) => void;
+  environment?: "PROD" | "QA" | "DEV";
+}
+
+/**
+ * @internal
+ */
+export function UseRepositoriesHierarchyProvider({ accessToken, itwinId, setIsLoading, environment }: UseRepositoriesHierarchyProviderProps) {
```
grigasp, line 23: functions should start in lower case

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesTree.tsx
```diff
@@ -0,0 +1,81 @@
...
+  environment?: "PROD" | "QA" | "DEV";
+  noDataMessage?: string;
+}
+
+/**
+ * @alpha
+ */
+export function RepositoriesTree({ selectionStorage, ...props }: RepositoriesTreeProps) {
+  return (
+    <UnifiedSelectionProvider storage={selectionStorage}>
```
grigasp, line 30: I don't think this tree should be hooked into unified selection.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesTree.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <Tree {...props} />
+    </UnifiedSelectionProvider>
+  );
+}
+
+function Tree({ accessToken, itwinId, noDataMessage, environment }: Omit<RepositoriesTreeProps, "selectionStorage">) {
+  const [isLoading, setIsLoading] = useState(true);
```
grigasp, line 37: I don't think we need to handle this ourselves. `useTree` hooks return an `isLoading` flag which we can use.

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+            const repositories = await getItwinRepositories(itwinId, accessToken, environment);
+            setIsLoading(false);
+            for (const repository of repositories) {
+              yield {
+                key: { type: "generic", id: repository.class },
+                label: formatLabel(repository.class),
+                children: !!repository.uri,
+                extendedData: { url: repository.uri },
+                parentKeys: [],
+              } as HierarchyNode;
```
grigasp, line 39:
```suggestion
              } satisfies HierarchyNode;
```

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+          } else {
+            const repositoryData = await getRepositoryData(accessToken, parentNode.extendedData?.url);
+            for (const data of repositoryData) {
+              yield {
+                key: { type: "generic", id: data.id ?? data.displayName },
+                label: data.displayName ?? data.name,
+                children: false,
+                extendedData: { type: data.type },
+                parentKeys: [...parentNode.parentKeys, parentNode.key],
+              } as HierarchyNode;
```
grigasp, line 50:
```suggestion
              } satisfies HierarchyNode;
```

apps/test-viewer/src/UseAccessToken.ts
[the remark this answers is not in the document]
MartynasStrazdas: used in `TestRepositoriesTree.tsx`.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
MartynasStrazdas: Each repository endpoint uses a different version of Accept so for now decided to not declare accept header.

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
```diff
@@ -51,10 +53,12 @@ export function UseRepositoriesHierarchyProvider({ accessToken, itwinId, setIsLo
             }
           }
         },
+        setHierarchyFilter() {
+          hierarchyChanged.raiseEvent();
```
grigasp, line 57: Do you see any issues without this?

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
MartynasStrazdas: All the repositories have a `displayName` except for the cesium one, which has a name instead, to not add a separate type added both displayName and name.

apps/test-viewer/src/UseAccessToken.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  useEffect(() => {
+    const fetchAccessToken = async () => {
+      if (state === AuthorizationState.Pending) {
+        setAccessToken(undefined);
+        setLoading(true);
+        return;
+      }
+
+      setLoading(true);
```
JonasDov, line 22: I think that we can `setLoading` at the start, since it is set to true in `if` block as well as after it
```suggestion
    const fetchAccessToken = async () => {
      setLoading(true);
      if (state === AuthorizationState.Pending) {
        setAccessToken(undefined);
        return;
      }

```

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
```diff
@@ -0,0 +1,110 @@
...
+  "RealityData" = 1,
+  "Storage" = 2,
+  "Forms" = 3,
+  "Issues" = 4,
+  "SensorData" = 5,
+  "CesiumCuratedContent" = 6,
+  "GeographicInformationSystem" = 7,
+  "Construction" = 8,
+  "Subsurface" = 9,
+}
```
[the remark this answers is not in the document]
MartynasStrazdas: My only problem with a union is changing a string means updating every place it's used (conditions). With an enum, you just update the enum, and it works everywhere. Right?

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
```diff
@@ -0,0 +1,110 @@
...
+    case RootNodeClassnameEnum.RealityData:
+      return <Svg3D />;
+    case RootNodeClassnameEnum.Storage:
+      return <SvgFolder />;
+    case RootNodeClassnameEnum.Forms:
+      return <SvgDetails />;
+    case RootNodeClassnameEnum.Issues:
+      return <SvgIssueReport />;
+    case RootNodeClassnameEnum.CesiumCuratedContent:
+      return <SvgGlobe />; // There is no Cesium icon in itwinUI-icons
```
JonasDov, line 102: Do we only use icons from itwinui-icons? You can get cesium svg from their github: https://github.com/CesiumGS/cesium/blob/main/Apps/Sandcastle/images/cesium-logomark.svg. Then place it in some folder and use it here instead.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
JonasDov: > I think we need to be more strict about these types.
> 
> For `ITwinRepository`, I think we should make `class` and `uri` required. `id` can probably be removed altogether.
> 
> For `Repository` (needs a rename), I think `id` and `displayName` should be required. Not sure what `name` means or how it's different from `displayName` - do we need it?

If we want to, we can be very strict about the types. There are some packages like zod https://zod.dev/,  that help enforce specific rules using regex  (This is out of scope for this PR, but just wanted to mention that).

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
```diff
@@ -0,0 +1,66 @@
...
+  environment?: "PROD" | "QA" | "DEV";
+}
+
+/**
+ * @internal
+ */
+export function UseRepositoriesHierarchyProvider({ accessToken, itwinId, setIsLoading, environment }: UseRepositoriesHierarchyProviderProps) {
+  return useMemo<() => HierarchyProvider>(
+    () => () => {
+      // let rootFilter: Props<HierarchyProvider["setHierarchyFilter"]>;
```
JonasDov, line 26: Forgot to remove this line

apps/test-viewer/src/TestRepositoriesTree.tsx
```diff
@@ -0,0 +1,27 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { RepositoriesTree } from "@itwin/tree-widget-react";
```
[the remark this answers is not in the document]
MartynasStrazdas: Moved, it did require adding additional dependencies to test-viewer, but I guess we never going to merge this so it should be fine.

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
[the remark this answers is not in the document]
MartynasStrazdas: Added a new file to hold it kept it an enum for now.

packages/itwin/tree-widget/src/components/trees/repositories-tree/FormatLabel.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { RootNodeClassnameEnum } from "./GetIcon";
+
+/**
+ * @internal
+ */
+export function formatLabel(Classname: string) {
```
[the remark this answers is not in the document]
MartynasStrazdas: renamed

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * @internal
+ */
+export function getRepositoryNodeIcon(node: PresentationHierarchyNode) {
+  if (node.nodeData.parentKeys.length === 0 && HierarchyNodeKey.isGeneric(node.nodeData.key)) {
+    return getRootNodeIcon(node.nodeData.key.id);
+  }
+
+  if (HierarchyNodeKey.isGeneric(node.nodeData.parentKeys[0]) && node.nodeData.extendedData?.type) {
```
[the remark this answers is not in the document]
MartynasStrazdas: Added repositoryType to extendedData I don't think we need to check the node type anymore.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
```diff
@@ -0,0 +1,78 @@
...
+      headers,
+    });
+    const result = await response.json();
+
+    // get the data avoiding links
+    delete result._links;
+    const values = Object.values(result);
+    const data = values[0];
+
+    return data as T;
```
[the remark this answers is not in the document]
MartynasStrazdas: changed for it to return unknown, kept `as` for the time being.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
MartynasStrazdas: changed name

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
MartynasStrazdas: Removed.

apps/test-viewer/src/UseAccessToken.ts
MartynasStrazdas: After refactor loading was removed as it was not used.

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
MartynasStrazdas: This was the only problem after updating both hierarchy dependancies.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesTree.tsx
MartynasStrazdas: changed

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
MartynasStrazdas: changed

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesTree.tsx
MartynasStrazdas: changed

apps/test-viewer/package.json
```diff
@@ -63,7 +63,10 @@
     "sass": "^1.77.5",
     "typescript": "~4.6.0",
     "vite": "^5.4.10",
-    "vite-plugin-static-copy": "^2.0.0"
+    "vite-plugin-static-copy": "^2.0.0",
+    "@itwin/presentation-hierarchies": "^1.3.0",
+    "@itwin/presentation-hierarchies-react": "^1.2.0",
+    "@itwin/presentation-shared": "^1.2.0"
```
grigasp, line 69: these should be under `dependencies`

apps/test-viewer/src/components/repositories-tree/RepositoriesTree.tsx
```diff
@@ -3,19 +3,17 @@
...
-import { useState } from "react";
 import { Flex, ProgressRadial, Text } from "@itwin/itwinui-react";
-import { TreeRenderer, UnifiedSelectionProvider, useUnifiedSelectionTree } from "@itwin/presentation-hierarchies-react";
-import { TreeWidget } from "../../../TreeWidget";
-import { Delayed } from "../common/components/Delayed";
-import { ProgressOverlay } from "../common/components/ProgressOverlay";
+import { TreeRenderer, useUnifiedSelectionTree } from "@itwin/presentation-hierarchies-react";
+import { Delayed } from "../../../../../packages/itwin/tree-widget/src/components/trees/common/components/Delayed";
+import { ProgressOverlay } from "../../../../../packages/itwin/tree-widget/src/components/trees/common/components/ProgressOverlay";
+import { TreeWidget } from "../../../../../packages/itwin/tree-widget/src/TreeWidget";
```
grigasp, line 10: the app should not import internal stuff from tree widget

apps/test-viewer/src/components/repositories-tree/RepositoriesTree.tsx
```diff
@@ -53,7 +43,7 @@ function Tree({ accessToken, itwinId, noDataMessage, environment }: Omit<Reposit
       );
     }
 
-    if (rootNodes.length === 0 && !isLoading) {
+    if (rootNodes.length === 0 && !treeProps.isLoading) {
```
grigasp, line 46: I think `isLoading` should be extracted from `useTree` result (similar to `rootNodes`) so that it doesn't land into the `treeProps` object.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesTree.tsx
grigasp: The `useUnifiedSelectionTree` hook should be changed to `useTree`

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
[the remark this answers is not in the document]
grigasp: > My only problem with a union is changing a string means updating every place it's used (conditions). With an enum, you just update the enum, and it works everywhere. Right?

Yes, but I think that's only a problem if you need decoupling between the key and the value. In our case, it's a "type" - we're not going to map multiple different values to the same type and it's unlikely these values are going to change (and even if they did, we'd probably want to update the enum keys too). So I think a string union is still preferred over enum.

apps/test-viewer/src/UseAccessToken.ts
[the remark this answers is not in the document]
grigasp: Hm, I think we could use `IModelApp.getAccessToken` for this. Of course we can't just take the token and keep it for the whole session, but we could request it everytime hierarchy provider sends a request. Only need to change `useRepositoriesHierarchyProvider` to take an access token getter rather than token itself. What do you think?

packages/itwin/tree-widget/src/components/trees/repositories-tree/FormatLabel.tsx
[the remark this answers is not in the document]
grigasp: I think the `switch` in this function clearly demonstrates why a string union for `ITwinRepositoryType` would be better. Instead of

```ts
export function formatLabel(repositoryClassname: string) {
  switch (ITwinRepositoryType[repositoryClassname as keyof typeof ITwinRepositoryType]) {
    // ...
  }
}
```

You'd have

```ts
export function formatLabel(repositoryType: ITwinRepositoryType) {
  switch (repositoryType) {
    // ...
  }
}
```

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
grigasp: What we store in these objects doesn't necessarily has to map 1:1 to what the service returns. I don't really like the idea of `fetchData` doing some magic like this:

```ts
    const result = await response.json();
    // get the data avoiding links
    delete result._links;
    const values = Object.values(result);
    const data = values[0];

    return data;
```

IMO, it should just return the JSON of `unknown` type. Also, since the request and response formats slightly differ between each service, I'd say we should have a function per service to supply proper headers and parse the response. So for example:

```ts
export async function getIModels(accessToken: string, url: string): Promise<RepositoryData[]> {
  const response = await fetchData({
    url, 
    accessToken,
    extraHeaders: {
      "Accept": "application/vnd.bentley.itwin-platform.v2+json",
      "Prefer": "return=representation",
    },
  });
  if (typeof response !== "object" || !("iModels" in response)) {
    throw new Error("Invalid response");
  }
  const imodels = response["iModels"];
  if (!Array.isArray(imodels)) {
    throw new Error("Invalid response");
  }
  return imodels.map((imodel) => {
    // convert `imodel` to `RepositoryData`.
  });
}
```

I guess @JonasDov suggestion to use a library like `zod` could be helpful here.

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
grigasp: Please file an issue and add a comment here pointing to that issue.

packages/itwin/tree-widget/src/components/trees/repositories-tree/GetIcon.tsx
grigasp: I think it's okay for now to use what we have in `itwinui-icons`. We can update the icons afterwards.

packages/itwin/tree-widget/src/components/trees/repositories-tree/RepositoriesService.ts
MartynasStrazdas: My idea was to avoid having 8-9 functions that would call for repository data while we are still in the investigation stage.

packages/itwin/tree-widget/src/components/trees/repositories-tree/FormatLabel.tsx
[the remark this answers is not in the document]
MartynasStrazdas: changed

apps/test-viewer/src/components/repositories-tree/RepositoriesTree.tsx
MartynasStrazdas: changed

apps/test-viewer/package.json
MartynasStrazdas: changed

apps/test-viewer/src/UseAccessToken.ts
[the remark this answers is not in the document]
MartynasStrazdas: Changed it to use imodelApp to get access token.

packages/itwin/tree-widget/src/components/trees/repositories-tree/UseRepositoriesHierarchyProvider.ts
MartynasStrazdas: Added an issue.

grigasp (commented):
LGTM

apps/test-viewer/src/components/repositories-tree/RepositoriesService.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function getRepositoriesUrl(itwinId: string, environment?: "PROD" | "QA" | "DEV") {
+  const prefix = {
+    QA: "qa-",
+    DEV: "dev-",
+    PROD: "",
+  };
+
+  return `https://${prefix[environment ?? "PROD"]}api.bentley.com/itwins/${itwinId}/repositories`;
```
aruniverse, line 71: why not continue to use `globalThis.IMJS_URL_PREFIX` here?

apps/test-viewer/src/components/repositories-tree/FormatLabel.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ITwinRepositoryType } from "./RepositoriesType";
+
+/**
+ * @internal
+ */
+export function formatLabel(iTwinRepositoryType: ITwinRepositoryType) {
```
aruniverse, line 11: maybe we just do something like this, https://stackoverflow.com/questions/26188882/split-pascal-case-in-javascript-certain-case?
convert from pascal case to sentence case?

grigasp: This looks interesting, I think it would work in majority of cases. But probably need to leave us ability to override the label with what we want, since the type won't necessarily convert to a proper label (e.g. for `iModel` I guess it might convert to "I Model"?)

apps/test-viewer/src/components/repositories-tree/RepositoriesService.ts
MartynasStrazdas: The idea was that this would be moved outside of test-viewer, with that intention would like not to use env directly.

aruniverse: when we do move it out into a separate pkg, i still think we should avoid exposing our envs. we should just have a prop for apps to pass in a baseurl for these services

apps/test-viewer/src/components/repositories-tree/FormatLabel.tsx
aruniverse: yea fair point about iModel

apps/test-viewer/src/components/repositories-tree/RepositoriesService.ts
aruniverse: Theoretically, if we hosted the service on some other endpoint or locally you should be able to point to that as well. I'm going to resolve this comment, but something we should def try to avoid when we publish this separately. 

We have been recommending users provide the url themselves if it's anything other than prod

aruniverse (approved):

apps/test-viewer/src/components/repositories-tree/RepositoriesService.ts
MartynasStrazdas: Didn't think about that, good point. Changed to use url instead of passing prefix.

PR #1122, opened 2024-12-17
Bump core, appui & presentation dependencies

grigasp:


packages/itwin/property-grid/src/hooks/UseDataProvider.ts
```diff
@@ -87,9 +87,11 @@ class PerformanceTrackingProvider implements IPresentationPropertyDataProvider {
     this._wrappedProvider.selectionInfo = selectionInfo;
   }
 
-  public dispose(): void {
-    this._wrappedProvider.dispose();
+  public [Symbol.dispose](): void {
+    this._wrappedProvider[Symbol.dispose]?.();
```
saskliutas, line 91: I guess this will break if property grid is used with older `presentation-components` versions as older provider won't have `[Symbol.dispose]`?

grigasp: you're right... applied the `safeDispose` pattern we use elsewhere

saskliutas (approved):

Issue #1099, opened 2024-11-27
[Tree Widget] model / category selection inconsistencies with selection based tools


jason-crow:
- the model tree in @itwin/tree-widget-react@3.x supports clicking on model / category nodes to select their related elements
- these selection events do not propagate to the imodel.selectionSet, unlike selection events from graphical elements
- due to this difference with how these model / category selection events are handled compared to graphical elements, there are some inconsistencies with selection based tools:
   - the appui's SyncUiEventId.SelectionSetChanged event only raised from imodel.selectionSet.onChange events
   - tools that synchronize with SyncUiEventId.SelectionSetChanged, such as clear selection tools and visibility tools, do not respond to these selections
   - furthermore, many of these tools that sync with SyncUiEventId.SelectionSetChanged use the imodel.selectionSet for their respective tool's action. (i.e. the clear selection tool when activated will clear all the selection from imodel.selectionSet)
   - consequently, tools that should respond to selection events fail to do so and even if they did they would fail to work correctly

grigasp:
Thanks for reporting this. 

I started some discussions with the team responsible for Tools which use the `SelectionSet`. Depending on the outcome, we may need to temporarily add an option to disable selection of certain nodes, like we had in 2.x widget, until we can properly support selecting Models and Categories. Will update on this shortly.

grigasp:
@jason-crow, I have a couple of PRs that will fix this, but they're only going to be available with `5.0` release of `itwinjs-core`. Can you wait for that or would you like the option in tree widget to disable selection of specific nodes, as a short-term solution?

jason-crow:
> [@jason-crow](https://github.com/jason-crow), I have a couple of PRs that will fix this, but they're only going to be available with `5.0` release of `itwinjs-core`. Can you wait for that or would you like the option in tree widget to disable selection of specific nodes, as a short-term solution?

We would consider this a requirement for enabling the tree widget v3, so it would be nice to have that enabled before itwinjs v5

@vermanas  @jjbeckman13

grigasp:
[...]

PR #1124, opened 2024-12-17
Tree widget: Add `selectionPredicate` to models tree building blocks

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/1099.

The predicate receives a node and its type to help consumers decide whether selection should be allowed or not. Example:

```tsx
function MyModelsTree(props) {
  return (
    <ModelsTreeComponent
      {...props}
      selectionPredicate={myNodesSelectionPredicate}
    />
  )
}

const myNodesSelectionPredicate: ComponentProps<typeof ModelsTreeComponent>["selectionPredicate"] = ({ node, type }) => {
  if (type === "element") {
    return true;
  }
  return node.label.includes("whatever");
};
```

### DIFF
--- a/apps/test-viewer/src/UiProvidersConfig.tsx
+++ b/apps/test-viewer/src/UiProvidersConfig.tsx
@@ -114,16 +114,21 @@ const configuredUiItems = new Map<string, UiItem>([
               {
                 id: ModelsTreeComponent.id,
                 getLabel: () => ModelsTreeComponent.getLabel(),
-                render: (props) => (
-                  <ModelsTreeComponent
-                    getSchemaContext={getSchemaContext}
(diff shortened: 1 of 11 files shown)


grigasp:
> Perhaps we could visually show which nodes are not selectable?

I guess that would be nice for users to know that before clicking on the node, but I think the visual should be very subtle to avoid extra noise. Got anything in mind?

Note: the primary use case for this feature is to allow consumers temporarily disable selection of specific nodes (e.g. subjects, models, categories) until other components are able to handle that stuff being selected. As such, I don't think we should invest much time into improving UX for it - the change should be simple to implement.

MartynasStrazdas (approved):

PR #1133, opened 2024-12-31
Pass `enableVirtualization` flag to itwinui `<Tree />`

### LINKED ISSUE #1132: Tree Widget: `enableVirtualization` prop passed to `<TreeRenderer/>` not passed down to iTwinUi's `<Tree/>`
[`<TreeRenderer />`](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/tree-widget/src/components/trees/common/components/TreeRenderer.tsx) accepts `enableVirtualization` as a prop but does not pass it down to iTwinUI's [<Tree />](https://itwinui.bentley.com/docs/tree#virtualization).

Either, `<TreeRenderer />` should not accept `enableVirtualization` as a prop or this prop should be passed down.

anmolshres98:
Closes #1132

### DIFF
--- a/change/@itwin-tree-widget-react-424293ad-271d-412a-8226-14565b130d88.json
+++ b/change/@itwin-tree-widget-react-424293ad-271d-412a-8226-14565b130d88.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Pass `enableVirtualization` flag to itwinui `Tree`",
+  "packageName": "@itwin/tree-widget-react",
+  "email": "anmolshres98@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/itwin/tree-widget/src/components/trees/common/components/TreeRenderer.tsx
+++ b/packages/itwin/tree-widget/src/components/trees/common/components/TreeRenderer.tsx
@@ -57,6 +57,7 @@ export function TreeRenderer({
   checkboxProps,
   reloadTree,
   size,
+  enableVirtualization,
   ...props
 }: TreeRendererProps) {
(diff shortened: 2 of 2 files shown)


grigasp:
@saskliutas, what do you think about this? IMO, our renderer should always use virtualization and the prop should not be available for consumers of our component.

saskliutas:
> @saskliutas, what do you think about this? IMO, our renderer should always use virtualization and the prop should not be available for consumers of our component.

I think it makes sense to allow consumer to decide whether virtualization is used or not. This `TreeRenderer` is one of the building blocks we provide so it might be used to build custom tree that do not require virtualization.

grigasp:
> > @saskliutas, what do you think about this? IMO, our renderer should always use virtualization and the prop should not be available for consumers of our component.
> 
> I think it makes sense to allow consumer to decide whether virtualization is used or not. This `TreeRenderer` is one of the building blocks we provide so it might be used to build custom tree that do not require virtualization.

Sounds good. Should `style` prop also be changed to have effect?

grigasp:
> > > > @saskliutas, what do you think about this? IMO, our renderer should always use virtualization and the prop should not be available for consumers of our component.
> > > 
> > > 
> > > I think it makes sense to allow consumer to decide whether virtualization is used or not. This `TreeRenderer` is one of the building blocks we provide so it might be used to build custom tree that do not require virtualization.
> > 
> > 
> > Sounds good. Should `style` prop also be changed to have effect?
> 
> I this this could be solved in a separate PR. We should probably also review if there are more props that are exposed but not propagated down to the base iTwinUI `Tree` component.

Filed a followup issue: https://github.com/iTwin/viewer-components-react/issues/1137

Issue #1143, opened 2025-01-15
Tree widget: Update AppUI dependencies to 5.x


grigasp:
Ideally, we should update `peerDependencies` to support `^5.0` in addition to what's already supported. And update regular / dev dependencies to the latest 5.x.

MartynasStrazdas:
blocked by https://github.com/iTwin/appui/issues/1165

grigasp:
blocked by https://github.com/iTwin/appui/issues/1185

PR #1147, opened 2025-01-22
Tree widget: Add support for AppUI 5.0

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/1143.

AppUI 5.x dropped cjs support, so we had to get our tests to run in ESM mode too. Also added `type: "module"` and `exports` attributes to `package.json`.


packages/itwin/tree-widget/src/test/setup.ts
```diff
@@ -66,6 +43,35 @@ global.DOMRect = class DOMRect {
...
+    chaiJestSnapshot.resetSnapshotRegistry();
+    getGlobalThis().IS_REACT_ACT_ENVIRONMENT = true;
+  },
+  beforeEach() {
+    // enable strict mode for each test by default
+    configure({ reactStrictMode: !process.env.DISABLE_STRICT_MODE });
+
+    // set up snapshot name
+    const currentTest = (this as unknown as Mocha.Context).currentTest!;
+    const sourceFilePath = currentTest.file?.replace(`lib${path.sep}cjs${path.sep}test`, `src${path.sep}test`).replace(/\.(jsx?|tsx?)$/, "");
```
saskliutas, line 61: Shouldn't it be:
```suggestion
    const sourceFilePath = currentTest.file?.replace(`lib${path.sep}esm${path.sep}test`, `src${path.sep}test`).replace(/\.(jsx?|tsx?)$/, "");
```

saskliutas (changes requested):

scripts/package-cjs.js
```diff
@@ -0,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+const fs = require("node:fs");
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
const fs = require("node:fs");
```

aruniverse (approved):
changes look fine for files outside tree-widget dir

there should a version of the viewer with the "next" dist tag you can use that has support for appui 5 for testing

packages/itwin/tree-widget/src/e2e-tests/ModelsTree.test.ts-snapshots/Models-tree-Density-default-instances-focus---too-many-results-1-chromium-linux.png
saskliutas, line 1: Any ideas why screenshot size changed?

grigasp: yes... I had an `.env.` file set up to enable tree _and property_ widgets, and widgets configuration in `.env` took precedence over query params, so my screenshots were smaller due to property grid taking space in the same panel. 

Changed test-viewer to make widgets config in query params take precedence, re-ran the tests and updated the screenshots.

grigasp:
> changes look fine for files outside tree-widget dir
> 
> there should a version of the viewer with the "next" dist tag you can use that has support for appui 5 for testing

We'll need to wait for all repo packages to upgrade for AppUI v5 support before we can bump test-viewer's dependency.

saskliutas (approved):

PR #1152, opened 2025-01-24
Drop certa in favor of vitest

a-gagnon:
### Drop certa in favor of vitest for running tests.

This package has a dependency on appui. In their latest major (5.x), they have dropped support for CJS. This package will eventually drop CJS as well in order to become an ESM only package.

As a first step towards this goal, this PR drops the current test runner `certa` in favor of `vitest` which supports ESM and is also simpler.
- Running and debugging tests using the launch configuration works as expected
- Coverage reports work as expected


.vscode/settings.json
aruniverse, line 1: i personally prefer these settings too, but need larger buy in from others in this repo
can you revert this?

change/@itwin-measure-tools-react-cf33bee1-ecf6-4c62-9114-68a6bad1fcdc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
aruniverse, line 2: is there a "none" type? if so lets do that, and remove the comment too, as these change files are user facing, and the api of the pkg didnt change, just the internals

packages/itwin/measure-tools/vite.config.ts
```diff
@@ -0,0 +1,30 @@
...
+      exclude: ["lib/*", "src/test/*"],
+      include: ["src/*"],
+      provider: "istanbul",
+      reportsDirectory: "lib/coverage",
+      reporter: ["html", "text-summary"],
+      thresholds: {
+        statements: 40,
+        branches: 30,
+        functions: 40,
+        lines: 40,
```
hl662, line 26: why the increased values in branches and lines? package.json previously only had 20 for all values

packages/itwin/measure-tools/vite.config.ts
```diff
@@ -0,0 +1,30 @@
...
+  test: {
+    globals: true,
+    environment: "jsdom",
+    setupFiles: ["src/test/measure-tools/_Setup.test.ts"],
+    include: ["src/test/measure-tools/**/*.test.*"],
+    exclude: ["src/test/measure-tools/_Setup.test.ts"],
+    coverage: {
+      exclude: ["lib/*", "src/test/*"],
+      include: ["src/*"],
+      provider: "istanbul",
```
hl662, line 19: I recommend using v8 instead of istanbul, see [this](https://github.com/iTwin/itwinjs-core/pull/7252) for explanation on why

packages/itwin/measure-tools/vite.config.ts
aruniverse: This is not a bad thing, honestly the closer we get to complete coverage the better

a-gagnon: Yeah, I adjusted the thresholds to better reflect the current state of the package. That'll force subsequent contributions to at least think of the coverage.

packages/itwin/measure-tools/vite.config.ts
a-gagnon: Done

change/@itwin-measure-tools-react-cf33bee1-ecf6-4c62-9114-68a6bad1fcdc.json
aruniverse: you actually want to change type not dependentChangeType

a-gagnon: should be good this time around.

change/@itwin-measure-tools-react-cf33bee1-ecf6-4c62-9114-68a6bad1fcdc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Drop certa in favor of vitest.",
```
aruniverse, line 3:
```suggestion
  "comment": "",
```

aruniverse (approved):

PR #1160, opened 2025-01-27
Drop CJS output from measure-tools-react package.

a-gagnon:
Following what appui did with their new major (^5.), drop the CJS build output for the measure-tools-react package.

To comply with ESM rules and requirements, add the extension to all the imports in files.


packages/itwin/measure-tools/tsconfig.json
```diff
@@ -1,6 +1,13 @@
 {
   "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "compilerOptions": {
+    "jsx": "react",
+    "module": "NodeNext",
+    "moduleResolution": "nodenext",
+    "target": "ES2020",
+    "outDir": "./lib"
```
aruniverse, line 8: i wonder if we should leave it as `/lib/esm`

packages/itwin/measure-tools/package.json
```diff
@@ -43,13 +44,13 @@
   },
   "peerDependencies": {
     "@itwin/appui-abstract": "^4.7.0",
-    "@itwin/appui-react": "^4.17.0",
-    "@itwin/components-react": "^4.17.0",
+    "@itwin/appui-react": "^4.17.0 || ^5.0.0",
```
aruniverse, line 47: thoughts on dropping appui 4.17?

aruniverse (commented):
can you test this out in the test-viewer in this repo?

packages/itwin/measure-tools/package.json
a-gagnon: Pineapple still uses it for now, and it seems quick to drop support for 4.17.x given the short amount of time everyone had to update to 5.x. I'd vote to keep like this for a few weeks at least.

packages/itwin/measure-tools/tsconfig.json
a-gagnon: That would only affect deep imports, which shouldn't be a thing in the first place.

a-gagnon:
> can you test this out in the test-viewer in this repo?

I did, and everything looks okay. Currently looking at the linter problem, I recall we had the same thing happen in our repo. Haven't thought of running `lint` locally.

aruniverse (approved):

PR #1162, opened 2025-01-28
Tree widget: Test improvements

grigasp:
- Fix coverage not being captured (wrong paths in config).
- Run tests in parallel.
- Move all source under `src/tree-widget` directory to avoid source filed laying around `test` and `e2e-tests` directories.


packages/itwin/tree-widget/src/tree-widget-react.ts
```diff
@@ -3,8 +3,8 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-export { TreeWidget } from "./TreeWidget.js";
-export { createTreeWidget, TreeWidgetComponent } from "./components/TreeWidgetUiItemsProvider.js";
-export { SelectableTree, SelectableTreeDefinition } from "./components/SelectableTree.js";
-export { TreeWithHeader } from "./components/tree-header/TreeWithHeader.js";
-export * from "./components/trees/index.js";
+export { TreeWidget } from "./tree-widget/TreeWidget.js";
```
saskliutas, line 6: Maybe source code folder should be renamed to `tree-widget-react` to match barrel file and package name. Same as you did for property grid?

grigasp: agree, changed

saskliutas (approved):

PR #1170, opened 2025-01-30
Tree widget: Enter alpha mode in `next`

grigasp:


aruniverse (approved):

aruniverse (approved):

aruniverse:
you think we can delete the [tree-widget-2.x](https://github.com/iTwin/viewer-components-react/tree/tree-widget-2.x) branch?

MartynasStrazdas (approved):

grigasp:
> you think we can delete the [tree-widget-2.x](https://github.com/iTwin/viewer-components-react/tree/tree-widget-2.x) branch?

That branch was used to release 2.x patches, which I hope we won't need to do anymore. So I would say yes, unless @saskliutas thinks otherwise.

saskliutas:
> > you think we can delete the [tree-widget-2.x](https://github.com/iTwin/viewer-components-react/tree/tree-widget-2.x) branch?
> 
> That branch was used to release 2.x patches, which I hope we won't need to do anymore. So I would say yes, unless @saskliutas thinks otherwise.

I don't see a reason why we need to delete it but I am not against it.. We can always recreate it from the latest 2.x tag.

PR #1177, opened 2025-02-03
Replace more usage of deprecated APIs in map-layers

hl662:
- Replace use of Popup in transparency popup
- AttachMapLayerPopup is also refactored, ~but needs more investigation into why the modal opens then closes immediately (event.stopPropagation isn't working well...). Relevant [SO](https://stackoverflow.com/questions/72315874/react-click-outside-event-happens-right-after-click-to-open-preventing-the-mod/72316017#72316017) question~

  Update:  The issue was when the modal opens from within the Popover's content, it's triggering state changes that affected the Popover's visibility. Fix- modified the state management to keep the Popover open while the MapUrlDialog modal is active - @pachhaibipin 

TODOs:
- ~Run `pnpm test`, make sure to fix all tests~
- ~Refactor SubLayersPopupButton to use iTwinUI's Popover.~

pr

DISCUSSION:
     if you look into the attached gif below, the attached map layer list does not get updated once we delete or edit the custom layer definitions. It didn't seem very intuitive. However, diving into the implementation, the code flow suggests once the layer is attached, it won't be updated with delete/edit action. Only way seems to be detaching it. Do we want to keep it that way? Should it detach/update the map layer list once we delete/edit the added definitions? wanted to confirm if that’s the intended flow @mdastous-bentley 

flow


packages/itwin/map-layers/src/ui/widget/SubLayersPopupButton.tsx
```diff
@@ -29,39 +25,33 @@ export function SubLayersPopupButton(props: SubLayersPopupButtonProps) {
...
-        onClick={togglePopup}
-      >
-        <SvgLayers />
-      </Button>
-      {/*eslint-disable-next-line @typescript-eslint/no-deprecated */}
-      <Popup isOpen={popupOpen} position={RelativePosition.BottomRight} onClose={onOutsideClick} target={buttonRef.current}>
-        <div className="map-transparency-popup-panel">
-          <div ref={panelRef} className="map-manager-sublayer-panel">
-            <SubLayersPanel {...props} width={390} height={350} />
+      <Popover
```
hl662, line 30: @aruniverse we previously had a discussion on whether our packages should embed ThemeProvider by default. While I don't know what came out of that, do you think we should add `portal` [property](https://itwinui.bentley.com/docs/popover#props) to the `Popover` components?

aruniverse: dont think portal is neeeded, nor should we wrap our component in theme provider

aruniverse:
> @mdastous-bentley please check the PR description and confirm that's the intended behavior

add a gif of your changes in dark mode, which often makes it easier to find bugs

pachhaibipin:
> > @mdastous-bentley please check the PR description and confirm that's the intended behavior
> 
> add a gif of your changes in dark mode, which often makes it easier to find bugs

Added gif to the description. couldn't find a way to turn dark mode on test-viewer app. Haven't made any changes yet to that. wanted to confirm if that's the intended flow first. 
Are you asking to add gif for the changes made in this PR though? @aruniverse

aruniverse:
> Are you asking to add gif for the changes made in this PR though?

yes

you should just need to pass in the theme prop to the viewer

pachhaibipin:
> * Replace use of Popup in transparency popup
> * AttachMapLayerPopup is also refactored, but needs more investigation into why the modal opens then closes immediately (event.stopPropagation isn't working well...). Relevant [SO](https://stackoverflow.com/questions/72315874/react-click-outside-event-happens-right-after-click-to-open-preventing-the-mod/72316017#72316017) question
>   Update:  The issue was when the modal opens from within the Popover's content, it's triggering state changes that affected the Popover's visibility. Fix- modified the state management to keep the Popover open while the MapUrlDialog modal is active - @pachhaibipin
> 
> TODOs:
> 
> * ~Run `pnpm test`, make sure to fix all tests~
> * ~Refactor SubLayersPopupButton to use iTwinUI's Popover.~
> 
pr

> DISCUSSION: 
   if you look into the attached gif below, the attached map layer list does not get updated once we delete or edit the custom layer definitions. It didn't seem very intuitive. However, diving into the implementation, the code flow suggests once the layer is attached, it won't be updated with delete/edit action. Only way seems to be detaching it. Do we want to keep it that way? Should it detach/update the map layer list once we delete/edit the added definitions? wanted to confirm if that’s the intended flow @mdastous-bentley

flow

pachhaibipin:
> > Are you asking to add gif for the changes made in this PR though?
> 
> yes
> 
> you should just need to pass in the theme prop to the viewer

updated with gifs

mdastous-bentley:
There are two concepts two understand:
1. The list of predefined map layer sources. (Which is later used to attach a map-layer to a viewport)
2. Lists of currently attached map-layers in the active viewport.

You can delete a map layer source from list #1, and keep the "associated" map-layers attached in the viewport, no problem.  All it means is that you won't be able to re-add by using the same map layer source later on.

I don't think we want to mess with this actual design at this point.

hl662:
> I don't think we want to mess with this actual design at this point.

Got it. The design has been reverted back to this a couple commits ago

change/@itwin-map-layers-c81f8292-ad05-41d1-80b8-08e9731fce23.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
aruniverse, line 2: should be none, as we will just be doing another patch-release

aruniverse (approved):

PR #1186, opened 2025-02-12
[Tree-widget]: Widget header + visual fixes

MartynasStrazdas:
Changes:
 - Added a search box based on mockups to the tree selector component
 - Renamed tree selector to widget header for clarity. (Internal)
 - Fixed Tree header button styling
 - Updated hierarchies react package and added buttons

### DIFF
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/tree-header/TreeHeader.tsx
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/tree-header/TreeHeader.tsx
@@ -5,67 +5,29 @@
 
 import "./TreeHeader.scss";
 import classnames from "classnames";
-import React, { useEffect, useState } from "react";
 
 import type { PropsWithChildren } from "react";
 import type { Viewport } from "@itwin/core-frontend";
 /** @public */
 export interface TreeHeaderButtonProps {
   viewport: Viewport;
-  // density?: "default" | "enlarged";
   onFeatureUsed?: (feature: string) => void;
 }
 
-/** @beta */
-export interface TreeFilteringProps {
-  /** Filtering is cleared after everything's loaded */
-  onFilterStart: (newFilter: string) => void;
-  /** listens for onClick event for Clear (x) icon */
-  onFilterClear: () => void;
-  /** Total number of results/entries */
-  resultCount?: number;
-  /** Current selected result index */
-  selectedIndex?: number;
-  /** Callback to currently selected result/entry change */
-  onSelectedChanged?: (index: number) => void;
-  /** Should the search box be disabled */
-  isDisabled?: boolean;
-}
-
 interface TreeHeaderProps {
-  filteringProps?: TreeFilteringProps;
-  /** Modifies the density of tree header. `enlarged` header contains larger content */
-  density?: "default" | "enlarged";
   className?: string;
 }
 
-/// TODO: ADD SEARCH BOX
 export function TreeHeader(props: PropsWithChildren<TreeHeaderProps>) {
-  const { filteringProps, density, className, children } = props;
-  const [isSearchOpen, setIsSearchOpen] = useState<boolean>(false);
-  // const { enabled: instanceFocusEnabled } = useFocusedInstancesContext();
-  const size = density === "enlarged" ? "large" : "small";
-
-  useEffect(() => {
-    if (filteringProps?.isDisabled) {
-      setIsSearchOpen(false);
-    }
-  }, [filteringProps?.isDisabled]);
+  const { className, children } = props;
 
   return (
     <div className={classnames("tree-widget-tree-header", className)}>
-      <HeaderButtons contracted={isSearchOpen} size={size}>
-        {children}
-      </HeaderButtons>
+      <HeaderButtons>{children}</HeaderButtons>
     </div>
   );
 }
 
-interface HeaderButtonsProps {
-  contracted: boolean;
-  size: "large" | "small";
-}
-
-function HeaderButtons(props: PropsWithChildren<HeaderButtonsProps>) {
+function HeaderButtons(props: PropsWithChildren) {
   return <>{props.children}</>;
 }
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/TreeRenderer.tsx
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/TreeRenderer.tsx
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 /* eslint-disable @itwin/no-internal */
 
-import { LocalizationContextProvider, TreeRenderer as PresentationTree } from "@itwin/presentation-hierarchies-react";
+import { createFilterAction, LocalizationContextProvider, TreeRenderer as PresentationTree } from "@itwin/presentation-hierarchies-react";
 import { useHierarchiesLocalization } from "../UseHierarchiesLocalization.js";
-import { TreeItemVisibilityButton } from "./TreeNodeVisibilityButton.js";
+import { createVisibilityAction } from "./TreeNodeVisibilityButton.js";
 
 import type { TreeItemVisibilityButtonProps } from "./TreeNodeVisibilityButton.js";
 /** @beta */
@@ -25,7 +25,10 @@ export function TreeRenderer({ rootNodes, onNodeClick, expandNode, visibilityBut
         onNodeClick={onNodeClick}
         expandNode={expandNode}
         rootNodes={rootNodes}
-        actionsRenderer={visibilityButtonProps ? (node) => <TreeItemVisibilityButton {...visibilityButtonProps} node={node} /> : undefined}
+        actions={[
+          ...(visibilityButtonProps ? [createVisibilityAction(visibilityButtonProps)] : []),
+          createFilterAction({ onFilter: props.onFilterClick, getHierarchyLevelDetails: props.getHierarchyLevelDetails, label: "Filter" }),
+        ]}
       />
     </LocalizationContextProvider>
   );
--- a/apps/test-viewer/pnpm-lock.yaml
+++ b/apps/test-viewer/pnpm-lock.yaml
@@ -960,14 +960,14 @@ packages:
       react: ^17.0.0 || ^18.0.0
       react-dom: ^17.0.0 || ^18.0.0
 
-  '@itwin/presentation-core-interop@1.2.0':
-    resolution: {integrity: sha512-MxlcJpOXZnMhfwGfrH1Lck8eR2f638bb+lIBa6EP4CY5xLSNk5/zer+Fi6UtOpYdTvIGJcwEJs3Q9BilSnAPgg==}
+  '@itwin/presentation-core-interop@1.3.0':
+    resolution: {integrity: sha512-3EK+a4BI418w+bCKi13n+GMCETOHbxsvXck1aMl9TFKUnYa4z0/UmYNR4WFyICdv6HYeb4lgfeUPN0li/hNNvw==}
     peerDependencies:
-      '@itwin/core-bentley': ^4.1.0
-      '@itwin/core-common': ^4.1.0
-      '@itwin/core-geometry': ^4.1.0
-      '@itwin/core-quantity': ^4.1.0
-      '@itwin/ecschema-metadata': ^4.1.0
+      '@itwin/core-bentley': ^4.1.0 || ^5.0.0
+      '@itwin/core-common': ^4.1.0 || ^5.0.0
+      '@itwin/core-geometry': ^4.1.0 || ^5.0.0
+      '@itwin/core-quantity': ^4.1.0 || ^5.0.0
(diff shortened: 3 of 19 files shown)


grigasp:
Purely based on the attached demo:

- Missing spacing between Select / Input and Search button:
  image
- Input outline doesn't fit.
- Spacing above / below action buttons is inconsistent.
- The spinner during search progress is temporary, until we get skeleton, right?

MartynasStrazdas:
> Purely based on the attached demo:
> 
> * Missing spacing between Select / Input and Search button:
>   image
> * Input outline doesn't fit.
> * Spacing above / below action buttons is inconsistent.
> * Are we using correct variables for spacing overall? It looks like mockups have more padding.
> * The spinner during search progress is temporary, until we get skeleton, right?

- Added spacing between select/search to icon button.

- For the input outline on focus it seems to be an itwinUI issue will make a demo for them later.

- Fixed the spacing for buttons.

- Right now correct paddings should be used, I'm not sure for what padding is used in mockups for search/button, using standard 8px (0.5rem) right now (visually it seems the mockup is using 0.5 rem as well).

- Did not touch loading, it is temporary until we get linear loading, I think we discussed using skeleton only for initial load.

grigasp:
All stuff in the header should be disable while we load available trees. Now it's usable:
image

Also, the "search" tooltip string should start with upper case.

grigasp:
> * Added spacing between select/search to icon button.

👍 

> * For the input outline on focus it seems to be an itwinUI issue will make a demo for them later.

@GerardasB @mayank99 The outline is a problem on the buttons too:
image
image
image

> * Fixed the spacing for buttons.

👍 

> * Right now correct paddings should be used, I'm not sure for what padding is used in mockups for search/button, using standard 8px (0.5rem) right now (visually it seems the mockup is using 0.5 rem as well).

As far as I can tell, the header in mockups is set to use 12px:
image

> * Did not touch loading, it is temporary until we get linear loading, I think we discussed using skeleton only for initial load.

👍

grigasp:
**Dark mode**

image

- Our tree selector has a shadow. Mockups don't.
- The border / separator under header buttons seems to be of wrong color.
- Why is the tree background darker than the rest of the widget?

MartynasStrazdas:
> **Dark mode**
> 
> image
> 
> * Our tree selector has a shadow. Mockups don't.
> * The border / separator under header buttons seems to be of wrong color.
> * Why is the tree background darker than the rest of the widget?

image
image

* Our tree selector has a shadow. Mockups don't. & Why is the tree background darker than the rest of the widget?
It seems the dark mode itself has incorrect background color, this could be because of theme bridge.
Light mode has tree have a gray background which should not be the case, kiwi sandbox does not seem to have this issue, still looking into it, but the fix for this will probably be in `hierarchies-react`.

 * The border / separator under header buttons seems to be of wrong color.
 Added an actual Seperator component, it is less visible in dark mode, because of incorrect background color.

grigasp:
>   Added an actual Seperator component, it is less visible in dark mode, because of incorrect background color.

Looks better, thanks. Why is there are space between the separator and the first node?

grigasp:
Also, while playing around with search and instances focus toggle button. Currently it gets disabled when we start searching (enter something in the input box). I think it would be better if it got disabled as soon as you click on the search button and the search input shows up.

MartynasStrazdas:
> Also, while playing around with search and instances focus toggle button. Currently it gets disabled when we start searching (enter something in the input box). I think it would be better if it got disabled as soon as you click on the search button and the search input shows up.

Wouldn't this cause unnecessary reloads/filtering, you click on focus mode it filters tree to selected items, you click on search it starts to reload tree to all nodes, and then you search it filters again.

MartynasStrazdas:
> > It seems the dark mode itself has incorrect background color, this could be because of theme bridge.
> > Light mode has tree have a gray background which should not be the case, kiwi sandbox does not seem to have this issue, still looking into it, but the fix for this will probably be in hierarchies-react.
> 
> This doesn't answer the question regarding shadow.

The shadow comes from kiwi side, will file an issue for it.

> > Added an actual Seperator component, it is less visible in dark mode, because of incorrect background color.
> 
> Looks better, thanks. Why is there are space between the separator and the first node?

In the mockups there is an additional padding after Seperator.

> > Ps. Adjusted paddings to reflect mockups better, this fixed the focus on selector/search problem on our end.
> 
> I think there's still a problem with outline on header buttons - the outline is displayed on top of sibling buttons. I think it's because we don't have any spacing between them.

This could be slightly tricky, because we allow buttons to be provided therefor, we can't add margins to the buttons itself. We could have something like this `.container-name button { margin-right: var(4px equivalent)} @saskliutas I'm not sure if this would be an appropriate solution as it could cause issues in the future. It would be best if kiwi would apply this focus border inside the component instead of outside.

grigasp:
> > Also, while playing around with search and instances focus toggle button. Currently it gets disabled when we start searching (enter something in the input box). I think it would be better if it got disabled as soon as you click on the search button and the search input shows up.
> 
> Wouldn't this cause unnecessary reloads/filtering, you click on focus mode it filters tree to selected items, you click on search it starts to reload tree to all nodes, and then you search it filters again.

Enabling focus mode and clicking on an element shows a filtered-down hierarchy. When I open the search, I expect it to be applied on the currently visible hierarchy, and that is not what's happening - the hierarchy changes to full as soon as I start typing. I think changing it to full as soon as user clicks on the search button makes it more obvious why that is happening and what the search is going to be applied on. I don't think we should worry too much about extra reloads/filtering.

grigasp:
> In the mockups there is an additional padding after Seperator.

In most - yes, in the one I was looking at - no 🤔 I'll ask UX for clarification...

saskliutas:
> > > It seems the dark mode itself has incorrect background color, this could be because of theme bridge.
> > > Light mode has tree have a gray background which should not be the case, kiwi sandbox does not seem to have this issue, still looking into it, but the fix for this will probably be in hierarchies-react.
> > 
> > 
> > This doesn't answer the question regarding shadow.
> 
> ~The shadow comes from kiwi side, will file an issue for it.~ Will fix
> 
> > > Added an actual Seperator component, it is less visible in dark mode, because of incorrect background color.
> > 
> > 
> > Looks better, thanks. Why is there are space between the separator and the first node?
> 
> In the mockups there is an additional padding after Seperator.
> 
> > > Ps. Adjusted paddings to reflect mockups better, this fixed the focus on selector/search problem on our end.
> > 
> > 
> > I think there's still a problem with outline on header buttons - the outline is displayed on top of sibling buttons. I think it's because we don't have any spacing between them.
> 
> This could be slightly tricky, because we allow buttons to be provided therefor, we can't add margins to the buttons itself. We could have something like this `.container-name button { margin-right: var(4px equivalent)} @saskliutas I'm not sure if this would be an appropriate solution as it could cause issues in the future. It would be best if kiwi would apply this focus border inside the component instead of outside.

I think ideally gaps between buttons would be handled by ButtonGroup component. For now you can put buttons in a flex container and set gap.

MartynasStrazdas:
Changed the focus mode do be disabled on search open. Added a gap for the buttons. For the select shadow waiting for a response from kiwi as the "outline" variant is currently broken in dark mode. 

https://github.com/user-attachments/assets/26163dfa-7459-42df-815c-77ca65cc3397

Ps. will have loading fixed in a separate pr

packages/itwin/tree-widget/src/tree-widget-react/components/tree-header/TreeHeader.tsx
```diff
@@ -5,67 +5,33 @@
...
   );
 }
 
-interface HeaderButtonsProps {
-  contracted: boolean;
-  size: "large" | "small";
-}
-
-function HeaderButtons(props: PropsWithChildren<HeaderButtonsProps>) {
+function HeaderButtons(props: PropsWithChildren) {
```
saskliutas, line 35: Looks like redundant

packages/itwin/tree-widget/src/tree-widget-react/components/tree-header/TreeHeader.tsx
```diff
@@ -5,67 +5,33 @@
...
+  const { className, children } = props;
 
   return (
-    <div className={classnames("tree-widget-tree-header", className)}>
-      <HeaderButtons contracted={isSearchOpen} size={size}>
-        {children}
-      </HeaderButtons>
-    </div>
+    <>
+      <div className={classnames("tw-tree-header", className)}>
```
saskliutas, line 27: Maybe this should be renamed to something like `tw-tree-tools`/`tw-tree-toolbar`. I also think it would be nice to avoid calling it TreeHeader as it might cause confusion with TreeWidgetHeader. Maybe something like `TreeToolbar`? @grigasp

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/TreeRenderer.tsx
```diff
@@ -25,7 +25,10 @@ export function TreeRenderer({ rootNodes, onNodeClick, expandNode, visibilityBut
         onNodeClick={onNodeClick}
         expandNode={expandNode}
         rootNodes={rootNodes}
-        actionsRenderer={visibilityButtonProps ? (node) => <TreeItemVisibilityButton {...visibilityButtonProps} node={node} /> : undefined}
+        actions={[
+          ...(visibilityButtonProps ? [createVisibilityAction(visibilityButtonProps)] : []),
```
saskliutas, line 29: Can we move code related to visibility buttons into `VisibilityTreeRenderer`?

saskliutas (changes requested):
I saw that you added `.css` file instead of `.scss`. I think we should be consistent and stick to one format. We do not have much css so I am fine with switching to `.css` instead of `.scss`.

packages/itwin/tree-widget/src/tree-widget-react/components/tree-header/TreeHeader.tsx
grigasp: agree, toolbar sounds better

MartynasStrazdas: Changed to toolbar

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/TreeRenderer.tsx
MartynasStrazdas: The only issue would be that filter button would go first then visibility. If this order is fine, have the changes ready.

saskliutas: I think we could duplicate filter creation in `VisibilityTreeRenderer`. This might result in use having 3 components: `TreeRenderer`, `VisibilityTreeRenderer` and `BaseTreeRenderer` (internal) but would allow to keep separate API for visibility and non-visibility trees. It could look like:

```tsx
function BaseTreeRenderer(props) {
  // setup and render tree
  return <TreeComp
    actions={props.actions}
  />
}

function TreeRenderer(props) {
  return <BaseTreeRenderer 
    actions={]createFillterAction(), ...props.consumerActions]}
  />;
}

function VisibilityTreeRenderer(props) {
  return <BaseTreeRenderer 
    actions={[createVisibilityAction(), createFilterAction(), ...props.consumerActions]}
  />
}
```

MartynasStrazdas: Changed and added a way to provide actions to all base trees, which was missing.

packages/itwin/tree-widget/api/tree-widget-react.api.md
```diff
@@ -33,6 +33,12 @@ import type { useSelectionHandler } from '@itwin/presentation-hierarchies-react'
 import type { Viewport } from '@itwin/core-frontend';
 import type { Widget } from '@itwin/appui-react';
 
+// @internal (undocumented)
+function BaseTreeRenderer({ rootNodes, onNodeClick, expandNode, actions, ...props }: BaseTreeRendererProps): JSX_2.Element;
```
saskliutas, line 37: These should not be exported.

packages/itwin/tree-widget/api/tree-widget-react.api.md
```diff
@@ -421,27 +389,48 @@ export class TreeWidget {
 }
 
 // @public
-export function TreeWidgetComponent(props: SelectableTreeProps): JSX_2.Element;
+export function TreeWidgetComponent(props: TreeWidgetComponentProps): JSX_2.Element | null;
```
saskliutas, line 392: I reallyl don't like the naming with all those `TreeWidgetSomething`. Maybe we could do something like:
`TreeWidgetComponent` -> `TreeSelector` (probably could make it internal)
`TreeWidgetComponentDefinition` -> `TreeDefinition`
`TreeWidgetComponentProvider` -> `TreeWidgetComponent` (revert to the previous name)

saskliutas (changes requested):

saskliutas (approved):

PR #1193, opened 2025-02-18
Remove Deprecated from map-layers Plus Bug Fixes

pachhaibipin:
### PR Description:
- [x] Replaced ImageCheckbox in SubLayersTree component with Checkbox.
       **Before:** 
      image (6)
       **After:** 
     checkbox

- [x] Dropped deprecated `ResizableContainerObserver` and introduced hook `useResizeObserver` to support same functionality. See the file `useResizeObserver.tsx` for details.

- [x] Replaced deprecated `NumberInput `with iTwinUI `Input` (for Exaggeration), to fit previous NumberInput functionality
       **Before:** 
      image (7)
     **After:** 
      Input
      
- [x] Created a scrollable div with `overflow: auto` and children component inherits the height from parent. 
Problem: The map manager settings popover had some inside elements going outside the bounds of the popover
       Input
       
- [x] Introduced customized hooks- `useCrossOriginPopup` and `useInterval` and their corresponding tests to support existing functionalities and remove deprecated ones. See the files for detailed references

- [x] Replaced/Removed all deprecated code


packages/itwin/map-layers/src/test/MapManagerSettings.test.tsx
```diff
@@ -413,7 +413,7 @@ describe("MapManagerSettings", () => {
 
   it("exaggeration", () => {
     const component = mountComponent();
-    const numericInputs = component.find(NumberInput);
+    const exaggerationInput = component.find(Input).last();;
```
hl662, line 416: I am very hesitant to rely on finding an array of inputs and using the 'last' input found, it's not foolproof. I would recommend adding a data-testid attribute to the Input component you are looking for. See [docs](https://testing-library.com/docs/queries/bytestid/) ref. I've also used data-testid in a previous PR for map-layers, so you can also find code examples within here

packages/itwin/map-layers/src/ui/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,47 @@
...
+import * as React from "react";
+
+// This file is inherited from the itwinui-useResizeObserver.tsx
+// https://github.com/iTwin/iTwinUI/blob/87bbc99316c17bd2b763a730135fbb45c84b0e82/packages/itwinui-react/src/utils/hooks/useResizeObserver.tsx
+
+const getWindow = () => {
+  return typeof window === "undefined" ? undefined : window;
+};
+
+export const useResizeObserver = <T extends HTMLElement>(
```
hl662, line 14: is the resize observer important to our component's functionality? We can't drop it?

pachhaibipin: > is the resize observer important to our component's functionality? We can't drop it?

I don't think we can drop it completely since it's being used where child component needs variable `height `and `width` being passed as props.
For example, Inside `FeatureInfoWidget`:

```
const [elementRef] = useResizeObserver<HTMLDivElement>((size) => {
   handleResize(size.width, size.height);
 });
 
 if (hasData && dataProvider.current) {
    return (
      <div ref={elementRef}>
        <VirtualizedPropertyGridWithDataProvider
          width={width}
          height={height}
          actionButtonRenderers={[copyButton]}
        />
      </div>
```

packages/itwin/map-layers/src/test/MapManagerSettings.test.tsx
```diff
@@ -10,9 +10,8 @@ import * as sinon from "sinon";
 import * as moq from "typemoq";
 import { PlanarClipMaskMode, PlanarClipMaskPriority, TerrainHeightOriginMode } from "@itwin/core-common";
 import { MockRender } from "@itwin/core-frontend";
-import { NumberInput } from "@itwin/core-react";
 import { QuantityNumberInput } from "@itwin/imodel-components-react";
```
hl662, line 13: with the removal of deprecated "@itwin/core-react" components in map-layers, do you think we can remove it from package.json? Less dependencies to handle and etc. I'd be curious if we need any of the css imports either

pachhaibipin: > with the removal of deprecated "@itwin/core-react" components in map-layers, do you think we can remove it from package.json? Less dependencies to handle and etc. I'd be curious if we need any of the css imports either

I doubt it's possible. There are lots of css dependencies inherited from core-react itself if I've understood the css inheritance correctly. I tried removing them from css, looked weird on viewer. Can you verify? For example, `MapLayerManager.scss` has lots of those dependencies, not sure if we can get rid of that. Few examples there would be:

```
@import "~@itwin/core-react/lib/cjs/core-react/index";

$divider-color: $buic-background-toolbutton-stroke;
$hovered-icon-color: --iui-color-background-accent;
$default-icon-height: 20px;
$default-font-size: --iui-font-size-1;

.map-manager-item-sub-layer-container {
  @include uicore-z-index(dialog-popup);
}
.map-layer-settings-popup-panel {
  color: $buic-foreground-body;
  font-size: $default-font-size;
  display: grid;
  grid-row-gap: 4px;
  padding: 4px;

  &:hover {
    color: $buic-foreground-activehover;
    cursor: auto;
  }
```

hl662: okay, removing core react will be more of a nice to have for now then, lets get this PR active without it

packages/itwin/map-layers/src/ui/widget/SubLayersTree.tsx
```diff
@@ -183,16 +194,7 @@ export function SubLayersTree(props: SubLayersTreeProps) {
...
-          {props.width === undefined && props.height === undefined && (
-            // eslint-disable-next-line @typescript-eslint/no-deprecated
-            <ResizableContainerObserver
-              onResize={(w, h) => {
-                setWidth(w);
-                setHeight(h);
-              }}
-            />
-          )}
+        <div className="map-manager-sublayer-tree-content" ref={elementRef}>
```
aruniverse, line 197: do we need all the checks on L198 now?

aruniverse (commented):
lgtm; youre missing rush change

packages/itwin/map-layers/src/ui/widget/SubLayersTree.tsx
pachhaibipin: `height `and `width `are being passed as props to `SubLayersTree `as `number | undefined` but` ControlledTree`  takes number only. On top of that, we've ResizeObserver to observe height and width so I kept it as it was for safety instead of making many changes on props (type) and observer

pachhaibipin:
> lgtm; youre missing rush change

done. Since it's another patch release, type should be none, right?

aruniverse (approved):

Issue #1195, opened 2025-02-19
Tree widget: Support custom "empty hierarchy" rendering


grigasp:
https://www.figma.com/design/MuSiGw5reQvOJEhmi6arq5/%5Bbntly%5D-treeView-%5Bv2%5D?node-id=3243-88801&t=aR8bTWtTYhNGMQMP-1
![Image](https://github.com/user-attachments/assets/113314c3-a442-4930-a67b-599e99f94610)

We should let consumers supply their own "empty hierarchy" renderer.

PR #1214, opened 2025-02-24
Tree widget: Allow custom no data message renderer

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/1195

Added ability to supply custom no data renderer. Also updated default no data renderer to renderer icon and added some padding.

ModelsTree
CategoriesTree
IModelContentTree
ExternalSourcesTree


grigasp (commented):
The only nitpick I have is that I'd like `noDataRenderer` name better than `noDataMessage`. What do you think?

saskliutas:
> The only nitpick I have is that I'd like `noDataRenderer` name better than `noDataMessage`. What do you think?

What about `emptyTreeContent`? `*Renderer` prop name for me associates with render functions like `(props: PropType) => React.ReactNode`. But I don't think in this case we need to take render functions and it is enough to take `React.ReactNode`

grigasp:
> What about `emptyTreeContent`?

Sounds good

grigasp (approved):

Issue #1196, opened 2025-02-19
Tree widget: Root level error state


grigasp:
https://www.figma.com/design/MuSiGw5reQvOJEhmi6arq5/%5Bbntly%5D-treeView-%5Bv2%5D?node-id=3243-89276&t=aR8bTWtTYhNGMQMP-1
![Image](https://github.com/user-attachments/assets/45121a7d-1f7a-4e5d-b694-b34e183ea74f)

PR #1217, opened 2025-02-24
Tree widget: Render error message on tree load error

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/1196

Added error boundary to render a fallback component when tree error is encountered.
image

@grigasp Should the same fallback component be used if error is encountered higher in the component tree (e.g. tree selector)?


grigasp:
> @grigasp Should the same fallback component be used if error is encountered higher in the component tree (e.g. tree selector)?

@ahilhorst, for the case when the loading available trees list fails, do you think is would be fine? Maybe with "widget" instead of tree.
image

ahilhorst:
hey @grigasp, yep that's fine. imho we can just keep "tree". thanks

MartynasStrazdas (approved):

PR #1219, opened 2025-02-25
Tree widget: Revive e2e tests

saskliutas:
~~Removed `Density` toggle from test-viewer as tree currently do not support it.~~
Removed test for different densities.
Revived e2e tests and produced new screenshots.


saskliutas:
> > Removed `Density` toggle from test-viewer as tree currently do not support it.
> 
> Do you think that's necessary? IMO, we could keep it and simply not use it in the tests for the time being.

I guess. We can keep the button. Added it back.

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
```diff
@@ -5,27 +5,24 @@
...
+    await treeWidget.getByRole("combobox").click();
+    await takeScreenshot(page, treeWidget);
+  });
 
-    test("tree selector badge", async ({ page }) => {
-      await selectTree(treeWidget, "External sources");
-      await page.getByText("The data required for this tree layout is not available in this iModel.").waitFor();
-      await takeScreenshot(page, treeWidget);
-    });
+  test("tree selector badge", async ({ page }) => {
```
grigasp, line 23: The "before" test targeted the "badge" / "icon" of the `Select` item. This one doesn't show it - I guess we don't support icons on `Select` items at the moment? I would be great to make it clear, because it's very confusing to have a test that succeeds on a component that doesn't work as expected. IMO, we need a comment before the test is necessary to explain that. Also, I'm debating whether we should add `.skip` on it or not 🤔

packages/itwin/tree-widget/src/e2e-tests/TreeWidget.test.ts
```diff
@@ -5,27 +5,24 @@
...
+import { initTreeWidgetTest, locateNode, selectTree, takeScreenshot } from "./utils.js";
 
-test.describe.skip("Widget", () => {
+test.describe("Widget", () => {
   let treeWidget: Locator;
 
   test.beforeEach(async ({ page, baseURL }) => {
     treeWidget = await initTreeWidgetTest({ page, baseURL });
-    await locateNode(treeWidget, "BayTown").getByRole("checkbox", { name: "Visible: All models are visible", exact: true }).waitFor();
+    await locateNode(treeWidget, "ProcessPhysicalModel").waitFor();
```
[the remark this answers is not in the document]
saskliutas: Shouldn't this handle it https://github.com/iTwin/viewer-components-react/blob/09086eb8d882bc37e1a0f3cb7754a9b5ad8d766b/packages/itwin/tree-widget/src/e2e-tests/utils.ts#L47?

packages/itwin/tree-widget/src/e2e-tests/CategoriesTree.test.ts
```diff
@@ -6,212 +6,148 @@
...
-      await takeScreenshot(page, node, { boundingComponent: treeContainer, expandBy: { top: 10, bottom: 10 } });
+    await page.getByRole("button", { name: "Apply" }).click();
 
-      // shrink panel
-      await expandStagePanel(page, "right", -100);
-      // scroll to the right side
-      await scrollTree(page, 10000, 0);
+    // wait for filtered children to appear
+    await locateNode(treeWidget, "Equipment - Insulation").waitFor();
+    await treeWidget.getByRole("button", { name: "Apply filter" }).waitFor();
```
grigasp, line 58: do we even need this? I think the widget has a numer of such buttons, so I'm not sure this serves any purpose

grigasp (approved):

MartynasStrazdas (approved):

PR #1225, opened 2025-02-27
Tests Migration to Use Vitest

pachhaibipin:
- Migrated Map-Layers tests from Mocha/Chai to use Vitest
- Removed Mocha/Chai and their dependencies 
- Use V8 coverage instead of nyc
-  No longer using enzyme for component rendering


packages/itwin/map-layers/vitest.config.mts
```diff
@@ -0,0 +1,28 @@
...
+        "**/*.d.ts",
+        "**/*.d.tsx"
+      ],
+      reporter: [
+        "text-summary",
+        "lcov",
+        "cobertura"
+      ],
+      reportsDirectory: "./lib/cjs/test/coverage",
+    }
```
hl662, line 26: can you add [minWorkers](https://vitest.dev/config/#minworkers) and maxWorkers to the config? Look at the vitest configs in core, they can match those numbers. We just want to avoid resource starvation

packages/itwin/map-layers/src/test/useInterval.test.tsx
```diff
@@ -2,40 +2,30 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import { expect } from "chai";
-import sinon from "sinon";
 import { renderHook } from "@testing-library/react";
-import { useInterval } from "../ui/hooks/useInterval";
+import { useInterval } from "../../src/ui/hooks/useInterval";
```
hl662, line 6: why change the file path? doesn't seem like it has to be done

packages/itwin/map-layers/src/test/useInterval.test.tsx
```diff
@@ -2,40 +2,30 @@
...
-    clock = sinon.useFakeTimers();
-  });
-
-  afterEach(() => {
-    clock.restore();
-  });
-
   it("should call interval's callback when timeout is reached", () => {
-    const spy = sinon.spy();
+    vi.useFakeTimers();
```
hl662, line 10: after using fake timers, make sure you reset them ([ref](https://vitest.dev/api/vi.html#vi-userealtimers)). If you have two tests that will utilize these timers, I'd call these in a beforeEach/afterEach test hook

hl662: oh, i see that you've implemented them in the setup oops. But should that be in there? seems like fake timers are used in only a few places

mdastous-bentley (approved):

change/@itwin-map-layers-e4e0f3ca-b87b-4fe2-aa2c-57a03c83157d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Test Migration to Vitest",
```
hl662, line 3: Update the comment, the testing was good and we've deemed the migration viable

aruniverse: the whole purpose of these change files is to communicate to the user changes to the api, they are supposed to be user facing. changing the internal test framework doesnt need to be shared explicitly via users in the changelog

```suggestion
  "comment": "",
```

hl662 (approved):

packages/itwin/map-layers/vitest.config.mts
```diff
@@ -0,0 +1,30 @@
+import { coverageConfigDefaults, defineConfig } from "vitest/config";
```
aruniverse, line 1: shouldnt this file include the same changes you made in another pr recently to clean out the cache?

pachhaibipin: map-layers doesn't have any `peerDeps` and direct `deps` from workspace so not needed

aruniverse: ah my bad, was looking at wrong repo

PR #1227, opened 2025-02-28
Update map-layers to use AppUi 5.0

### LINKED ISSUE #1188: update map-layers to appui@5
Update map-layers to use AppUi 5.0

As long as we are not using deprecated APIs removed in appUI 5, we can add || ^5.0.0 to the [peers](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/map-layers/package.json#L107)

Would prefer to not support both AppUI 4.x and 5.x. Consult with consumers if they still need to support AppUI 4.x

pachhaibipin:
- Closes #1188


change/@itwin-map-layers-def69615-32a5-48bc-bde7-448853d1192b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "appUIv5 usage",
```
hl662, line 3:
```suggestion
  "comment": "Add support for AppUI 5.x",
```

hl662:
Hmm we should look into centering this if possible. If adding a custom map layer takes too quick for you to see the loading spinner, I'd open up dev tools and artificially throttle the network and CPU. Or the better thing would be to just set the conditional that renders that spinner to always true while fixing the position.
image

mdastous-bentley (approved):

mdastous-bentley (approved):

pachhaibipin:
> Hmm we should look into centering this if possible. If adding a custom map layer takes too quick for you to see the loading spinner, I'd open up dev tools and artificially throttle the network and CPU. Or the better thing would be to just set the conditional that renders that spinner to always true while fixing the position. image

Introduced little bit of div styling.
Result:
image (5)

hl662 (approved):

PR #1232, opened 2025-03-06
Tree widget & Property grid: Update core imports

saskliutas:
Update imports to import directly from `@itwin/core-bentley` instead of `@itwin/core-common` that just reexports from `@itwin/core-bentley`.


grigasp (approved):

aruniverse:
we should also make the same tree-widget changes in `tree/next`

grigasp:
> we should also make the same tree-widget changes in `tree/next`

Agree. The changes will be carried over to that branch when we merge `master` into it - we already did that once and the plan is to keep doing that regularly.

Issue #1190, opened 2025-02-18
Unit tests experiencing memory issue when using testdouble


JonasDov:
Tests that are using `testdouble` don't properly clean up after themselves (`td.reset()` doesn't help), this causes `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`. 

Noticed this when running `tree-widget` tests, but it also affects `property-widget` tests as it uses `testdouble` in a similar way. `process.memoryUsage().heapUsed` was used to determine memory usage. Before measuring usage garbage was collected (`global.gc?.()`). Measured usage before any tests were executed and after all of them were executed, these were the results:
With tests that use `testdouble`:
- Before: `~152 MB`
- After: `~960 MB`

Without tests that use `testdouble`:
- Before: `~152 MB`
- After: `~162 MB`

This is probably related to https://github.com/testdouble/testdouble.js/issues/534

PR #1233, opened 2025-03-06
Tree & Property widgets: Drop usage of `testdouble`

grigasp:
Closes https://github.com/iTwin/viewer-components-react/issues/1190.

With the changes, running the whole test suite bumps memory consumption by:
- property widget: `13.497.104 bytes`
- tree widget: `15.031.064 bytes`

Ideally, we should avoid module mocking as much as possible. For property widget I managed to do that, for tree widget - in some cases.


packages/itwin/property-grid/src/property-grid-react/PropertyGridUiItemsProvider.tsx
```diff
@@ -113,34 +116,56 @@ export type PropertyGridWidgetProps = PropertyGridComponentProps & {
...
-function PropertyGridWidget({ shouldShow, widgetId, ...props }: PropertyGridWidgetProps) {
+/**
+ * Props for `createPropertyGrid`.
+ * @public
+ */
+export type PropertyGridWidgetProps = PropertyGridComponentProps & PropertyGridWidgetOwnProps;
+
+/** @internal */
+type PropertyWidgetInternalProps = PropertyGridWidgetOwnProps & {
+  PropertyGridComponent: ReactNode;
```
saskliutas, line 149: We should be consistent with property names.
```suggestion
  propertyGridComponent: ReactNode;
```

I think it should be in PascalCase only if you take it as `React.ComponentType<SomeProps>` and use it like
```tsx
interface Props {
    OtherComponent: React.ComponentType<SomeProps>
}

function MyComponent({ OtherComponent }: Props) {
    return <div>
        <OtherComponent {...someProps} />
    </div>
}
```

packages/itwin/property-grid/src/test/PropertyGridUiItemsProvider.test.tsx
```diff
@@ -106,43 +65,57 @@ describe("PropertyGridUiItemsProvider", () => {
...
-      "",
-      appuiReactModule.StageUsage.General,
-      appuiReactModule.StagePanelLocation.Right,
-      appuiReactModule.StagePanelSection.End,
-    );
-    render(<>{widget.content}</>);
-
-    expect(propertyGridComponentStub).to.be.called;
+describe("createPropertyGrid", () => {
+  const testPropertyGridComponent = <>Test PropertyGridComponent</>;
```
saskliutas, line 72:
```suggestion
  function TestPropertyGridComponent() {
    return <>Test PropertyGridComponent</>;
  }
```

packages/itwin/property-grid/package.json
```diff
@@ -46,7 +46,7 @@
     "lint:stylelint": "stylelint \"./src/**/*.scss\"",
     "lint:fix": "npm run lint:eslint -- --fix && npm run lint:stylelint -- --fix",
     "pseudolocalize": "betools pseudolocalize --englishDir ./public/locales/en --out ./lib/public/locales/en-PSEUDO",
-    "test:dev": "cross-env NODE_OPTIONS=\"--import ../../../node-hooks/ignore-styles/register.cjs\" mocha --enable-source-maps --config ./.mocharc.json",
+    "test:dev": "node --enable-source-maps --import ../../../node-hooks/ignore-styles/register.cjs ./node_modules/mocha/bin/mocha.js --config ./.mocharc.json",
```
saskliutas, line 49: Can we run tests like this?
```suggestion
    "test:dev": "cross-env NODE_OPTIONS=\"--enable-source-maps --import ../../../node-hooks/ignore-styles/register.cjs\" mocha --config ./.mocharc.json",
```

packages/itwin/property-grid/src/property-grid-react/components/MultiElementPropertyGrid.tsx
```diff
@@ -46,6 +46,12 @@ export interface MultiElementPropertyGridProps extends Omit<PropertyGridProps, "
 
   /** Renders controls for ancestors navigation. If set to `undefined`, ancestors navigation is disabled. */
   ancestorsNavigationControls?: (props: AncestorsNavigationControlsProps) => ReactNode;
+
+  /**
+   * An optional callback to get parent instance key of the given instance. Defaults
+   * to using `computeSelection` from `@itwin/unified-selection`.
+   */
+  getParentInstanceKey?: (key: InstanceKey) => Promise<InstanceKey | undefined>;
```
saskliutas, line 54: I think it would be worth mentioning that this callback should be memoized.

saskliutas (changes requested):

packages/itwin/property-grid/src/property-grid-react/PropertyGridUiItemsProvider.tsx
grigasp: changed

packages/itwin/property-grid/src/test/PropertyGridUiItemsProvider.test.tsx
grigasp: applied suggestion

packages/itwin/property-grid/package.json
grigasp: yes, the change is only necessary if we use node mocks, which we don't for property grid tests

saskliutas (approved):

aruniverse (approved):

PR #1241, opened 2025-03-14
[Tree-widget]: update alpha deps for release

MartynasStrazdas:


grigasp (dismissed):

grigasp (commented):
could we focus instead of click the node we want to expand?

packages/itwin/tree-widget/src/e2e-tests/ModelsTree.test.ts
```diff
@@ -57,13 +60,16 @@ test.describe("Models tree", () => {
...
   test("expanded tree node", async ({ page }) => {
     const physicalModelNode = locateNode(treeWidget, "ProcessPhysicalModel");
-    await physicalModelNode.getByRole("button", { name: "Toggle", includeHidden: true }).click();
+    await physicalModelNode.click();
+    await page.keyboard.press("ArrowRight");
 
     const pipeSupportNode = locateNode(treeWidget, "PipeSupport");
-    await pipeSupportNode.getByRole("button", { name: "Toggle", includeHidden: true }).click();
+    await pipeSupportNode.click();
+    await page.keyboard.press("ArrowRight");
```
saskliutas, line 68: Can this be extracted into utility function like `expandTreeNode()`?

MartynasStrazdas: Good call, added utility function and changed to use focus instead of click.

grigasp (approved):

Issue #1230, opened 2025-03-05
Tree widget: Elements in Categories tree


grigasp:
Requirements:
1. Ability to opt-out from showing sub-category nodes.
2. Ability to opt-in for showing geometric elements under category nodes. Initially we're only showing root elements to avoid problems described in:
   - https://github.com/iTwin/viewer-components-react/issues/1018
   - https://github.com/iTwin/viewer-components-react/issues/1100
   - https://github.com/iTwin/viewer-components-react/issues/1129

PR #1247, opened 2025-03-18
Add option to not show SubCategories in categories tree

JonasDov:
Related to #1230

### DIFF
--- a/packages/itwin/tree-widget/api/tree-widget-react.api.md
+++ b/packages/itwin/tree-widget/api/tree-widget-react.api.md
@@ -48,10 +48,8 @@ export const CategoriesTreeComponent: {
 };
 
 // @public (undocumented)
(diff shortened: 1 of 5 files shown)


grigasp (changes requested):
Let's do this through a `hierarchyConfig` prop, similar to what we have in Models tree: https://github.com/iTwin/viewer-components-react/blob/8abc14f068a91fe6ebc39a3a3056f2db84bc6e58/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/UseModelsTree.tsx#L39

JonasDov:
> Let's do this through a `hierarchyConfig` prop, similar to what we have in Models tree:
> 
> https://github.com/iTwin/viewer-components-react/blob/8abc14f068a91fe6ebc39a3a3056f2db84bc6e58/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/UseModelsTree.tsx#L39

Added hierarchyConfig

grigasp (approved):

PR #1253, opened 2025-03-19
Fix modeled element having children when it does not

### LINKED ISSUE #1250: Tree widget: Models tree shows node as having children, but it doesn't
1. Open http://localhost:3000/?iTwinId=1b9e8c1f-d497-4e78-adc2-e97896af9fd5&iModelId=bbfff4fd-60ab-49f6-b6cc-7c40d5aacddc
2. Expand the nodes according to the below screen capture - notice that a couple of nodes show up as having children, but the children don't load when the nodes are expanded. Also, it's not clear why visibility is set to "partial".

Image

JonasDov:
Closes #1250


packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -257,13 +258,14 @@ export class ModelsTreeIdsCache {
...
   private async *queryModelCategories() {
     const query = /* sql */ `
-      SELECT Model.Id modelId, Category.Id categoryId
-      FROM ${this._hierarchyConfig.elementClassSpecification}
-      WHERE Parent.Id IS NULL
-      GROUP BY modelId, categoryId
+      SELECT this.Model.Id modelId, this.Category.Id categoryId, m.IsPrivate isPrivate
+      FROM BisCore.Model m
+        JOIN ${this._hierarchyConfig.elementClassSpecification} this ON m.ECInstanceId = this.Model.Id
+      WHERE this.Parent.Id IS NULL
```
grigasp, line 264:
```suggestion
      FROM BisCore.Model m
      JOIN ${this._hierarchyConfig.elementClassSpecification} this ON m.ECInstanceId = this.Model.Id
      WHERE this.Parent.Id IS NULL
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -272,19 +274,31 @@ export class ModelsTreeIdsCache {
...
+          SELECT false
+          FROM (
+            SELECT ECInstanceId id, IsPrivate isPrivate FROM BisCore.Model
+          )
+          WHERE id = pe.ECInstanceId AND isPrivate = false
+          LIMIT 1
+        ), true) isModeledElementPrivate
       FROM BisCore.Model m
       JOIN ${this._hierarchyConfig.elementClassSpecification} pe ON pe.ECInstanceId = m.ModeledElement.Id
+      JOIN ${this._hierarchyConfig.elementClassSpecification} ce ON ce.Model.Id = m.ECInstanceId
```
grigasp, line 288: why is this join needed? if you're using it to filter models with elements - you'll need to find another way, because this one will create a row for each element in the model.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -272,19 +274,31 @@ export class ModelsTreeIdsCache {
...
-        pe.Model.Id modelId
+        pe.Model.Id modelId,
+        IFNULL((
+          SELECT false
+          FROM (
+            SELECT ECInstanceId id, IsPrivate isPrivate FROM BisCore.Model
+          )
+          WHERE id = pe.ECInstanceId AND isPrivate = false
+          LIMIT 1
+        ), true) isModeledElementPrivate
```
grigasp, line 285: can't we just use `m.IsPrivate` here?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -272,19 +274,31 @@ export class ModelsTreeIdsCache {
...
   }
 
   private async getModelWithCategoryModeledElements() {
     this._modelWithCategoryModeledElements ??= (async () => {
       const modelWithCategoryModeledElements = new Map<Id64String, Id64Set>();
-      for await (const { modelId, categoryId, modeledElementId } of this.queryModeledElements()) {
+      for await (const { modelId, categoryId, modeledElementId, isPrivate } of this.queryModeledElements()) {
+        if (isPrivate) {
+          continue;
+        }
```
grigasp, line 301: we should filter out with query so we don't request unnecessary stuff from the backend just to drop it here

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -257,13 +258,14 @@ export class ModelsTreeIdsCache {
...
-      GROUP BY modelId, categoryId
+      SELECT this.Model.Id modelId, this.Category.Id categoryId, m.IsPrivate isPrivate
+      FROM BisCore.Model m
+        JOIN ${this._hierarchyConfig.elementClassSpecification} this ON m.ECInstanceId = this.Model.Id
+      WHERE this.Parent.Id IS NULL
+      GROUP BY modelId, categoryId, isPrivate
     `;
     for await (const row of this._queryExecutor.createQueryReader({ ecsql: query }, { rowFormat: "ECSqlPropertyNames", limit: "unbounded" })) {
-      yield { modelId: row.modelId, categoryId: row.categoryId };
+      yield { modelId: row.modelId, categoryId: row.categoryId, isPrivate: !!row.isPrivate };
```
grigasp, line 268: maybe `isPrivate` should be renamed to `isModelPrivate` to make it clear what's private? because categories, I think, also have this property

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -461,8 +473,14 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
...
-                        UNION ALL
-                        SELECT ModeledElement.Id ParentId FROM bis.GeometricModel3d
+                        ${
+                          modeledElements.length === 0
+                            ? ""
+                            : `
+                              UNION ALL
+                              SELECT ModeledElement.Id ParentId FROM BisCore.GeometricModel3d WHERE ModeledElement.Id IN (${modeledElements.join(",")})
+                            `
+                        }
```
grigasp, line 483: do we need to run this query at all, if we know it's going to return something? it looks like `hasChildren` should be set to true rather than `{ selector: ... }` if `modeledElements` is not empty. also, it seems we don't need `modeledElements` to be an array - a boolean is enough.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -522,8 +548,14 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
...
-                        UNION ALL
-                        SELECT ModeledElement.Id ParentId FROM bis.GeometricModel3d
+                        ${
+                          hasSubModel
+                            ? `
+                              UNION ALL
+                              SELECT ModeledElement.Id ParentId FROM bis.GeometricModel3d WHERE IsPrivate = false
+                            `
+                            : ""
+                        }
```
grigasp, line 558: this is similar to my comment in `createSpatialCategoryChildrenQuery` - I don't think we need this query for `hasChildren` if we already know the answer when creating the query.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
JonasDov: Adjusted the query, it now does not use this `JOIN`. Instead it creates a property `hasChildren` like so:
```ts
IFNULL((
  SELECT true
  FROM ${this._hierarchyConfig.elementClassSpecification} ce
  WHERE ce.Model.Id = pe.ECInstanceId
  LIMIT 1
), false) hasChildren
```
And filters only those values that have children

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
JonasDov: Yes, changed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
JonasDov: Renamed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
JonasDov: Changed, now if it has subModel, then hasChildren is set to `true`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
grigasp: seems overly complicated.. why not this?

```sql
WHERE m.ECInstanceId IN (
  SELECT Model.Id
  FROM ${this._hierarchyConfig.elementClassSpecification}
)
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -468,6 +489,14 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
       filter: instanceFilter,
       contentClass: { fullName: this._hierarchyConfig.elementClassSpecification, alias: "this" },
     });
+    let hasSubModel = false;
```
grigasp, line 492: is it just me, or is this easier to read?

```ts
const hasSubModel = await firstValueFrom(
  from(elementIds).pipe(
    mergeMap(async (elementId) => this._idsCache.hasSubModel(elementId)),
    filter((has) => has),
    defaultIsEmpty(false),
    take(1),
  )
);
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
grigasp: I see what you mean... Could we avoid the duplication in query by doing something like this?

```ts
hasChildren: {
  selector: `
    IIF(
      ${modeledElements.length ? `this.ECInstanceId IN (${modeledElements.join(",")})` : `FALSE`}
      1,
      IFNULL((
        SELECT 1
        FROM ${this._hierarchyConfig.elementClassSpecification} ce
        WHERE ce.Parent.Id = this.ECInstanceId
        LIMIT 1
      ), 0)
    )
  `
}
```

BTW, doesn't this apply to `createGeometricElement3dChildrenQuery` as well? It looks like we're checking if parent node's element has a submodel and set its child nodes' `hasChildren` flag based on that. Shouldn't we be looking at children' submodel?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -407,6 +407,15 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
...
     });
+    const modeledElements = new Array<Id64String>();
+    const foundModeledElements = await Promise.all(
+      modelIds.map(async (modelId) => {
+        return this._idsCache.getCategoriesModeledElements(modelId, categoryIds);
+      }),
+    );
+    foundModeledElements.forEach((foundElements) => {
+      modeledElements.push(...foundElements);
+    });
```
grigasp, line 418: I think this could be replaced with this:

```ts
const modeledElements = await firstValueFrom(
  from(modelIds).pipe(
    mergeMap(async (modelId) => this._idsCache.getCategoriesModeledElements(modelId, categoryIds)),
    toArray(),
  )
);
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
JonasDov: Changed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
JonasDov: 1. Applied suggestion.
2. You are right about `createGeometricElement3dChildrenQuery` changed how we determine `hasChildren` for them to this:
```ts
IFNULL((
  SELECT 1
  FROM ${this._hierarchyConfig.elementClassSpecification} ce
  JOIN BisCore.Model m ON ce.Model.Id = m.ECInstanceId
  WHERE ce.Parent.Id = this.ECInstanceId OR (ce.Model.Id = this.ECInstanceId AND m.IsPrivate = false)
  LIMIT 1
), 0)
```
3. Added some more tests to `ModelsTreeDefinition.test.ts`:
- `children of child element is set to true when child element has subModel that contains children`
- `children of child element is set to false when it's subModel is private`
- `children of child element is set to false when it's subModel has no children`
- `children of element is set to true when element has subModel that contains children`
- `children of element is set to false when it's subModel is private`
- `children of element is set to false when it's subModel has no children`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
JonasDov: It is not entirely correct what you suggested. Your suggestion returns `Id64Array[]`, but we want to get `Id64Array`. Changed to this:
```ts
const modeledElements = await firstValueFrom(
  from(modelIds).pipe(
    mergeMap(async (modelId) => this._idsCache.getCategoriesModeledElements(modelId, categoryIds)),
    reduce((acc, foundModeledElements) => {
      return acc.concat(foundModeledElements);
    }, new Array<Id64String>()),
  ),
);
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
JonasDov: I removed this part completely

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
grigasp: that's okay. you could've also used `mergeAll` between `mergeMap` and `toArray`

grigasp (approved):

PR #1265, opened 2025-03-26
Add option to show elements in categories tree

JonasDov:
closes #1230

PR #1271, opened 2025-04-01
Fix flaky tests

### LINKED ISSUE #1230: Tree widget: Elements in Categories tree

JonasDov:
closes #1230 
Following https://github.com/iTwin/viewer-components-react/pull/1265#issuecomment-2768171226
Tried a few things to reproduce the issue:
1. Tried running only those two specific tests 400 times, all of them passed.
2. Tried running all the tests in that specific file (5 times in the pipeline) that those two tests are in, all of them passed.
3. Tried running all the tree-widget tests (20 times) in the pipeline. It seems that only on Windows it started to fail (once after 4 tries, other time after 5 times). On Mac and Linux it passed every time.
4. Tried to reproduce this locally, but couldn't.

Currently, `waitFor` is used after node's visibility is changed and is used to wrap a function (`validateHierarchyVisibility`) that validates visibility of all nodes in the hierarchy. Moved `waitFor` to wrap a function (`validateNodeVisibility`) that validates visibility for a single node.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
@@ -29,7 +29,7 @@ import {
   insertSubCategory,
   insertSubModel,
 } from "../../../IModelUtils.js";
-import { TestUtils, waitFor } from "../../../TestUtils.js";
+import { TestUtils } from "../../../TestUtils.js";
 import { createIModelAccess } from "../../Common.js";
 import {
   createCategoryHierarchyNode,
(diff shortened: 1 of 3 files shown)


grigasp (approved):
I don't see how this helps besides adding extra time to the timeout (now the timeout is applied on each node check rather than the whole hierarchy check). So I guess this was a problem of a too small timeout, after all...

PR #1282, opened 2025-04-09
Add child elements category to models tree

### LINKED ISSUE #1100: [Tree widget] Incorrect category used for element visibility in filtered tree
There might be cases where incorrect category is used when determining or changing element visibility in filtered tree. For example ,if we have tree like:
```
Subject
|- Model
  |- Category A
    |- Element A
      |- Element B (from Category B)
```

Filter path to Element B will be:
`Subject -> Model -> Category A -> Element A -> Element B`

At the moment when we try to determine/change visibility of `Element B` we go up in filter path and look for category. In this particular example we will find `Category A` and use it. However, `Element B` is in `Category B` and it should be used instead in all visibility operations.

Need a way to find real element category instead of the one that is shown in tree as an ancestor.

### LINKED ISSUE #1129: Tree widget: Parent/child elements visibility controls in Models tree are inconsistent
1. Open http://localhost:3000/?iTwinId=b391ba44-add7-47a0-8375-f2889a3540e8&iModelId=ba504f88-a479-4156-9d81-658ee169588e in test viewer
2. Expand "ProcessPhysicalModel" model, "E-HEX" category, "Plate-type Heat Exchanger" elements' class grouping node, "E-102 [4-1MG]" element. The "E-102 [4-1MG]" element has its own geometry, but it also has geometric Nozzle child elements.
3. Changing visibility of Nozzle child elements doesn't affect visibility of the parent "E-102 [4-1MG]" element.
4. Changing visibility of "E-102 [4-1MG]" parent element doesn't affect visibility of its children.

Image

How should it work, keeping in mind that parent has its own geometry? The first solution that comes to mind is that when a parent is turned off, its children should also be turned off. But what if a user wants to only turn off the parent, but not the children?

JonasDov:
closes #1100


JonasDov:
Moved implementation to https://github.com/iTwin/viewer-components-react/pull/1716. Closing

PR #1286, opened 2025-04-10
Simplify visibility tooltips

### LINKED ISSUE #1279: Tree widget: Simplify visibility tooltips
Our visibility tooltips are too detailed and don't make much sense to users - need to simplify them.

**Expectation:**
Image
Image
Image

JonasDov:
closes #1279
- Simplified tooltips in Models and Categories tree to `Show` (When hidden or Partial), `Hide` (When Visible), `Disabled` and `Determining visibility...`.


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/Tooltip.ts
```diff
@@ -26,26 +25,13 @@ export function createVisibilityStatus(status: Visibility | "disabled", { useToo
   return {
     state: status === "disabled" ? "hidden" : status,
     isDisabled: status === "disabled",
-    tooltip: useTooltip === false ? undefined : createTooltip(status, useTooltip),
+    tooltip: useTooltip === false ? undefined : createTooltip(status),
```
grigasp, line 28: The `useTooltip` prop is optional, but you're checking it for `false` here. So it looks like callers must specify the prop if they want to omit the tooltip. But everywhere I look the prop is not specified, which means the tooltip is always created.

grigasp (commented):
Could you merge 2 sets of changes using 2 different PRs?
1. Wait for https://github.com/iTwin/viewer-components-react/pull/1284 to merge.
2. Backport it to `tree-widget/next`.
3. Create a tooltips PR on top of it.

ahilhorst:
> > Simplified tooltips in Models and Categories tree to `Visible`, `Hidden` and `Partially visible`.
> 
> According to mockups, they should be: "Show", "Hide" and "Partially visible". Now that I think of it, I'm not sure why two of them are verbs and one is a status. Maybe "Partially visible" should be "Show" instead, since we're going to make it visible if user clicks on it? @ahilhorst, can you comment?

Hey @grigasp good catch. Makes sense to use "Show" indeed, I have updated the mockup to reflect this.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/Tooltip.ts
saskliutas: Do we even need to have `tooltip` on `VisibilityStatus`? With this implementation it could be determined by state when rendering action.

grigasp (approved):

Issue #1283, opened 2025-04-09
Tree widget: Search functionality is enabled for trees that don't support search


grigasp:
![Image](https://github.com/user-attachments/assets/c10527cc-de15-42a9-9da1-f6cc0592864b)

![Image](https://github.com/user-attachments/assets/b2cdf329-023d-457e-9e1d-6bc22dfd5e7d)

PR #1289, opened 2025-04-14
[Tree-widget]: make header search box optional 

MartynasStrazdas:
closes https://github.com/iTwin/viewer-components-react/issues/1283

Preview:
When set to false:
image
When set to true/undefined:
image


packages/itwin/tree-widget/src/tree-widget-react/components/TreeWidgetComponentImpl.tsx
```diff
@@ -40,6 +40,12 @@ export interface TreeDefinition {
    * If callback is `undefined` tree is shown for all iModel connections.
    */
   shouldShow?: (imodel: IModelConnection) => Promise<boolean>;
+  /**
+   *  Boolean flag to determine if the search box should be rendered.
+   *  - false, the search box will not be rendered.
+   *  - true or undefined, the search box will be rendered.
+   */
+  isSearchable?: boolean;
```
grigasp, line 48: Generally, we want `undefined` to match `false`, otherwise it results in very weird API. And I think the default should be to not show the button:

```ts
  /**
   * A flag indicating that this tree supports filtering. In that case, a search button is rendered, allowing users
   * to enter a filtering value and see a filtered tree.
   */
  supportsFiltering?: boolean;
```

grigasp: I think `isFilterable` is also fine. I personally don't like `isSearchable`, but maybe it's just me.

MartynasStrazdas: Because the tree item could have filter action I don't want to include `filter` to not make it confusing.

grigasp: Good point regarding similar names of these two things. However, I think using the term "search" will only add more confusion, since we already have a number of public APIs in this package, where "filter" is used for hierarchy-wide filtering, e.g.:

```ts
// @beta (undocumented)
interface UseModelsTreeProps {
    filter?: string;
    getFilteredPaths?: (props: {
        createInstanceKeyPaths: (props: {
            targetItems: Array<InstanceKey | ElementsGroupInfo>;
        } | {
            label: string;
        }) => Promise<HierarchyFilteringPath[]>;
    }) => Promise<HierarchyFilteringPath[]>;
    // (undocumented)
    onModelsFiltered?: (modelIds: Id64String[] | undefined) => void;
```

I think we should instead rename "filter" to "hierarchy level filter" in `hierarchies-react` package.

apps/test-viewer/src/UiProvidersConfig.tsx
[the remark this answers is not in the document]
MartynasStrazdas: You're right some of the trees don't take filter, removed wrong isSearchable.

packages/itwin/tree-widget/src/tree-widget-react/components/TreeWidgetComponentImpl.tsx
grigasp: Based on the private discussion that we had, we're leaning towards renaming hierarchy-wide filtering to "search", in which case `isSearchable` is okay. I added [a comment](https://github.com/iTwin/presentation/issues/932#issuecomment-2808688778) regarding the need to rename other APIs according to this. So this thread is resolved.

grigasp (approved):

grigasp (approved):

Issue #1239, opened 2025-03-13
Tree widget: "Hierarchy level too large" error message doesn't take into account that parent is not filterable


grigasp:
> The goal is to hide ability to filter. I tried that code and nothing changed for me. Tried on "@itwin/tree-widget-react": "3.5.1".
> 
> ```
> import { isPresentationHierarchyNode, PresentationTreeNode } from "@itwin/presentation-hierarchies-react";
> ```
> ```
> <div className="visibility-panel-tree">
>       <VisibilityTree
>         {...modelsTreeProps}
>         hierarchyLevelSizeLimit={500}
>         noDataMessage={noDataMessage}
>         selectionMode="multiple"
>         density="enlarged"
>         getSchemaContext={getSchemaContext}
>         selectionStorage={FieldApp.getUnifiedSelectionStorage()}
>         imodel={imodel}
>         treeRenderer={(props) => <VisibilityTreeRenderer {...props} {...rendererProps} rootNodes={props.rootNodes.map(disableFiltering)} filterButtonsVisibility="hide" enableVirtualization={MobileCore.isAndroidPlatform /* This fixes scrolling issues on iOS, https://github.com/iTwin/viewer-components-react/issues/1157 */} />}
>       />
>     </div>
>   </>;
> }
> 
> function disableFiltering(node: PresentationTreeNode): PresentationTreeNode {
>   console.log({ node });
>   return isPresentationHierarchyNode(node)
>     ? {
>       ...node,
>       isFilterable: false,
>       children: node.children === true ? true : node.children.map(disableFiltering),
>     }
>     : node;
> }
> ```
> ![Image](https://github.com/user-attachments/assets/9dc732f8-4751-4e01-98f0-97a5ba881849)
> 
> I checked and disableFiltering does get executed and isPresentationHierarchyNode returns true for the node from the picture ("A-DOOR").
> 
> Btw, `VisibilityTree` `hierarchyLevelSizeLimit` seems to not work https://github.com/iTwin/viewer-components-react/issues/1237 

 _Originally posted by @ValdasSorys in [#1236](https://github.com/iTwin/viewer-components-react/issues/1236#issuecomment-2720894391)_

ValdasSorys:
@grigasp Was this released (on 3.6.1 I still reproduce this)?

saskliutas:
[...]

Issue #1237, opened 2025-03-13
Tree widget: VisibilityTree hierarchyLevelSizeLimit prop doesn't work


ValdasSorys:
Nodes are still rendered as if `VisibilityTree` `hierarchyLevelSizeLimit` is 1000 even though `hierarchyLevelSizeLimit` is set to some value.

```
<VisibilityTree
        {...modelsTreeProps}
        hierarchyLevelSizeLimit={5}
        noDataMessage={noDataMessage}
        selectionMode="multiple"
        density="enlarged"
        getSchemaContext={getSchemaContext}
        selectionStorage={FieldApp.getUnifiedSelectionStorage()}
        imodel={imodel}
        treeRenderer={(props) => <VisibilityTreeRenderer {...props} {...rendererProps} rootNodes={props.rootNodes.map(disableFiltering)} filterButtonsVisibility="hide" enableVirtualization={MobileCore.isAndroidPlatform /* This fixes scrolling issues on iOS, https://github.com/iTwin/viewer-components-react/issues/1157 */} />}
      />
```

Example with hierarchyLevelSizeLimit={5}:
![Image](https://github.com/user-attachments/assets/13255128-c7f6-4a41-bcb8-c8555f0338e6)

Example of "Please provide additional filtering.." message mentioning 1000 instead of 5:
![Image](https://github.com/user-attachments/assets/168c766c-84db-494f-bf00-6870ef5f6d04)

ValdasSorys:
@grigasp Do you have a guess when this would be released?

grigasp:
> [@grigasp](https://github.com/grigasp) Do you have a guess when this would be released?

Sorry, this got out of the radar with all that was going on for the new `4.0`. The fix should be fairly simple, I hope @saskliutas can take of it this week.

saskliutas:
@ValdasSorys `@itwin/tree-widget-react@3.6.1` is available with a fix.

PR #1292, opened 2025-04-17
Tree widget level size config

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/1239
Closes https://github.com/iTwin/viewer-components-react/issues/1237

Bumped `@itwin/presentation-hierachies-react` to the latest version that does correctly handles hierarchy level size limit in levels that do not support filtering.
Updated visibility tree delivered by `@itwin/tree-widget-react` package to correctly apply custom hierarchy level size limit.

### DIFF
--- a/change/@itwin-tree-widget-react-3c0d5f0d-6322-4687-b98c-a090d4527382.json
+++ b/change/@itwin-tree-widget-react-3c0d5f0d-6322-4687-b98c-a090d4527382.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed `ModelsTree` and `CategoriesTree` not applying custom hierarchy level size limit.",
+  "packageName": "@itwin/tree-widget-react",
+  "email": "24278440+saskliutas@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/Tree.tsx
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/Tree.tsx
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { useCallback, useMemo, useState } from "react";
+import { useCallback, useState } from "react";
 import { BeEvent } from "@itwin/core-bentley";
 import { Flex, ProgressRadial, Text } from "@itwin/itwinui-react";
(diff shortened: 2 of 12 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/Tree.tsx
```diff
@@ -71,26 +71,39 @@ export type TreeProps = Pick<FunctionProps<typeof useIModelTree>, "getFilteredPa
...
+      getSchemaContext={getSchemaContext}
+      selectionStorage={selectionStorage}
+      imodelAccess={imodelAccess}
+      currentHierarchyLevelSizeLimit={currentHierarchyLevelSizeLimit}
+    />
+  );
+}
 
+/** @internal */
+export function TreeBase({ getSchemaContext, ...props }: MarkRequired<TreeProps, "imodelAccess"> & { currentHierarchyLevelSizeLimit: number }) {
```
grigasp, line 92: do we need this? It looks like we could just rename `TreeBaseImpl` to `TreeBase` and export that.

change/@itwin-tree-widget-react-3c0d5f0d-6322-4687-b98c-a090d4527382.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fixed `ModelsTree` and `CategoriesTree` not applying custom hierarchy level size limit.",
```
grigasp, line 3: is it just those trees or others too?

saskliutas: Only those two as they were using `VisibilityTree` component. Others used `Tree` which correctly handled hierarchy level size config.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/components/Tree.tsx
saskliutas: It add `SchemaMetadataContextProvider`. Moving it inside `TreeBaseImpl` didn't work. I guess some hook depends on that context. Others solution would be to add `SchemaMetadataContextProvider` in both `Tree` and `VisbilityTree` components. That way we could remove this one.

grigasp: nah, if some hook used in `TreeBaseImpl` needs this context, then I think let's keep this as is

grigasp (approved):

Issue #1285, opened 2025-04-10
Tree widget: Performance tests


grigasp:
There're a number of performance-intensive operations being done in tree widget package - need performance tests to help us identify problematic areas and avoid regressions.

I think the performance tests should be run in a similar way to what we have in `presentation` repo: https://github.com/iTwin/presentation/blob/master/.github/workflows/benchmark-pr-hierarchies.yml. Ideally, the fix to https://github.com/iTwin/presentation/issues/809 should also be applied.

PR #1310, opened 2025-05-07
Add performance tests

JonasDov:
closes [#1285](https://github.com/iTwin/viewer-components-react/issues/1285)

The comment was generated because in one commit I had added 'auto-push' flag. Action committed those values to the [gh-pages](https://github.com/iTwin/viewer-components-react/tree/gh-pages) branch. Then I removed the `auto-push` flag, and it was able to compare those two results.


packages/itwin/test-utilities/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "devDependencies": {
+    "@itwin/core-bentley": "^4.11.1",
+    "@itwin/core-common": "^4.11.1",
+    "@itwin/core-geometry": "^4.11.1",
+    "@itwin/eslint-plugin": "^5.1.0",
+    "@types/node": "^22.15.4",
+    "eslint": "^9.26.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-react": "^7.37.4",
```
grigasp, line 32: this package is not using react, so no need to have `eslint-plugin-react` dependency

packages/itwin/test-utilities/tsconfig.cjs.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "CommonJS",
```
grigasp, line 4: I don't think this package needs to support cjs

change/@itwin-tree-widget-react-a80f0991-a700-490d-9139-2c14f79db496.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Export internal packages that shouldn't be used by consumers under ./internal",
```
grigasp, line 3: I don't think this needs to be mentioned at all.

apps/performance-tests/.mocharc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "require": ["raf/polyfill", "ignore-styles", "./lib/setup.js"],
+  "timeout": 300000,
+  "check-leaks": true,
+  "globals": ["IS_REACT_ACT_ENVIRONMENT"],
```
grigasp, line 5: - do we need to require "raf/polyfill", "ignore-styles"?
- do we need `IS_REACT_ACT_ENVIRONMENT`?

apps/performance-tests/eslint.config.cjs
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const reactPlugin = require("eslint-plugin-react");
```
grigasp, line 7: we're not using react in this package

apps/performance-tests/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "compilerOptions": {
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "jsx": "react-jsx",
```
grigasp, line 6: we're not using react

apps/performance-tests/src/setup.ts
```diff
@@ -0,0 +1,74 @@
...
+    public width = 0,
+    public height = 0,
+  ) {}
+  public static fromRect(other?: DOMRectInit): DOMRect {
+    return new DOMRect(other?.x, other?.y, other?.width, other?.height);
+  }
+  public toJSON() {
+    return JSON.stringify(this);
+  }
+};
```
grigasp, line 38: do we need these?

grigasp: also, do we need "react" stuff below?

apps/performance-tests/src/main.ts
```diff
@@ -0,0 +1,42 @@
...
+          mode: HierarchyCacheMode.Memory,
+        },
+      },
+    },
+    rpcs: [SnapshotIModelRpcInterface, IModelReadRpcInterface, PresentationRpcInterface, ECSchemaRpcInterface],
+  });
+  ECSchemaRpcImpl.register();
+  await Datasets.initialize("./datasets");
+  await UiFramework.initialize();
+  await TreeWidget.initialize(new EmptyLocalization());
```
grigasp, line 35: are we using anything from tree widget package that requires it to be initialized?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -0,0 +1,239 @@
...
+    label: "",
+    parentKeys: [],
+    extendedData: {
+      isModel: true,
+      modelId: modelId ?? "0x1",
+    },
+  };
+}
+
+export async function getAllItems(imodelAccess: IModelAccess): Promise<Array<InstanceKey>> {
```
grigasp, line 210: can we improve its name somehow?..

apps/performance-tests/src/tree-widget/StatelessHierarchyProvider.ts
```diff
@@ -0,0 +1,150 @@
...
+    paths: HierarchyFilteringPath[];
+  };
+}
+type ProviderOptionsWithIModel = { iModel: IModelDb } & ProviderOptionsBase;
+
+type ProviderOptionsWithIModelAccess = { imodelAccess: IModelAccess } & ProviderOptionsBase;
+
+type ProviderOptions = ProviderOptionsWithIModel | ProviderOptionsWithIModelAccess;
+
+const LOG_CATEGORY = "Presentation.PerformanceTests.StatelessHierarchyProvider";
```
grigasp, line 36:
```suggestion
const LOG_CATEGORY = "TreeWidget.PerformanceTests.StatelessHierarchyProvider";
```

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        imodelAccess,
+        getHierarchyFactory: () => new ModelsTreeDefinition({ imodelAccess, idsCache, hierarchyConfig: defaultModelsTreeHierarchyConfiguration }),
+        filtering,
+      });
+      const result = await provider.loadHierarchy({ depth: 2 });
+      expect(result).to.eq(2);
+    },
+  });
+  run<{ iModel: SnapshotDb; imodelAccess: IModelAccess; targetItems: Array<InstanceKey> }>({
+    testName: "creates initial filtered view for 5k target items",
```
grigasp, line 72: I don't think there's a point to have a test for 50k and 5k. 50k is enough.

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      props.provider[Symbol.dispose]();
+      props.idsCache[Symbol.dispose]();
+    },
+    test: async ({ viewport, handler, provider, models }) => {
+      await validateHierarchyVisibility({
+        provider,
+        handler,
+        viewport,
+        expectations: "all-hidden",
+      });
```
grigasp, line 153: shouldn't this be in `setup`?

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  run<{
+    iModel: SnapshotDb;
+    idsCache: ModelsTreeIdsCache;
+    imodelAccess: IModelAccess;
+    viewport: Viewport;
+    handler: HierarchyVisibilityHandler & Disposable;
+    provider: HierarchyProvider & Disposable;
+    models: Id64Array;
+  }>({
+    testName: "changes visibility for 5k items",
```
grigasp, line 117: I think it would be better to have more targeted tests. So instead of changing visibility of a bunch of models+categories+subcategories+elements, have 4 tests that test each of them individually. If needed, we can increase the number of items to make sure the test takes enough time.

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
grigasp: same for categories tree

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
grigasp: same for categories tree

.github/workflows/benchmark-pr-tree-widget.yml
```diff
@@ -0,0 +1,69 @@
+name: PR tree-widget performance benchmark
+
+on:
+  pull_request:
+    branches:
+      - master
+      - stable
```
grigasp, line 7: don't need `stable` here

.github/workflows/benchmark-pr-tree-widget.yml
```diff
@@ -0,0 +1,69 @@
...
+    paths:
+      - apps/performance-tests/tree-widget/**
+      - apps/performance-tests/util/**
+      - packages/itwin/tree-widget/**
+      - pnpm-lock.yaml
```
grigasp, line 12: why do we need the root lockfile?

.github/workflows/benchmark-tree-widget.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Store benchmark result
+        uses: JonasDov/github-action-benchmark@04d30c72aa7aad091b58aacbb6c182ebb2ecb6c0
+        if: ${{ github.ref == 'refs/heads/master' }}
+        with:
+          name: Tree-Widget benchmark
+          tool: 'customSmallerIsBetter'
+          output-file-path: ./apps/performance-tests/tree-widget-benchmark.json
+          # Access token to deploy GitHub Pages branch
+          github-token: ${{ secrets.IMJS_ADMIN_GH_TOKEN }}
```
saskliutas, line 70: Based on permissions you set up for this workflow wouldn't it be enough to use `secrets.GITHUB_TOKEN`?

packages/itwin/test-utilities/eslint.config.cjs
```diff
@@ -0,0 +1,42 @@
...
+ *--------------------------------------------------------------------------------------------*/
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const reactPlugin = require("eslint-plugin-react");
+const eslintConfigPrettier = require("eslint-config-prettier");
+const unusedImports = require("eslint-plugin-unused-imports");
+
+module.exports = [
+  {
+    files: ["**/*.{ts,tsx}"],
+    ...iTwinPlugin.configs.uiConfig,
```
saskliutas, line 13: I don't think this package should use `uiConfig`.

packages/itwin/test-utilities/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "devDependencies": {
+    "@itwin/core-bentley": "^4.11.1",
+    "@itwin/core-common": "^4.11.1",
+    "@itwin/core-geometry": "^4.11.1",
+    "@itwin/eslint-plugin": "^5.1.0",
+    "@types/node": "^22.15.4",
+    "eslint": "^9.26.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-import": "^2.29.1",
```
saskliutas, line 31: Is this used anywhere?

apps/performance-tests/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "vcr-performance-tests",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "benchmark:tree-widget": "cross-env NODE_OPTIONS=\"--import ../../node-hooks/ignore-styles/register.cjs\" npm run test:tree-widget -- -O BENCHMARK_OUTPUT_PATH=./tree-widget-benchmark.json",
```
saskliutas, line 7: is `ignore-styles` plugin is necessary for running these tests?

apps/performance-tests/eslint.config.cjs
```diff
@@ -0,0 +1,43 @@
...
+const eslintConfigPrettier = require("eslint-config-prettier");
+const unusedImports = require("eslint-plugin-unused-imports");
+
+module.exports = [
+  {
+    files: ["**/*.{ts,tsx,cts}"],
+    ...iTwinPlugin.configs.uiConfig,
```
saskliutas, line 14: This package also should not use `uiConfig`

saskliutas (changes requested):

packages/itwin/test-utilities/package.json
JonasDov: Removed

packages/itwin/test-utilities/tsconfig.cjs.json
JonasDov: Removed

apps/performance-tests/.mocharc.json
JonasDov: Removed

apps/performance-tests/eslint.config.cjs
JonasDov: Removed

apps/performance-tests/src/setup.ts
JonasDov: Removed everything except this:
```ts
import globalJsdom from "global-jsdom";
import * as jsdom from "jsdom";
globalJsdom(undefined, {
  virtualConsole: new jsdom.VirtualConsole().sendTo(console, { omitJSDOMErrors: true }),
});
```
It is needed since we are using viewport in some tests

apps/performance-tests/src/main.ts
JonasDov: Yes, visibility tests fail without these.

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
JonasDov: Renamed to `getAllIModelElements` not sure if this is better...

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: I think there is a point. We can see if the performance drops linearly or exponentially. 
But removed.

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: Moved to setup

.github/workflows/benchmark-pr-tree-widget.yml
JonasDov: Removed

.github/workflows/benchmark-pr-tree-widget.yml
JonasDov: Removed

.github/workflows/benchmark-tree-widget.yml
JonasDov: I think it would, changed

packages/itwin/test-utilities/eslint.config.cjs
JonasDov: Removed

apps/performance-tests/package.json
JonasDov: Yes, we are using Appui and without it we get:
`Exception during run: TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".scss" for .../viewer-components-react/apps/performance-tests/node_modules/.pnpm/@itwin+appui-react@5.4.0_f1d58e90919febdfe342535e76db7f18/node_modules/@itwin/appui-react/lib/appui-react/layers.scss`

apps/performance-tests/eslint.config.cjs
JonasDov: Removed

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: Not sure what you mean. 
In models tree visibility tests we change visibility of one model and in turn it changes visibility of one category and 50k elements.
In categories tree visibility tests we change visibility of one category and in turn it changes visibility for 50k subCategories.
Then we validate if it changed what was expected.

Not sure if you are suggesting to change visibility of elements/subCategories directly? If so, should we change visibility of all of them (in such case the end result would be the same as before, except the test would probably take longer), or only some of them?
Also, if we test models/categories/subcategories/elements separately, then we should create two more iModels: one for 50k categories and one for 50k models (since current two cover subCategories and Elements)?

change/@itwin-tree-widget-react-a80f0991-a700-490d-9139-2c14f79db496.json
grigasp: I think `type` should be changed to `none`.

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
grigasp: > We can see if the performance drops linearly or exponentially.

We goal to is to capture performance regressions and improvements, not to draw diagrams :) Thanks.

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -118,27 +106,23 @@ export class Datasets {
   }
 
   /**
-   * Create an iModel with `numElements` functional 3D elements all belonging to the same spatial category, physical model and functional model.
+   * Create an iModel with `numElements` 3D elements all belonging to the same spatial category, physical model and functional model.
```
grigasp, line 109:
```suggestion
   * Create an iModel with `numElements` 3D elements all belonging to the same spatial category and physical model.
```

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -118,27 +106,23 @@ export class Datasets {
...
    * The elements are set up in a hierarchical manner, with 1000 top level 3D elements, each having 1 child element, which has 1 child element,
    * and so on until the depth of `numElements` / 1000 elements is reached. Each 3D element has a related functional element.
```
grigasp, line 111:
```suggestion
   * and so on until the depth of `numElements` / 1000 elements is reached.
```

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
grigasp: For some reason it looked to me that you're changing visibility of everything that `getAllItems` call returns 🤔 Yeah, it makes sense to change visibility of one instance, and measure the time it takes for the tree visibilities to be updated. 

Do you think we should also have tests for Models tree, where we change visibility of: a category, a per-model-category override, a single element?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
grigasp: Not really, and maybe even worse, since the function returns instance keys of not just elements, but also models. What if we do the following?

```ts
async function getVisibilityTargets(imodel: IModelAccess): Promise<{ models: Id64Array; categories: Id64Array; subCategories: Id64Array; elements: Id64Array }> {
  // implementation
}
```

change/@itwin-tree-widget-react-a80f0991-a700-490d-9139-2c14f79db496.json
JonasDov: Changed

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: I agree with having those tests. Added

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -207,7 +203,46 @@ export function createModelHierarchyNode(modelId?: Id64String, hasChildren?: boo
   };
 }
 
-export async function getAllIModelElements(imodelAccess: IModelAccess): Promise<Array<InstanceKey>> {
+export function createElementHierarchyNode(props: {
+  modelId: Id64String | undefined;
+  categoryId: Id64String | undefined;
+  hasChildren?: boolean;
+  elementId?: Id64String;
```
grigasp, line 212: why are `modelId`, `categoryId` and `elementId` optional?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -6,25 +6,40 @@
...
+  differentNodeExpectations?: VisibilityExpectations;
 }
 
-async function validateNodeVisibility({ node, handler, expectations }: ValidateNodeProps & { node: HierarchyNode }) {
+async function validateNodeVisibility({ node, handler, expectations, differentNodeExpectations }: ValidateNodeProps & { node: HierarchyNode }) {
+  if (HierarchyNode.isClassGroupingNode(node)) {
+    return;
+  }
+  assert(HierarchyNode.isInstancesNode(node));
+  const { id } = node.key.instanceKeys[0];
```
grigasp, line 39: any reason not to check all `instanceKeys`?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -6,25 +6,40 @@
...
+
+export interface VisibilityExpectations {
+  [id: string]: Visibility;
+}
 
 export interface ValidateNodeProps {
   handler: HierarchyVisibilityHandler;
   viewport: Viewport;
   expectations: "all-visible" | "all-hidden";
+  differentNodeExpectations?: VisibilityExpectations;
```
grigasp: instead of `expectations` + `differentNodeExpectations`, could we have something like this?

```ts
expectations: "all-visible" | "all-hidden" | {
  default: "all-visible" | "all-hidden";
  instances: { [id: Id64String]: Visibility };
};
```

apps/performance-tests/src/tree-widget/CategoriesTree.test.ts
```diff
@@ -109,4 +110,58 @@ describe("categories tree", () => {
...
+      return { iModel, imodelAccess, viewport, provider, handler, models: visibilityTargets.models };
+    },
+    cleanup: async (props) => {
+      props.iModel.close();
+      props.viewport.dispose();
+      props.handler[Symbol.dispose]();
+      props.provider[Symbol.dispose]();
+    },
+    test: async ({ viewport, handler, provider, models }) => {
+      await Promise.all(models.map(async (model) => handler.changeVisibility(createDefinitionContainerHierarchyNode(model), true)));
```
grigasp, line 158: I believe `model` also contains physical model id, doesn't it? I think the intention is to only change visibility of the definition container.

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -124,4 +126,225 @@ describe("models tree", () => {
...
+    },
+    cleanup: async (props) => {
+      props.iModel.close();
+      props.viewport.dispose();
+      props.handler[Symbol.dispose]();
+      props.provider[Symbol.dispose]();
+      props.idsCache[Symbol.dispose]();
+    },
+    test: async ({ viewport, handler, provider, categories, elementsModel }) => {
+      await Promise.all(categories.map(async (category) => handler.changeVisibility(createCategoryHierarchyNode(category, true, elementsModel), false)));
```
grigasp, line 190: I find this part in all the tests pretty confusing. The test says it's changing visibility of a single category, but here we're changing it for an array. Maybe the `setup` functions that create the iModel and know that the array contains only 1 element could return just 1 category instead of an array with 1 element?

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -124,4 +126,225 @@ describe("models tree", () => {
...
+            SELECT
+              Model.Id AS ECInstanceId
+            FROM bis.GeometricElement3d
+            LIMIT 1
+          `,
+      };
+      let elementsModel = "";
+      for await (const row of imodelAccess.createQueryReader(query, { limit: "unbounded" })) {
+        elementsModel = row.ECInstanceId;
+      }
```
grigasp, line 179: maybe we could instead use `iModel.elements.getElementProps<GeometricElementProps>(visibilityTargets.elements[0])` for this?

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
```diff
@@ -124,4 +126,225 @@ describe("models tree", () => {
...
+    },
+    cleanup: async (props) => {
+      props.iModel.close();
+      props.viewport.dispose();
+      props.handler[Symbol.dispose]();
+      props.provider[Symbol.dispose]();
+      props.idsCache[Symbol.dispose]();
+    },
+    test: async ({ viewport, handler, provider, categories, elementsModel }) => {
+      elementsModel;
```
grigasp, line 261: needs cleanup?

JonasDov: Removed

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/ModelsTree.test.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/CategoriesTree.test.ts
JonasDov: Changed to only change definition container visibility

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
JonasDov: Changed

apps/performance-tests/src/tree-widget/CategoriesTree.test.ts
```diff
@@ -146,16 +148,28 @@ describe("categories tree", () => {
...
+            SELECT
+              Model.Id AS ECInstanceId
+            FROM bis.SpatialCategory
+            LIMIT 1
+          `,
+      };
+      let definitionContainer = "";
+      for await (const row of imodelAccess.createQueryReader(query, { limit: "unbounded" })) {
+        definitionContainer = row.ECInstanceId;
+      }
```
grigasp, line 162: could we instead do the following?
```ts
const definitionContainer = iModel.elements.getElementProps(visibilityTargets.categories[0]).model;
```

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -0,0 +1,306 @@
...
+      if (id in expectations.instances) {
+        expect(actualVisibility.state).to.eq(expectations.instances[id]);
+        return true;
+      }
+      return false;
+    })
+  ) {
+    return;
+  }
+  expect(actualVisibility.state).to.eq(expectations.default === "all-hidden" ? "hidden" : "visible");
```
grigasp, line 58:
```suggestion
  const idInExpectations = ids.find((id) => id in expectations.instances)
  if (idInExpectations) {
    const expectedVisibility = expectations.instances[idInExpectations];
    expect(actualVisibility.state).to.eq(expectedVisibility);
  } else {
    expect(actualVisibility.state).to.eq(expectations.default === "all-hidden" ? "hidden" : "visible");
  }
```

apps/performance-tests/src/tree-widget/CategoriesTree.test.ts
JonasDov: Yes, changed

grigasp (approved):

saskliutas (approved):

Issue #1309, opened 2025-05-06
Tree widget: Models tree view fails to hide every element in a group


dovydasjaru:
When using Models tree view component in iModel viewer some elements with the same category label, but different categories get grouped into the same group. But hiding/unhiding the group does affect the whole group as shown below in video. 

[Here is a link](https://qa-infrastructurecloud.bentley.com/dcb7e33b-ba59-42df-84db-bbd1ccc739f7/manage-imodel/unified-viewer/2bd6cb7e-9874-44ba-82b8-acada84bf82c/c930077c118ff1d97b61090db7412e5c0fcbecaa) to iModel. If access is needed I can grant it.

https://github.com/user-attachments/assets/49e377ca-04a6-4ef6-88ee-096ca4df2fa4

grigasp:
@dovydasjaru, the fix was released with version [3.8.0](https://www.npmjs.com/package/@itwin/tree-widget-react/v/3.8.0).

PR #1311, opened 2025-05-08
Make merged nodes visibility work as expected

JonasDov:
closes [#1309](https://github.com/iTwin/viewer-components-react/issues/1309)

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/models-tree/internal/VisibilityValidation.ts
+++ b/packages/itwin/tree-widget/src/test/trees/models-tree/internal/VisibilityValidation.ts
@@ -7,6 +7,7 @@ import { assert, expect } from "chai";
 import { expand, from, mergeMap } from "rxjs";
 import { PerModelCategoryVisibility } from "@itwin/core-frontend";
 import { HierarchyNode } from "@itwin/presentation-hierarchies";
+import { waitFor } from "@testing-library/react";
 import { toVoidPromise } from "../../../../tree-widget-react/components/trees/common/Rxjs.js";
 import { ModelsTreeNode } from "../../../../tree-widget-react/components/trees/models-tree/internal/ModelsTreeNode.js";
 
@@ -123,7 +124,7 @@ export async function validateHierarchyVisibility({
   await toVoidPromise(
     from(provider.getNodes({ parentNode: undefined })).pipe(
       expand((node) => provider.getNodes({ parentNode: node })),
-      mergeMap(async (node) => validateNodeVisibility({ ...props, node })),
+      mergeMap(async (node) => waitFor(async () => validateNodeVisibility({ ...props, node }))),
     ),
   );
 }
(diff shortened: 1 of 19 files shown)


packages/itwin/tree-widget/src/test/trees/models-tree/internal/VisibilityValidation.ts
```diff
@@ -7,6 +7,7 @@ import { assert, expect } from "chai";
 import { expand, from, mergeMap } from "rxjs";
 import { PerModelCategoryVisibility } from "@itwin/core-frontend";
 import { HierarchyNode } from "@itwin/presentation-hierarchies";
+import { waitFor } from "@testing-library/react";
```
saskliutas, line 10: Can you import from `TestUtils.tsx` so that we have single place where it's imported in tests.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -54,7 +53,7 @@ import type { FilteredTree } from "./FilteredTree.js";
 
 /** @beta */
 interface GetCategoryVisibilityStatusProps {
-  categoryId: Id64String;
+  categoryId: Id64Arg;
```
saskliutas, line 56:
```suggestion
  categoryIds: Id64Arg;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -966,32 +982,50 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
       }),
+      mergeVisibilityStatuses(),
     );
   }
 
-  private getAlwaysDrawnElements(props: AlwaysOrNeverDrawnElementsQueryProps) {
-    return this._alwaysAndNeverDrawnElements.getElements({ ...props, setType: "always" });
+  private getAlwaysDrawnElements(props: { categoryId?: Id64Arg; modelId: Id64String }): Observable<Id64Set> {
+    return from(props.categoryId ? Id64.iterable(props.categoryId) : [undefined]).pipe(
+      mergeMap((categoryId) => this._alwaysAndNeverDrawnElements.getElements({ modelId: props.modelId, categoryId, setType: "always" })),
```
saskliutas, line 1010: Maybe `AlwaysAndNeverDrawnElementInfo` could be updated to make multiple categories instead of the single one?

saskliutas (changes requested):

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
grigasp: It was my suggestion to keep the same name, to avoid breaking callers of visibility handler. I just realized that visibility handler is internal, so changing the prop name would only break consumers who we already break by changing its type - the ones who use `ModelsTreeVisibilityHandlerOverrides`, likely nobody.

Let's do the rename.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
JonasDov: Changed

saskliutas (approved):

PR #1321, opened 2025-05-15
Add support for Google Places API in `<GeoAddressSearch>` 

mdastous-bentley:
Address provider can now be optionally set to Google, which will make the components consume the [Google Places API](https://developers.google.com/maps/documentation/places/web-service/op-overview#place_autocomplete).  
An appropriate Google API key needs to be provided.
image

**IMPORTANT**:
You must use a Google base map in order to use the Google address search, this is required by the  Google policy.  If you switch to another base map, the widget get automatically disabled until you re-enable a Google base map:
image

### DIFF
--- a/change/@itwin-geo-tools-react-13cc9b66-4040-4f9c-95e5-48bda9bfffc5.json
+++ b/change/@itwin-geo-tools-react-13cc9b66-4040-4f9c-95e5-48bda9bfffc5.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add support for Google Places API in `<GeoAddressSearch>` ",
+  "packageName": "@itwin/geo-tools-react",
+  "email": "mdastous-bentley@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}


packages/itwin/geo-tools/src/GeoToolsUiItemsProvider.tsx
```diff
@@ -3,12 +3,19 @@
...
 
+import type { GeoToolsOptions } from "./GeoToolsItemDef";
+import type { ToolbarItem, UiItemsProvider } from "@itwin/appui-react";
+
 export class GeoToolsAddressSearchProvider implements UiItemsProvider {
   public readonly id = "GeoToolsAddressSearchProvider";
+  private _opts?: GeoToolsOptions;
+
+
+  public constructor(opts?: GeoToolsOptions) {
```
mdastous-bentley, line 16: @aruniverse If some other custom API would need to be used, a custom address provider can be passed here.

This new new provider could be created by sub-classing one of the two existing providers.

aruniverse (approved):

packages/itwin/geo-tools/src/GoogleAddressProvider.ts
```diff
@@ -0,0 +1,128 @@
...
+  public supportsAddressLocation(): this is GeoCoder  {
+    return true;
+  }
+
+  protected async getAuthRequestHeader(): Promise<Record<string, string>> {
+    return { "X-Goog-Api-Key": this._apiKey};
+  }
+
+  protected async getPlacesBaseUrl(): Promise<string> {
+    return "https://places.googleapis.com/v1/places/";
```
aruniverse, line 42: any attributions we need to show when using this?

mdastous-bentley: The attribution is the Google logo display on the base, hence the google base map most be used.

PR #1325, opened 2025-05-16
Move CI to github actions

saskliutas:
Majority of pipelines in this repo are using github actions so moved CI pipeline to github actions as well.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          node-version: 22
+          cache: 'pnpm'
+
+      - run: pnpm install
+
+      - run: pnpm exec lage lint extract-api cover --grouped
+        name: Build Lint Test and check API changes
+        env:
+          TF_BUILD: 1
```
aruniverse, line 90: whats this env var?

aruniverse (approved):

.github/workflows/ci.yml
saskliutas: It's used by `@itwin/build-tools` `extract-api script to determine if script is running in CI job https://github.com/iTwin/itwinjs-core/blob/5cf7b4dcbf2442215a4433dc2281a90688834c62/tools/build/scripts/extract-api.js#L23

Issue #1327, opened 2025-05-19
Tree widget: Add iModel classifications tree


grigasp:
Requirements defined internally.

grigasp:
Released with version `4.0.0-alpha.11`

PR #1331, opened 2025-05-23
Tree widget/classifications tree on next

grigasp:
Part of https://github.com/iTwin/viewer-components-react/issues/1327.

This covers the read-only case, when getting Categories under a Classification is done through elements. 

For read-write we need to show Categories without elements too, and for that we need to know the name of relationship between  a Category and a Classification. So when given that name, we'll use a different approach to get the Categories.


.npmrc
```diff
@@ -1,5 +1,6 @@
-@bentley:registry=https://registry.npmjs.org/
+@bentley:registry=https://pkgs.dev.azure.com/bentleycs/_packaging/Packages/npm/registry/
```
aruniverse, line 1: Are we sure about adding this in this repo?

grigasp: No. I thought the classification systems' schema is only available on internal registry, but I see it in npm too, so the change shouldn't be necessary. reverted

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTreeDefinition.ts
```diff
@@ -0,0 +1,378 @@
...
+      instanceLabelSelectClauseFactory: this._nodeLabelSelectClauseFactory,
+    });
+  }
+
+  public async defineHierarchyLevel(props: DefineHierarchyLevelProps) {
+    return (await this.#getHierarchyDefinition()).defineHierarchyLevel(props);
+  }
+
+  async #getHierarchyDefinition(): Promise<HierarchyDefinition> {
+    this._impl ??= (async () => {
```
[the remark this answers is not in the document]
grigasp: it doesn't - cleaned up

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/internal/UseIdsCache.ts
```diff
@@ -0,0 +1,52 @@
...
+  const [getCache, setCacheGetter] = useState<() => ClassificationsTreeIdsCache>(() => createCacheGetterRef.current(imodel, rootClassificationSystemCode));
+
+  useEffect(() => {
+    // clear cache in case it was created before `useEffect` was run first time
+    clearCacheRef.current();
+
+    // make sure all cache users rerender
+    setCacheGetter(() => createCacheGetterRef.current(imodel, rootClassificationSystemCode));
+    return () => {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
grigasp: image

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/internal/ClassificationsTreeIdsCache.ts
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  private async *queryModeledElements(): AsyncIterableIterator<{
+    modelId: Id64String;
+    modeledElementId: Id64String;
+    categoryId: Id64String;
+    rootCategoryId: Id64String;
+  }> {
+    const query = `
+      SELECT *
```
saskliutas, line 122: One more unnecessary outer select

saskliutas (approved):

PR #1336, opened 2025-05-30
[measure-tools] Update quantity formatting logic, add bearing to distance measure

hl662:
- Bump peer deps for @itwin/measure-tools-react to itwin v5
- Add bearing field to MeasureDistanceTool's widget properties.
- Enable more text markers to dynamically update when formats changed events are emitted via IModelApp.quantityFormatter.
- Integrate Distance, Angle, Area, Location, Radius Measurement subclasses to use IModelApp.formatsProvider to retrieve formatting props.
- Add a `measurementFormattingProps` property to `MeasureToolsUiProviderOptions`. By default, Bearing will not be included unless an app explicitly passes in the `MeasurementFormattingProps` for bearing, like so:
```ts

	....other props for MeasureToolsUiItemsProvider
        measurementFormattingProps: {
          distance: {
            bearing: {
              koqName: "RoadRailUnits.Bearing",
              persistenceUnitName: "Units.RAD"
            }
          }
        }
```


packages/itwin/measure-tools/src/api/FormatterUtils.ts
```diff
@@ -55,14 +55,26 @@ export class FormatterUtils {
...
+    if (!coordSpec) {
+      coordSpec =
+        IModelApp.quantityFormatter.findFormatterSpecByQuantityType(
+          QuantityType.Coordinate
+        );
+      if (undefined === coordSpec) return "";
+      result = FormatterUtils.formatCoordinatesWithSpec(point, coordSpec);
+    } else {
+      // eslint-disable-next-line @itwin/no-internal
+      const oldFormatTraits = coordSpec.format.formatTraits;
```
hl662, line 69: @aruniverse these apis are marked as beta, but linter flags warnings for them complaining about the no internal rule. Is the rule a bug or should I disable the no internal rule line by line like this (or disable across file)

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -89,6 +90,15 @@ function onIModelConnected(imodel: IModelConnection) {
...
   }, 1000);
+
+  imodel.schemaContext.getSchema(new SchemaKey("AecUnits", SchemaMatchType.Latest)).then((schema) => {
+    if (schema) {
+      IModelApp.formatsProvider = new SchemaFormatsProvider(imodel.schemaContext, IModelApp.quantityFormatter.activeUnitSystem);
+      console.log("Registered SchemaFormatsProvider");
+    }
+  }).catch((_) => {
+    console.error("No common AecUnits schema found in iModel, not registering a SchemaFormatsProvider");
+  });
```
grigasp: Do we expect this to be done by all apps? Maybe it should be done by `IModelApp` then?

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -36,12 +36,17 @@ import { MeasurementSelectionSet } from "../api/MeasurementSelectionSet.js";
...
 import { DistanceMeasurement } from "./DistanceMeasurement.js";
 import { MeasureTools } from "../MeasureTools.js";
+import type { FormatterSpec } from "@itwin/core-quantity";
 
 /**
  * Props for serializing a [[AreaMeasurement]].
  */
 export interface AreaMeasurementProps extends MeasurementProps {
   polygonPoints: XYZProps[];
+  lengthKoQ?: string;
```
hl662, line 46: @grigasp should I wrap these new properties inside of one object? Like
```ts
...
formattingSpecProps: {
  lengthKoQ?: string;
  lengthPersistenceUnitName?: string;
  areaKoQ?: string;
  areaPersistenceUnitName?: string;
}
```

Could declare the above as it's own `AreaFormattingSpecProps`, and shorten the static create method at the bottom of this file... Fyi, that create method is the main entry point for the higher level APIs to instantiate these Measurement subclasses (up to the app level and them registering a MeasureToolsUiItemsProvider)

grigasp: I'm not familiar with APIs in this package, so don't have a strong opinion... But I like the idea of grouping similar props together. In fact, I'd probably go a step further and do this:
```ts
formatting: {
  length?: {
    /** Defaults to `AecUnits.LENGTH` */
    koqName?: string;
    /** Defaults to `Units.M` */
    persistenceUnitName?: string;
  },
  area?: {
    /** Defaults to `AecUnits.AREA` */
    koqName?: string;
    /** Defaults to `Units.SQ_M` */
    persistenceUnitName?: string;
  },
}
```

packages/itwin/measure-tools/src/api/FormatterUtils.ts
grigasp: I believe this is because of this config in package's `eslint.config.js`:
```
      "@itwin/no-internal": [
        "warn",
        {
          tag: [
            "internal",
            "alpha",
            "beta",
          ],
        },
      ],
```

apps/test-viewer/src/components/Viewer.tsx
grigasp: > the reason I put this in is that the measurements decorated onto the viewer calls getFormat. getFormat fails with a 'Schema not found', and with the amount of draw calls (its done on any mouse movement) this could overload the backend with so many RPC calls. Successful schema retrievals is fine though, as those schemas would be cached on the frontend, so getting a format would not perform RPC calls every draw call.

If we know that not finding a schema for `SchemaFormatsProvider` is kind of expected, maybe it should keep a set of schemas that it didn't find and avoid requesting them more than once?

On the other hand, after looking through the docs, I lack guidance on what I, as a tool (or anything else that needs formatting capabilities) developer, should do if I don't get the format for KoQ I requested.

Assuming an app is capable of opening a number of different iModels, some of which may or may not have schemas used for formatting by different tools, I don't see how such way of registering a formats provider could work.

hl662: > On the other hand, after looking through the docs, I lack guidance on what I, as a tool (or anything else that needs formatting capabilities) developer, should do if I don't get the format for KoQ I requested.

Yes, we'll add clarification on that... 
`IModelApp.quantityFormatter.formatQuantity()` asks for a value and an **optional**  formatterSpec. If the spec was undefined coming in, the method will simply return the `value.toString()`. That's a soft way of handling not finding a KoQ - to keep the formatterSpec that would be generated via the format be undefined.

Or, another way is for the tools to have fallbacks, notably by using its own schemas. This was suggested by @diegoalexdiaz, the tools can have its own schema locaters to fallback on, and have a dependency on some of the common schemas like 'AecUnits' or 'RoadRailUnits'. We can improve the npm packages that contain these schemas to [allow easier imports](https://github.com/iTwin/bis-schemas/issues/519) (rather than import through 'node_modules/units-schema/units.schema.json' ) and have the tools conditionally fallback on these, if the IModelApp.formatsProvider does not have these schemas. This would guarantee the tools will always format a quantity given.

packages/itwin/measure-tools/src/api/FormatterUtils.ts
hl662: thanks, I've omitted alpha and beta from the warning

aruniverse: id recommend leaving alpha in, need to be careful using those apis

hl662: alright, i added alpha back

apps/test-viewer/src/components/Viewer.tsx
hl662: > If we know that not finding a schema for SchemaFormatsProvider is kind of expected, maybe it should keep a set of schemas that it didn't find and avoid requesting them more than once?

I changed the way the measurements retrieve formats, to use `IModelApp.quantityFormatter.getSpecsByName`. That won't query the formatsProvider on every drawCall. On init of each of the measurements, the measurement will try to find the formatterSpec in quantityFormatter via `getSpecsByName`, and if it doesn't find something, it'll call the ImodelApp.formatsProvider to generate the specs and populate the cache that `getSpecsByName` uses, for later.

packages/itwin/measure-tools/package.json
```diff
@@ -44,41 +44,44 @@
     "url": "http://www.bentley.com"
   },
   "peerDependencies": {
-    "@itwin/appui-abstract": "^4.7.0",
+    "@itwin/appui-abstract": "^4.7.0 || ^5.0.0",
```
aruniverse, line 47: should we just drop support for 4?

packages/itwin/measure-tools/src/api/FormatterUtils.ts
```diff
@@ -55,14 +56,23 @@ export class FormatterUtils {
...
+    if (!coordSpec) {
+      coordSpec =
+        IModelApp.quantityFormatter.findFormatterSpecByQuantityType(
+          QuantityType.Coordinate
+        );
+      if (undefined === coordSpec) return "";
+      result = FormatterUtils.formatCoordinatesWithSpec(point, coordSpec);
+    } else {
+      const oldFormatTraits = coordSpec.format.formatTraits;
+      coordSpec.format.formatTraits &= ~FormatTraits.ShowUnitLabel;
```
aruniverse, line 70: what does the `~` do here?

packages/itwin/measure-tools/src/measurements/AreaMeasurement.ts
```diff
@@ -576,6 +640,11 @@ export class AreaMeasurement extends Measurement {
       if (this.isDynamic && this._dynamicEdge)
         this.updateDynamicPolygon(this._dynamicEdge.endPointRef);
     }
+
+    if (jsonArea.formatting?.area?.koqName) this._areaKoQ = jsonArea.formatting.area.koqName;
+    if (jsonArea.formatting?.area?.persistenceUnitName) this._areaPersistenceUnitName = jsonArea.formatting.area.persistenceUnitName;
+    if (jsonArea.formatting?.length?.koqName) this._lengthKoQ = jsonArea.formatting.length.koqName;
```
aruniverse, line 646: note: be careful since `length` is property on string and arays

packages/itwin/measure-tools/src/tools/MeasureToolDefinitions.ts
```diff
@@ -36,14 +41,14 @@ export class MeasureToolDefinitions {
     });
   }
 
-  public static getMeasureDistanceToolCommand(allowedViewportCallback: (vp: ScreenViewport) => boolean = (() => true), enableSheetMeasurements: boolean) {
+  public static getMeasureDistanceToolCommand(allowedViewportCallback: (vp: ScreenViewport) => boolean = (() => true), enableSheetMeasurements: boolean, formatting?: DistanceMeasurementFormattingProps) {
```
aruniverse, line 44: can you remove the deprecated tool definitions so we can make sure users arent using the older ones which wont be using the formatting apis weve added in this pr?

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -89,6 +90,29 @@ function onIModelConnected(imodel: IModelConnection) {
...
+
+      IModelConnection.onClose.addOnce(() => {
+        removeListener();
+        IModelApp.resetFormatsProvider();
+        void IModelApp.quantityFormatter.resetToUseInternalUnitsProvider();
+        console.log("Unregistered SchemaFormatsProvider, SchemaUnitProvider");
+      });
+    }
+  }).catch((_) => {
+    console.error("No common AecUnits schema found in iModel, not registering a SchemaFormatsProvider");
```
tcobbs-bentley, line 114: Two things:

1. If getSchema returns `null` or `undefined` without throwing, you won't print the console warning. I feel that this is incorrect behavior, but I could be wrong.
2. I don't like using `.then ... .catch` instead of `await`. I realize that the current function isn't `async`, but I feel that doing it this way obfuscates the code in a way that makes it non-obvious what is actually happening. An easy workaround to the missing `async` is as follows:

```ts
const setupFormatsProvider = async () => {
  try {
    const schema = await imodel.schemaContext.getSchema(new SchemaKey("AecUnits", SchemaMatchType.Latest);
    if (schema) {
      // ...
    }
  } catch (_ex) {
    console.error("No common AecUnits schema found in iModel, not registering a SchemaFormatsProvider");
  }
}
void setupFormatsProvider();
```

packages/itwin/measure-tools/package.json
hl662: okay, dropped

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -89,6 +90,34 @@ function onIModelConnected(imodel: IModelConnection) {
   setTimeout(() => {
     IModelConnection.onOpen.raiseEvent(imodel);
   }, 1000);
+  const setupFormatsProvider = async () => {
+    try {
+      const schema = await imodel.schemaContext.getSchema(new SchemaKey("AecUnits", SchemaMatchType.Latest));
+      if (schema) {
```
tcobbs-bentley, line 96: What about:
```ts
if (!schema)
  throw new Error("No schema");
```
That will do two things:

1. Unindent the active code by one level
2. Do the console log if schema isn't defined

Note: since the error is being ignored in the catch handler, it doesn't need to be fancy.

tcobbs-bentley: Oops; I see the error is no longer ignored in the catch handler; you can put a more descriptive error here.

apps/test-viewer/src/components/Viewer.tsx
hl662: I'm going to revisit this in a future [pr](https://github.com/iTwin/viewer-components-react/issues/1345), hoping to clean up event emitters surrounding formatsChanged.

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -823,6 +823,8 @@ export abstract class Measurement {
   /** Notify subclasses when the display units have changed. */
   public onDisplayUnitsChanged(): void { }
 
+  public async populateFormattingSpecsRegistry(): Promise<void> { }
```
raplemie, line 826: Can this new method be documented like every other method ?

packages/itwin/measure-tools/package.json
aruniverse: should prob update the appui deps to match the ver that supports core

hl662: I've updated the peer deps for appui to be 5.6.0, also dev deps too.

apps/test-viewer/src/UiProvidersConfig.tsx
```diff
@@ -238,7 +238,16 @@ const configuredUiItems = new Map<string, UiItem>([
         await MeasureTools.startup();
         MeasurementActionToolbar.setDefaultActionProvider();
       },
-      createUiItemsProviders: () => [new MeasureToolsUiItemsProvider()],
+      createUiItemsProviders: () => [new MeasureToolsUiItemsProvider({
+        measurementFormattingProps: {
+          distance: {
+            bearing: {
+              koqName: "RoadRailUnits.Bearing",
+              persistenceUnitName: "Units.RAD" // TODO: Once units schema 1.0.9 is released, change to Units.HORIZONTAL_DIR_RAD
```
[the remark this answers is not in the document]
hl662: @diegoalexdiaz mentioned RoadRailUnits.Bearing should have its Bearing [KoQ](https://github.com/iTwin/bis-schemas/blob/114a5e913bdc6aebb46e6fd394374ea02aab04a8/Domains/2-DisciplinePhysical/Civil/RoadRailUnits/RoadRailUnits.ecschema.xml#L23) changed to use HORIZONTAL_DIRECTION units, like `HORIZONTAL_DIR_RAD` which would be available in a later version of units schema. That, or add a new KoQ separate from `RoadRailUnits.Bearing`. 

newer iModels should(?) support it, such as iModels that go through OpenSite+ workflow, but perhaps in that case, the MeasureToolsUiProvider used by OpenSite+ will change it's persistenceUnitName to `Units.HORIZONTAL_DIR_RAD`.

While for the test viewer here, we can leave it as `Units.RAD` instead, and remove the TODO. The bearing value, unlike the other measurements to be displayed, is not shown unless an application explicitly provides the a `distance.bearing` property to the `MeasurementFormattingProps` of the `MeasureToolsUIProvider()` constructor

packages/itwin/measure-tools/vite.config.ts
```diff
@@ -7,12 +7,29 @@ export default defineConfig({
...
+  },
   test: {
     globals: true,
     environment: "jsdom",
     setupFiles: ["src/test/measure-tools/_Setup.test.ts"],
     include: ["src/test/measure-tools/**/*.test.*"],
     exclude: ["src/test/measure-tools/_Setup.test.ts"],
+    server: {
+      deps: {
+        inline: [
```
aruniverse, line 25: why? can you add a comment

packages/itwin/measure-tools/src/test/measure-tools/_Setup.test.ts
```diff
@@ -4,16 +4,27 @@
...
+import * as path from "path";
 
 // Before all tests, initialize any global services
 beforeAll(async () => {
   const localization = new EmptyLocalization();
   await NoRenderApp.startup({ localization });
   await MeasureTools.startup({ localization });
+
+  const schemaContext = new SchemaContext();
+  const unitSchemaFilePath = path.join(__dirname, "..", "..", "..", "node_modules", "@bentley", "units-schema");
```
aruniverse, line 23: why not use `require.resolve`?

aruniverse (commented):
first pass looks ok. please have civil team review, they will be impacted

packages/itwin/measure-tools/src/test/measure-tools/_Setup.test.ts
hl662: fair point, changed to use require.resolve

packages/itwin/measure-tools/src/measurements/AngleMeasurement.ts
```diff
@@ -43,6 +43,13 @@ export interface AngleMeasurementProps extends MeasurementProps {
   startPoint?: XYZProps;
   center?: XYZProps;
   endPoint?: XYZProps;
+  formatting?: AngleMeasurementFormattingProps;
+}
+
+/** Formatting properties for angle measurement. */
+export interface AngleMeasurementFormattingProps {
+  /** Defaults to "AecUnits.ANGLE" and "Units.RAD" */
```
tcobbs-bentley, line 51: Is there a reason that all the angle formats in this PR default to radians? These are designed to be displayed to people right? Wouldn't most people prefer degrees?

aruniverse (approved):

PR #1342, opened 2025-06-04
[Tree-widget]: add rename action to classifications tree

MartynasStrazdas:
Preview
image
image


packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTree.tsx
```diff
@@ -31,13 +36,35 @@ export function ClassificationsTree({
...
+    },
+    [getActions],
+  );
+
+  const getEditingProps = useCallback<Required<ComponentProps<typeof StrataKitTreeRenderer>>["getEditingProps"]>(
+    (node) => {
+      return {
+        onLabelChanged: onLabelChanged
+          ? (newLabel) => {
+              node.label = newLabel;
```
saskliutas, line 59: This does not look right. Component should not directly mutate node. Why is this needed?

saskliutas (changes requested):

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTree.tsx
MartynasStrazdas: Forgot to save after experimenting, removed.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTree.tsx
```diff
@@ -18,6 +22,7 @@ export type ClassificationsTreeProps = Pick<VisibilityTreeProps, "imodel" | "sel
     hierarchyLevelConfig?: {
       sizeLimit?: number;
     };
+    onLabelChanged?: (node: PresentationHierarchyNode, newLabel: string) => void;
```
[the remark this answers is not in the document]
grigasp: I agree, and actually I'm expecting to have this situation of only some of the nodes being editable in our immediate consumer app.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTree.tsx
```diff
@@ -31,13 +36,34 @@ export function ClassificationsTree({
...
 }: ClassificationsTreeProps) {
   const { categoriesTreeProps, rendererProps } = useClassificationsTree({
     activeView,
     hierarchyConfig,
     emptyTreeContent,
   });
 
+  const nodeActions = useCallback(
+    (node: PresentationHierarchyNode) => {
+      return [<RenameAction key="RenameAction" />, ...(getActions ? getActions(node) : [])];
```
grigasp, line 49: I don't think we should be adding this action ourselves - we'll have consumers that show this tree as read-only, so they won't want it.

Instead, I think we should just add a piece of docs next to `onLabelChanged` / `getEditingProps`, explaining that label editing requires adding a `<RenameAction />`.

grigasp: I agree that docs in `@itwin/presentation-hierarchies-react` is probably the best place to explain this, since both the `<RenameAction />` and `getEditingProps` prop are defined there.

PR #1346, opened 2025-06-11
Some cleanup of the FormatterUtils functions. Also converted class to…

a-gagnon:
- Cleaned up the `removeUnitSuffixes` function which now uses the `FormatTraits` properly

### DIFF
--- a/packages/itwin/measure-tools/src/api/FormatterUtils.ts
+++ b/packages/itwin/measure-tools/src/api/FormatterUtils.ts
@@ -10,89 +10,57 @@ import type { FormatProps } from "@itwin/core-quantity";
 import { FormatTraits, type FormatterSpec } from "@itwin/core-quantity";
 import { MeasureTools } from "../MeasureTools.js";
 
-export class FormatterUtils {
-  private static removeUnitSuffixes(s: string) {
-    s = s.replace(/ m/g, "");
-    s = s.replace(/ ft/g, "");
-    return s;
+export namespace FormatterUtils {
+
+  /** Formats a sequence of values with spec without the unit label */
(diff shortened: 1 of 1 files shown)


packages/itwin/measure-tools/src/api/FormatterUtils.ts
```diff
@@ -10,97 +10,65 @@ import type { FormatProps } from "@itwin/core-quantity";
 import { FormatTraits, type FormatterSpec } from "@itwin/core-quantity";
 import { MeasureTools } from "../MeasureTools.js";
 
-export class FormatterUtils {
-  private static removeUnitSuffixes(s: string) {
-    s = s.replace(/ m/g, "");
-    s = s.replace(/ ft/g, "");
-    return s;
-  }
+export namespace FormatterUtils {
```
hl662, line 13: just out of curiosity and learning, why convert to a namespace?

a-gagnon: To be quite frank, it's mostly syntactic sugar, but I can think of the following things:

- You can always create an instance of a class
```ts
const utils = new FormatterUtils();
```
Similarly, you can also **extend** any class. Again, not something that someone will ever **want** to do in this case.

- You can't do nesting or define types in classes. It's possible to do it in namespaces:
```ts
namespace Foo {
  namespace Bar {
    type NumberType = number;
  }
}
const n: Foo.Bar.NumberType = 0;
```
(It's possible, but it's convoluted. An easier way is to have a namespace with the same name as your class and typescript will merge both declarations!!)

- Both `namespace` and all-static `class` can have an internal state.

hl662 (approved):

PR #1380, opened 2025-07-21
[Tree-widget] Classifications tree filtering

JonasDov:
closes [#1363](https://github.com/iTwin/viewer-components-react/issues/1363)

PR #1382, opened 2025-07-22
[Tree-widget]: Format using prettier

JonasDov:
Creating due to: https://github.com/iTwin/viewer-components-react/pull/1380#issuecomment-3102268720


grigasp (approved):
thanks

Issue #1362, opened 2025-07-01
Tree widget: Unify visibility handlers


grigasp:
Currently all visibility trees use their own visibility handler implementation and the implementations have A LOT of duplicate code. We need to extract that common functionality into a base class to share code and simplify maintenance. 

The base implementation should handle (determine & control) visibility of common types:
- `bis.Model` (2d & 3d)
- `bis.Category` (2d & 3d)
- `bis.GeometricElement` (2d & 3d)

Then tree-specific implementations would reuse base implementation and add support for extra types on top:
- Models tree:
  - `bis.Subject`
  - Per-Model Category
- Categories tree:
  - `bis.DefinitionContainer`
  - `bis.SubCategory`
- Classifications tree:
  - `clsf.ClassificationTable`
  - `clsf.Classification`

We need to look into whether it would be possible to handle grouping nodes in a generic fashion.

grigasp:
@JonasDov I believe we can close this, right?

PR #1400, opened 2025-08-11
[Tree Widget]: Use Id64Arg in tree cache

JonasDov:
Changed models, categories and classifications trees cache methods to use `Id64Arg`.
Prerequisite for https://github.com/iTwin/viewer-components-react/issues/1362


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/Utils.ts
```diff
@@ -108,3 +108,8 @@ export function getClassesByView(viewType: "2d" | "3d") {
     ? ({ categoryClass: CLASS_NAME_DrawingCategory, elementClass: CLASS_NAME_GeometricElement2d, modelClass: CLASS_NAME_GeometricModel2d } as const)
     : ({ categoryClass: CLASS_NAME_SpatialCategory, elementClass: CLASS_NAME_GeometricElement3d, modelClass: CLASS_NAME_GeometricModel3d } as const);
 }
+
+/** @internal */
+export function getArrayFromId64Arg(arg: Id64Arg): Id64Array {
```
grigasp, line 113: As far as I can tell this function is used to create an array, which we later, in all cases, convert to a string. Would it make more sense to instead have `reduceId64Arg`? It would allow us to skip possibly unnecessary conversion from Set to Array.

PR #1404, opened 2025-08-12
Introduce @itwin/quantity-formatting-react

hl662:
TODO:
- [ ] Flesh out README.md with images.


grigasp (commented):
Could you add a gif or at least a static image of the components you're adding to the PR description?

packages/itwin/quantity-formatting/README.md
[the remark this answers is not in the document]
hl662: I haven't updated the README after moving it over from appui repo. README's been updated to not mention any 'improvements' or 'V2'

.github/CODEOWNERS
```diff
@@ -46,3 +46,6 @@
 
 /packages/itwin/map-layers  @itwin/viewer-components-reviewers @mdastous-bentley @eringram
 /change/@itwin-map-layers*  @itwin/viewer-components-reviewers @mdastous-bentley @eringram
+
+/packages/itwin/quantity-formatting  @itwin/viewer-components-reviewers @itwin/itwinjs-core-presentation @hl662
+/change/@itwin-quantity-formatting-react*  @itwin/viewer-components-reviewers @itwin/itwinjs-core-presentation
```
aruniverse, line 51:
```suggestion
/change/@itwin-quantity-formatting-react*  @itwin/viewer-components-reviewers @itwin/itwinjs-core-presentation @hl662 
```

packages/itwin/measure-tools/LICENSE.md
```diff
@@ -1,6 +1,6 @@
 # MIT License
 
-Copyright © 2021-2024 Bentley Systems, Incorporated. All rights reserved.
+Copyright © 2021-2025 Bentley Systems, Incorporated. All rights reserved.
```
aruniverse, line 3:
```suggestion
Copyright © Bentley Systems, Incorporated. All rights reserved.
```

packages/itwin/quantity-formatting/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "./node_modules/@itwin/build-tools/tsconfig-base.json",
+  "compilerOptions": {
+    "jsx": "react",
+    "module": "NodeNext",
+    "moduleResolution": "nodenext",
+    "target": "ES2024",
```
aruniverse, line 7: i would target es2023 like everything else

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "Formatting",
+    "Units",
+    "iTwin.js"
+  ],
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "peerDependencies": {
+    "@itwin/core-bentley": "^5.1.1",
```
aruniverse, line 46: does this need to be a peer?

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "@itwin/core-common": "^5.1.1",
```
aruniverse, line 47: this is only used for types, can be dev dep

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "@itwin/ecschema-metadata": "^5.1.1",
+    "@itwin/itwinui-react": "^3.16.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0"
+  },
+  "devDependencies": {
+    "@itwin/build-tools": "^5.1.1",
+    "@itwin/core-bentley": "^5.1.1",
+    "@itwin/core-common": "^5.1.1",
+    "@itwin/core-frontend": "^5.1.1",
```
aruniverse, line 59: make sure peers of core-frontend are defined as dev deps

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+  "author": {
+    "name": "Bentley Systems, Inc.",
+    "url": "http://www.bentley.com"
+  },
+  "peerDependencies": {
+    "@itwin/core-bentley": "^5.1.1",
+    "@itwin/core-common": "^5.1.1",
+    "@itwin/core-frontend": "^5.1.1",
+    "@itwin/core-quantity": "^5.1.1",
+    "@itwin/ecschema-metadata": "^5.1.1",
```
aruniverse, line 50: only used for types as well, can be devdep

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "@itwin/eslint-plugin": "^5.2.1",
+    "@itwin/itwinui-react": "^3.19.3",
+    "@testing-library/react": "^16.3.0",
+    "@testing-library/user-event": "^14.6.1",
+    "@types/react": "^19.1.9",
```
aruniverse, line 66: youre using react 19 types for a react 18 pkg
same thing goes for react-dom

packages/itwin/quantity-formatting/src/QuantityFormatting.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    QuantityFormatting._isInitialized = true;
+  }
+
+  /**
+   * Terminates the QuantityFormatting class and unregisters the localization namespace.
+   */
+  public static terminate(): void {
+    if (QuantityFormatting._isInitialized) {
+      IModelApp.localization.unregisterNamespace(QuantityFormatting._i18nNamespace);
```
aruniverse, line 65: Should this be QuantityFormatting._localization.unregister... ?

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSelector.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <ComboBox
+            id={formatListId}
+            className="format-selector-list"
+            options={formatOptions}
+            value={activeFormatDefinitionKey || ""}
+            onChange={handleFormatChange}
+            enableVirtualization
+            dropdownMenuProps={{
+              style: {
+                maxBlockSize: "min(calc((var(--iui-component-height) - 1px) * 30), 70vh)",
```
aruniverse, line 72: this is bit of a doozy to be hardcoding

packages/itwin/quantity-formatting/package.json
```diff
@@ -0,0 +1,87 @@
...
+  },
+  "type": "module",
+  "types": "./lib/quantity-formatting-react.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib/quantity-formatting-react.d.ts",
+      "default": "./lib/quantity-formatting-react.js"
+    }
+  },
+  "scripts": {
```
aruniverse, line 18: might be nice to setup api-extractor in this pkg. not needed in this pr, but def needed before 1.0

apps/test-viewer/src/components/QuantityFormatUiItemsProvider.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }, []);
+
+  const handleApply = useCallback(() => {
+    // Here you could apply the format changes to the active iModel
+    console.log("Applied format changes:", formatDefinition);
+    setIsModalOpen(false);
+  }, [formatDefinition]);
+
+  // Memoize the unitsProvider to prevent unnecessary re-renders
+  const memoizedUnitsProvider = React.useMemo(() => IModelApp.quantityFormatter.unitsProvider, []);
```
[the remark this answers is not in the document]
hl662: Right, updated to listen to when unitsProvider changes from IModelApp now

packages/itwin/quantity-formatting/tsconfig.json
hl662: changed

packages/itwin/quantity-formatting/package.json
hl662: nope, removed

packages/itwin/quantity-formatting/package.json
hl662: whoops, downgraded to v18

packages/itwin/quantity-formatting/src/QuantityFormatting.ts
hl662: yeah... switched

packages/itwin/quantity-formatting/package.json
hl662: i've defined them all as devDeps now

packages/itwin/quantity-formatting/package.json
aruniverse: will you be creating an issue for this if you dont do it in this pr?

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSelector.tsx
hl662: The [default](https://github.com/iTwin/iTwinUI/blob/bbf478f2bcae8fca3e483020e0e048aee37f4dde/packages/itwinui-css/src/menu/menu.scss#L26) is configured to only be around 8 items, and based on the number of KoQs any normal imodel has, 8 is not enough, so I increased it to 30 items. If the dropdown is larger then we restrict the height to be 70% of the screen's height. I can leave a comment explaining why this was done

apps/test-viewer/src/components/FormatManager.tsx
```diff
@@ -0,0 +1,216 @@
...
+            }
+          }
+        } catch (error) {
+          console.warn(`Schema ${schemaName} not found or failed to load:`, error);
+        }
+      }
+
+      // Get all used KindOfQuantities from the iModel, and populate the formatSet.
+      const reader = iModel.createQueryReader(
+        "SELECT\n  ks.Name || '.' || k.Name AS kindOfQuantityFullName,\n  COUNT(*) AS propertyCount,\n  json_group_array(p.Name) AS propertyNames\nFROM\n  ECDbMeta.ECPropertyDef p\n  JOIN ECDbMeta.KindOfQuantityDef k ON k.ECInstanceId = p.KindOfQuantity.Id\n  JOIN ECDbMeta.ECSchemaDef ks ON ks.ECInstanceId = k.Schema.Id\nGROUP BY\n  ks.Name,\n  k.Name\nORDER BY\n  propertyCount DESC;",
```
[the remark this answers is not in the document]
hl662: changed to use a template string, and formatted the query. Tried it out in the console too. I was thinking of turning this into a utils function, like `getAllUsedFormatsFromiModel` and put it in quantity-formatting-react, but it seems more useful to put it in core somewhere (a different question entirely, cuz it'll need IModelConnection interface, so only place to put is core-frontend :( )

hl662:
@grigasp If I moved the FormatManager to `@itwin/quantity-formatting-react`, could a new ContextMenuItem be created that can use it to update the format a Field description uses, similar to the [button](https://github.com/iTwin/viewer-components-react/pull/1404/files#diff-9731ccbdd29659138c34d1abf42ac55ab673581efa92a856ba9ad96c286f71c7R25-R28) I made in the test-viewer? All fields that formats their properties should have field.properties.koq right

grigasp (approved):
Reviewed API, package configuration and some top-level code of `@itwin/quantity-formatting-react`, also changes to test-viewer. Looks good.

packages/itwin/quantity-formatting/src/components/quantityformat/internal/StationBaseFactorPreset.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * Provides quick selection for common station configurations.
+ * @internal
+ */
+export function StationBaseFactorPreset(props: StationBaseFactorPresetProps) {
+  const { formatProps, onChange } = props;
+  const { translate } = useTranslation();
+
+  const currentBaseFactor = React.useMemo(() => {
+    return (formatProps as any).stationBaseFactor ?? 1;
+  }, [formatProps]);
```
saskliutas, line 29: No need to use `useMemo` for primitive values like numbers:
```suggestion
  const currentBaseFactor = (formatProps as any).stationBaseFactor ?? 1;
```

packages/itwin/quantity-formatting/src/components/quantityformat/internal/StationBaseFactorPreset.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const currentOffsetSize = React.useMemo(() => {
+    return formatProps.stationOffsetSize ?? 2;
+  }, [formatProps]);
```
saskliutas, line 33:
```suggestion
  const currentOffsetSize = formatProps.stationOffsetSize ?? 2;
```

packages/itwin/quantity-formatting/src/components/quantityformat/internal/StationBaseFactor.tsx
```diff
@@ -0,0 +1,110 @@
...
+ */
+export function StationBaseFactor(props: StationBaseFactorProps) {
+  const { formatProps, onChange } = props;
+  const { translate } = useTranslation();
+  const stationBaseFactorInputId = React.useId();
+
+  const currentValue = React.useMemo(() => {
+    // Default to 1 if no base factor is set
+    return (formatProps as any).stationBaseFactor ?? 1;
+  }, [formatProps]);
```
saskliutas, line 32:
```suggestion
  // Default to 1 if no base factor is set
  const currentValue = (formatProps as any).stationBaseFactor ?? 1;
```

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ScientificType.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/** Component to show/edit scientific type.
+ * @internal
+ */
+export function ScientificType(props: ScientificTypeProps) {
+  const { formatProps, onChange, disabled = false } = props;
+  const { translate } = useTranslation();
+
+  const handleSetFormatProps = React.useCallback(
```
saskliutas, line 32: Looks like this function can be inlined in `handleScientificTypeChange`. This applies to quite a few files in where pattern like this is used:

```tsx
const handleSet* = React.useCallback((newProps) => {
  onChange(newProps);
}, [deps]);

const handle*Change = React.useCallack(() => {
  // handle change
  handleSet*(newProps);
}, [otherDeps, handleSet*]);
```

In my opinion it adds unecessary complexity when reading code and could be inlined instead:
```tsx
const handle*Change = React.useCallack(() => {
  // handle change
  onChange(newProps);
}, [otherDeps, handleSet*]);
```

I would suggest going through all components and checking for functions defined using `useCallback` that are used only in other single function defined with `useCallback`.

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ThousandsSeparator.tsx
```diff
@@ -0,0 +1,171 @@
...
+/** Component to enable/disable the use of thousand separator.
+ * @internal
+ */
+export function UseThousandsSeparator(props: UseThousandsSeparatorProps) {
+  const { formatProps, onChange } = props;
+  const { translate } = useTranslation();
+
+  const useThousandsId = React.useId();
+
+  const handleSetFormatProps = React.useCallback(
```
saskliutas, line 32: Looks like this function can be inlined into `setFormatTrait`

packages/itwin/quantity-formatting/src/components/quantityformat/internal/misc/UnitDescr.tsx
```diff
@@ -0,0 +1,196 @@
...
+    readonly,
+    unitsProvider,
+  } = props;
+  const [unitOptions, setUnitOptions] = React.useState<SelectOption<string>[]>([
+    { value: name, label: getUnitName(name) },
+  ]);
+  const [currentUnit, setCurrentUnit] = React.useState({ name, label });
+  const { translate } = useTranslation();
+
+  const isMounted = React.useRef(false);
```
saskliutas, line 92: You should avoid `isMounted` pattern when using React hooks. There is no problem if `setState` is called on unmounted component as that will be noop. In this case it looks like you want to discard result of async operations started in `useEffect`. For that it's better to use local variable inside same `useEffect` that starts async operation:
```ts
useEffect(() => {
  let disposed = false;

  const doSomething = async () => {
    const result = await doWork(dep);
    if (disposed) {
      // ignore result because somponent is unmounted or `useEffect` was rerun with new `dep`
      return;
    }
    // handle result
  }

  void doSomething();
  return () => {
    disposed = true;
  }
}, [dep])
```

This way you will also avoid possible race conditions between `useEffect` runs with different dependencies. E.g:
1. Component rendered with `dep = 1` and `useEffect` starts an async operation `doWork(1)`
2. Component rerenders with `dep = 2` and `useEffect` starts another async operation `doWork(2)`.
3. `doWork(2)` resolves and component updates state accordingly.
4. `doWork(1)` resolves and component overrides state updated after `doWork(2)` and we end up situation where `dep = 2` but state is set to `dep = 1` result. (`isMounted` flag does nothing as it is set to `true` for both `useEffect` runs)

I saw `useEffects` with async operations in other components and they are also prone to race conditions on `useEffect` dependency changes so they probably should also be updated.

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ScientificType.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { SvgHelpCircularHollow } from "@itwin/itwinui-icons-react";
+import { useTranslation } from "../../../useTranslation.js";
+import "../FormatPanel.scss";
+
+/** Properties of [[ScientificType]] component.
+ * @internal
+ */
+export interface ScientificTypeProps {
+  formatProps: FormatProps;
+  onChange?: (format: FormatProps) => void;
```
saskliutas, line 21: Why is `onChange` optional?

packages/itwin/quantity-formatting/src/components/quantityformat/internal/misc/DecimalPrecision.tsx
```diff
@@ -0,0 +1,102 @@
...
+    },
+    {
+      value: 12,
+      label: translate("QuantityFormat:decimal_precision.twelve"),
+    },
+  ];
+
+  const handleOnChange = React.useCallback(
+    (newValue: number) => {
+      onChange && onChange(newValue);
```
saskliutas, line 87: Quite a few components under `misc` folder has code like `onChange && onChange(//)` but `onChange` prop is not optional. I think in that case `handleOnChange` should be removed and `onChange` should be passed instead.

saskliutas (commented):
Looked over React components code. Added some comments that apply to multiple files but didn't want to spam same comment on each file.

grigasp:
A some questions regarding layout of the dialog:
image

1. Why does it take full height of the screen? I think it should:
   - take as little vertical space as possible, to accommodate its content,
   - as the amount of content grows, grow vertically until ~80% of the vertical screen size, then start showing the scrollbar.

2. Why is the preview block narrower than the other blocks?

3. Why do the blocks have so much left/right padding? Why aren't they in line with "Available Formats" section?

In short, seems like way too much blank space in that dialog.

hl662:
> In short, seems like way too much blank space in that dialog.

Crux of getting something quick to demonstrate functionality.

I've edited the styling of the component housed in test-viewer, and updated styling in the components exported by the package themselves to be more aligned:

image


image

hl662:
> Why is the preview block narrower than the other blocks?
It's a flex box that expands as the formatted value increases in size, it's meant to draw less attention than panels below it, which serves the largest purpose (configuring formatting)

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ScientificType.tsx
hl662: that's an outdated thing from the old component, I've changed it to be required

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ThousandsSeparator.tsx
hl662: thanks, made the change here and in other components that uses setFormatTrait

packages/itwin/quantity-formatting/src/components/quantityformat/internal/ScientificType.tsx
hl662: thanks, made the change here and in other components

packages/itwin/quantity-formatting/src/components/quantityformat/QuantityFormatPanel.tsx
```diff
@@ -0,0 +1,129 @@
...
+  } = props;
+
+  const { translate } = useTranslation();
+
+  // Clone the formatDefinition to work with internally
+  const [clonedFormatDefinition, setClonedFormatDefinition] = React.useState<FormatDefinition>(() =>
+  ({ ...formatDefinition })
+  );
+
+  const [_saveEnabled, _setSaveEnabled] = React.useState(false);
```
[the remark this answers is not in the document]
hl662: no reason in particular, i removed the prefix

packages/itwin/quantity-formatting/src/components/quantityformat/FormatPanel.tsx
```diff
@@ -0,0 +1,157 @@
...
+  getBearingSecondaryChildren,
+} from "./panels/Bearing.js";
+import "./FormatPanel.scss";
+import {
+  getRatioPrimaryChildren,
+  getRatioSecondaryChildren,
+} from "./panels/Ratio.js";
+
+// Temporary translation function - this should be replaced with proper i18n
+function useTranslation() {
```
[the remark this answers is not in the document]
hl662: I added the useTranslation.ts after this component was done,  forgot to switch it. Fixed it now

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSelector.tsx
hl662: thanks, your recommendation got the className approach to work. I reduced it to 20.5, and got rid of 70 vh

packages/itwin/quantity-formatting/src/components/quantityformat/internal/misc/UnitDescr.tsx
```diff
@@ -0,0 +1,191 @@
...
+    void fetchAllowableUnitSelections();
+
+    return () => {
+      disposed = true;
+    };
+  }, [index, label, name, parentUnitName, translate, unitsProvider]);
+
+  const handleOnUnitChange = React.useCallback(
+    (newValue: string) => {
+      onUnitChange && onUnitChange(newValue, index);
```
saskliutas, line 160:
```suggestion
      onUnitChange(newValue, index);
```

packages/itwin/quantity-formatting/src/components/quantityformat/internal/misc/UnitDescr.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+    [index, onUnitChange]
+  );
+
+  const handleOnLabelChange = React.useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      e.preventDefault();
+      onLabelChange && onLabelChange(e.target.value, index);
```
saskliutas, line 168:
```suggestion
      onLabelChange(e.target.value, index);
```

bsy-nicholasw (approved):

PR #1430, opened 2025-08-28
Update mapLayers.json

zhang-zimin:
Added an entry for MapboxImagery to the localization JSON files.


zhang-zimin:
image

Localized WellKnownBaseMaps are missing Mapbox entries.

PR #1432, opened 2025-08-28
[map-layers] Fix settings popup height

wilmaier:
Current QA version does not show all settings without needing to scroll:
image

Prod version shows all settings with no scrollbar:
image

With this change it looks like this:
image


wilmaier:
That worked - looks like they setup the middleware just to allow someone to override their 400px default.

PR #1456, opened 2025-09-19
Add FormatSetSelector, FormatSetPanel, make UI more consistent

hl662:
- Adds exported FormatSetSelector, FormatSetPanel components to @itwin/quantity-formatting-react. FormatSetPanel component can set an `editable` boolean flag to allow users to edit a format set.

- Update FormatPanel, QuantityFormatPanel to be more consistent UI-wise.


apps/test-viewer/src/components/quantity-formatting/FormatSetsTabPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+    formats: {},
+  },
+];
+
+/** Tab panel component for Format Sets management */
+export const FormatSetsTabPanel: React.FC = () => {
+  const [formatSets, setFormatSets] = React.useState<FormatSet[]>([]);
+  const [activeFormatSetKey, setActiveFormatSetKey] = React.useState<string | undefined>();
+  const [selectedFormatSetKey, setSelectedFormatSetKey] = React.useState<string | undefined>();
+  const [selectedFormatSet, setSelectedFormatSet] = React.useState<FormatSet | undefined>();
```
saskliutas, line 44: Why do you need to track `selectedFormatSetKey` and `selectedFormatSet` separately? Wouldn't it be enough to track `selectedFormatSet`?

apps/test-viewer/src/components/quantity-formatting/FormatSetsTabPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+      return () => {
+        removeListener();
+      };
+    } else {
+      // If no FormatManager, just use test format sets
+      setFormatSets(testFormatSets);
+    }
+  }, []);
+
+  const handleFormatSetChange = React.useCallback((formatSet: FormatSet, key: string) => {
```
saskliutas, line 83: Maybe rename this to `handleSelectedFortmatSetChange`. As it is very confusing having `handleFormatSetChange` and `handleFormatSetChanged`.

apps/test-viewer/src/components/quantity-formatting/FormatSetsTabPanel.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [formatSets, setFormatSets] = React.useState<FormatSet[]>([]);
+  const [activeFormatSetKey, setActiveFormatSetKey] = React.useState<string | undefined>();
+  const [selectedFormatSetKey, setSelectedFormatSetKey] = React.useState<string | undefined>();
+  const [selectedFormatSet, setSelectedFormatSet] = React.useState<FormatSet | undefined>();
+  const [clonedSelectedFormatSet, setClonedSelectedFormatSet] = React.useState<FormatSet | undefined>();
+  const [saveEnabled, setSaveEnabled] = React.useState(false);
+
+  // Load format sets from FormatManager
+  React.useEffect(() => {
+    const formatManager = FormatManager.instance;
```
saskliutas, line 50: Is there any reason to render this component if `FormatManager.instance` is `undefined`? I think it would much cleaner if we could avoid all these undefined check with something like:
```tsx
export function FormatSetsTabPanel() {
  if (FormatManager.instance === undefined) {
    return null;
  }
  return <FormatSetsTabPanelImpl formatManager={FortmatManager.instance} />
}

function FormatSetsTabPanelImpl({ formatManager }: { formatManager: FormatManager }) {
  // render format components
}
```

Maybe this could be even handled one layer above in `QuantityFormatButton`?

apps/test-viewer/src/components/quantity-formatting/FormatPropertyContextMenuItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+    UiFramework.dialogs.modal.close();
+  }, []);
+
+  React.useEffect(() => {
+    const _removeListener = IModelApp.quantityFormatter.onUnitsProviderChanged.addListener(() => {
+      setUnitsProvider(IModelApp.quantityFormatter.unitsProvider);
+    });
+    return () => {
+      _removeListener();
+    };
```
saskliutas, line 70: I noticed this pattern repeats in several places but it can be simplified:

```suggestion
    return IModelApp.quantityFormatter.onUnitsProviderChanged.addListener(() => {
      setUnitsProvider(IModelApp.quantityFormatter.unitsProvider);
    });
```

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSetSelector.tsx
```diff
@@ -0,0 +1,99 @@
...
+      setSearchTerm(event.target.value);
+    },
+    []
+  );
+
+  return (
+    <Flex flexDirection="column" gap="none" alignItems="flex-start" className="quantityFormat--formatSetSelector-container">
+      <Input
+        value={searchTerm}
+        onChange={handleSearchChange}
```
saskliutas, line 65: I think this can be inlined `onChange={(e) => setSearchTerm(e.currentTarget.value)}`

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSetPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+    if (editable && formatSet && onFormatSetChange) {
+        const updatedFormatSet: FormatSet = {
+          ...formatSet,
+          unitSystem: value as UnitSystemKey,
+        };
+        onFormatSetChange(updatedFormatSet);
+    }
+  }, [editable, formatSet, onFormatSetChange]);
+
+  if (!formatSet) {
```
saskliutas, line 94: Why does this component allows passing `undefined` `formatSet`? I think message about selecting format set should be rendered by the parent of this component.

packages/itwin/quantity-formatting/src/components/quantityformat/FormatSetPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { ComboBox, Flex, IconButton, Input, Label, Text, Textarea } from "@itwin/itwinui-react";
+import { useTranslation } from "../../useTranslation.js";
+
+import type { UnitSystemKey } from "@itwin/core-quantity";
+import type { FormatSet } from "@itwin/ecschema-metadata";
+/**
+ * Props for the FormatSetPanel component
+ * @internal
+ */
+interface FormatSetPanelProps {
```
saskliutas, line 17: I think props could be imroved to require `onFormatSetChange` when `editable` is `true`:
```ts
type FormatSetPanelProps = {
  formatSet: FormatSet;
} & ({ 
  editable: true,
  onFormatSetChange: (formatSet: FormatSet) => void 
} | {
  editable?: false,
  onFormatSetChange?: undefined 
})
```

apps/test-viewer/src/components/quantity-formatting/FormatPropertyContextMenuItem.tsx
hl662: thanks, i've updated here and across other spots

apps/test-viewer/src/components/quantity-formatting/FormatSetsTabPanel.tsx
hl662: yeah... removed

Issue #1411, opened 2025-08-18
Tree widget: Performance issues when loading Models tree


DenisMicha:
**Package**
@itwin/tree-widget-react 3.11.0

**Describe the bug**
On bigger iModels it takes quite a lot of time to load the nodes in Models tree after opening iModel, and even after part of elements appear in 3D view, in Models tree all nodes are still marked as invisible for some time. In one of the examples it took more than 5 seconds to load the nodes, and even after that, when a lot elements were already visible in 3D view, all nodes where still marked as invisible for 2-3 seconds.

**To Reproduce**
Open any bigger iModel with immediately opened tree view

**Expected behavior**
After opening iModel, tree view is always displayed immediately/almost immediately, and when part of elements is loaded in 3D view, in tree view respective nodes are marked as visible/partially visible.

**Desktop:**
- iTwin.js 5.0.4

DenisMicha:
Could you please advise how to get that queries info you mentioned from the ModelsTree? I have never done this before. Or I can share the private information about the specific iModel separately.

DenisMicha:
@grigasp As discussed, getting back to you with UX improvement suggestion. Currently, when we expand some node and its subnode is loading, instead of the eye and chevron a small spinner is visible. Maybe the same could be done for initially visible nodes while their eye states are still loading? Which is after the tree becomes visible, but the eyes are displayed as hidden for a couple of seconds.

grigasp:
> [@grigasp](https://github.com/grigasp) As discussed, getting back to you with UX improvement suggestion. Currently, when we expand some node and its subnode is loading, instead of the eye and chevron a small spinner is visible. Maybe the same could be done for initially visible nodes while their eye states are still loading? Which is after the tree becomes visible, but the eyes are displayed as hidden for a couple of seconds.

You mean show a spinner instead of a disabled "eye" next to each node?

grigasp:
Any chance you could provide a low-fidelity mockup for a situation when there are multiple nodes in the initial tree view and we're loading visibility states for them?

DenisMicha:
Basically instead of this:

<img width="292" height="84" alt="Image" src="https://github.com/user-attachments/assets/65de5716-6cf6-4bdc-8897-6642b356fec4" />

Would be this:

<img width="263" height="89" alt="Image" src="https://github.com/user-attachments/assets/15b5d7b2-911f-4aa4-b7f8-a1839abf814f" />

grigasp:
[...]

PR #1458, opened 2025-09-24
Tree widget: Add "loading" state to `TreeNodeCheckboxState` type

grigasp:
closes https://github.com/iTwin/viewer-components-react/issues/1411


saskliutas (approved):
LGTM. Have you checked if there is no flickering when visibility is changed outside tree and checkboxes state is updated?

Issue #1389, opened 2025-08-01
Map layers - not use inline styles for onMouseEnter/Leave


gedkek:
Using map-layers@6.0.1 in electron application;
Our platforming apps content security policy blocks inline styles (used by map layers for hovering over external layers), hence not showing appropriate buttons for menus.

<img width="1149" height="231" alt="Image" src="https://github.com/user-attachments/assets/bd758044-ddb1-4e4f-be92-1cdb426c08cc" />

Could this be fixed from your side?

andremig-bentley:
@itwin/map-layers version 6.0.8 has been published with the fix for this issue.

PR #1463, opened 2025-09-29
Remove Dynamically Created Inline Styles from Map Layers

andremig-bentley:
Resolves https://github.com/iTwin/viewer-components-react/issues/1389.

As previously addressed in [this PR](https://github.com/iTwin/viewer-components-react/pull/1455), some apps' content security policies do not allow inline styles. That PR attempted to fix the bug described in the attached issue by converting all instances of inline styles in map-layers with corresponding css classes, however it missed the inline styles created in the `changeVisibilityByElementId()` function. This function was used to override existing styles in order to allow certain menu items to appear when hovering over an attached map layer. This PR addresses this issue by removing `changeVisibilityByElementId()` and letting css handle the hover functionality. Attached below is a clip of the hover functionality working with these changes implemented.

https://github.com/user-attachments/assets/41ce6e09-1e29-418b-ac8e-29cb8554a7dd

### DIFF
--- a/packages/itwin/map-layers/src/ui/widget/MapLayerDroppable.scss
+++ b/packages/itwin/map-layers/src/ui/widget/MapLayerDroppable.scss
@@ -10,3 +10,18 @@
 .map-manager-display-none {
   display: none;
 }
+
+// Hover-based visibility for settings menus
+.map-manager-source-item {
+  #MapLayerSettingsMenuWrapper,
+  #MapLayerSettingsSubLayersMenu {
+    visibility: hidden;
+  }
+
+  &:hover {
+    #MapLayerSettingsMenuWrapper,
+    #MapLayerSettingsSubLayersMenu {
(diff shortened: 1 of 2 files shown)


aruniverse:
please test out with the test app in this repo and turn on the CSP settings that caught this

scripts/check-inline-styles.js
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Script to check for inline styles
+ * This validates that the map-layers package contains no inline style attributes
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+// Function to recursively find all .ts and .tsx files
+function findTsxFiles(dir) {
```
aruniverse, line 12: why couldnt we just use a glob search?

andremig-bentley: We could, done in latest commit.

Issue #1445, opened 2025-09-10
Tree widget: Provide viewport abstraction for visibility controls


grigasp:
At the moment all tree widget's visibility trees expect a `Viewport` from iTwin.js to determine and control elements' visibility. Some of our consumers want to use the component with other kinds of viewports, e.g. Cesium or Unreal - need to provide them this option.

The viewport abstraction has to be documented very clearly and explain how model / category / per-model-category / element display priorities work.

PR #1466, opened 2025-10-06
[Tree widget]: Provide viewport abstraction for visibility controls

JonasDov:
closes [#1445](https://github.com/iTwin/viewer-components-react/issues/1445)

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/CategoriesVisibilityUtils.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/CategoriesVisibilityUtils.test.ts
@@ -202,7 +202,6 @@ describe("CategoryVisibilityUtils", () => {
     });
 
     after(async function () {
-      nonMockedViewport[Symbol.dispose]();
       await imodel.close();
       await terminatePresentationTesting();
     });
--- a/packages/itwin/tree-widget/src/test/trees/Common.ts
+++ b/packages/itwin/tree-widget/src/test/trees/Common.ts
@@ -97,7 +97,6 @@ export function createFakeSinonViewport(
     }),
     iModel: createIModelMock({ queryHandler: props?.queryHandler, getCategoryInfo: props?.getCategoryInfo }),
     renderFrame: sinon.fake,
-    [Symbol.dispose]() {},
   };
   return result;
 }
--- a/packages/itwin/tree-widget/src/test/trees/TreeUtils.ts
+++ b/packages/itwin/tree-widget/src/test/trees/TreeUtils.ts
@@ -43,7 +43,7 @@ function getDefaultSubCategoryId(categoryId: Id64String) {
 
 export type TreeWidgetTestingViewport = TreeWidgetViewport & {
   renderFrame: () => void;
-} & Disposable;
+};
 
 export function createTreeWidgetTestingViewport({
   iModel,
@@ -238,14 +238,6 @@ export function createTreeWidgetTestingViewport({
         onPerModelCategoriesOverridesChanged.shouldFireOnRender = false;
         onPerModelCategoriesOverridesChanged.event.raiseEvent();
       }
-    },
-    [Symbol.dispose]() {
-      onPerModelCategoriesOverridesChanged.event.clear();
-      onDisplayedCategoriesChanged.event.clear();
-      onDisplayStyleChanged.clear();
-      onDisplayedModelsChanged.event.clear();
-      onAlwaysDrawnChanged.event.clear();
-      onNeverDrawnChanged.event.clear();
     },
   };
 }
--- a/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
@@ -137,7 +137,6 @@ describe("CategoriesTreeVisibilityHandler", () => {
       ...commonProps,
       [Symbol.dispose]() {
         commonProps.idsCache[Symbol.dispose]();
-        commonProps.viewport[Symbol.dispose]();
         handler[Symbol.dispose]();
         provider[Symbol.dispose]();
       },
--- a/packages/itwin/tree-widget/src/test/trees/classifications-tree/ClassificationsTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/classifications-tree/ClassificationsTreeVisibilityHandler.test.ts
@@ -84,7 +84,6 @@ describe("ClassificationsTreeVisibilityHandler", () => {
       viewport,
       [Symbol.dispose]() {
         idsCache[Symbol.dispose]();
-        viewport[Symbol.dispose]();
         handler[Symbol.dispose]();
         provider[Symbol.dispose]();
       },
@@ -798,7 +797,6 @@ describe("ClassificationsTreeVisibilityHandler", () => {
         viewport,
         [Symbol.dispose]() {
           idsCache[Symbol.dispose]();
-          viewport[Symbol.dispose]();
           handler[Symbol.dispose]();
           defaultProvider[Symbol.dispose]();
           filteredProvider[Symbol.dispose]();
--- a/packages/itwin/tree-widget/src/test/trees/models-tree/UseModelsTree.test.tsx
+++ b/packages/itwin/tree-widget/src/test/trees/models-tree/UseModelsTree.test.tsx
@@ -162,7 +162,6 @@ describe("useModelsTree", () => {
       });
 
       after(async function () {
-        viewport[Symbol.dispose]();
         await imodel.close();
         await terminatePresentationTesting();
         TreeWidget.terminate();
--- a/packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
@@ -1774,7 +1774,6 @@ describe("ModelsTreeVisibilityHandler", () => {
         hierarchyConfig: commonProps.hierarchyConfig,
         [Symbol.dispose]() {
           commonProps.idsCache[Symbol.dispose]();
-          commonProps.viewport[Symbol.dispose]();
           handler[Symbol.dispose]();
           provider[Symbol.dispose]();
         },
@@ -3390,7 +3389,6 @@ describe("ModelsTreeVisibilityHandler", () => {
             commonProps.idsCache[Symbol.dispose]();
             handler[Symbol.dispose]();
             defaultProvider[Symbol.dispose]();
-            commonProps.viewport[Symbol.dispose]();
             filteredProvider[Symbol.dispose]();
           },
         };


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,243 @@
...
+ * 5. Category and sub-category visibility - if category or sub-category is not visible, elements with that category/sub-category should not be displayed.
+ *
+ * Based on this order of precedence, element can only be displayed in these scenarios:
+ * - Model is visible *AND* element is in [[alwaysDrawn]] list.
+ * - Model is visible *AND* element is not in [[neverDrawn]] list *AND* per-model category override is set to show `Show`.
+ * - Model is visible *AND* element is not in [[neverDrawn]] list *AND* per-model category override is not set to `Hide` *AND* category and sub-category are visible.
+ * @alpha
+ */
+export interface TreeWidgetViewport {
+  /** The type of the view. Generally it should be either `spatial` or `2d`. */
```
JonasDov, line 94: @grigasp any suggestions how to make this comment better?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/Utils.ts
```diff
@@ -24,8 +26,9 @@ export type FunctionProps<THook extends (props: any) => any> = Parameters<THook>
  * Disables display of all given models.
  * @public
  */
-export async function hideAllModels(models: string[], viewport: Viewport) {
-  viewport.changeModelDisplay(models, false);
+export async function hideAllModels(models: string[], viewport: Viewport | TreeWidgetViewport) {
+  const treeWidgetViewport = isTreeWidgetViewport(viewport) ? viewport : createTreeWidgetViewport(viewport);
```
grigasp, line 30: this gets repeated a lot, maybe `createTreeWidgetViewport` could take `Viewport | TreeWidgetViewport` and take care of this? then we could simply do this:
```
createTreeWidgetViewport(viewport).changeModelDisplay({ modelIds: models, display: false });
```

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -124,10 +124,74 @@ export async function createViewport({
...
-  });
+  return createTreeWidgetTestingViewport(
+    OffScreenViewport.create({
+      view: viewState,
+      viewRect: new ViewRect(),
+    }),
+  );
+}
+
+/** @internal */
```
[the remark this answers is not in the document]
JonasDov: Removed internal tags from this file apis

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
```diff
@@ -124,10 +124,74 @@ export async function createViewport({
...
+export type TreeWidgetTestingViewport = TreeWidgetViewport & { renderFrame: () => void } & Disposable;
+
+/** @internal */
+export function createTreeWidgetTestingViewport(viewport: Viewport): TreeWidgetTestingViewport {
```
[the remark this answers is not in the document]
JonasDov: Not sure what you mean.
We are using a real (non mocked) `Viewport` here: we are creating `TreeWidgetTestingViewport` using the provided `Viewport`. The purpose of this function is to create `TreeWidgetViewport`, while also to be able to dispose original viewport and to be able to call `renderFrame` on the original viewport.
Not sure which path you want us to go with:
1. Make this function take `TreeWidgetViewport` - in this case the function loses its meaning. We can't call `renderFrame` on `TreeWidgetViewport` and we can't dispose it. We have to get `Viewport` to be able to do this. So we would have to create `TreeWidgetViewport` in each place where `createTreeWidgetTestingViewport` is called and then use the original `Viewport` when we want to call `renderFrame` or when we are finished and want to dispose it.
2. Make this function not accept `Viewport` argument at all. In some cases this viewport is created asynchronously (e.g. `CategoriesTreeVisibilityHandler.test.ts`), in others it is created synchronously (e.g. `ModelsTreeVisibilityHandler.test.ts`). `Viewport` is also created using different configurations. This means that we would need to accept `Viewports` configurations inside this function and make it asynchronous. 
3. Leave it as is. I think this is the best way to go: 1. Each test suite can create `Viewport` based on what it needs.  2. We have the `[Symbol.dispose]` and `renderFrame` functions available so we can use the `TreeWidgetTestingViewport` when disposing or when calling `renderFrame`.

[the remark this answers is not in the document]
JonasDov: Changed `createTreeWidgetTestingViewport` to accept `viewState`. Didn't need to make it async.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/UseActiveViewport.tsx
```diff
@@ -4,21 +4,33 @@
...
-  }, []);
+    if (treeWidgetViewport) {
+      setActiveViewport(treeWidgetViewport);
+      return;
+    } else {
+      setActiveViewport(createTreeWidgetViewportInternal(IModelApp.viewManager.selectedView));
+      return IModelApp.viewManager.onSelectedViewportChanged.addListener((args) => {
+        setActiveViewport(createTreeWidgetViewportInternal(args.current));
+      });
+    }
```
grigasp, line 28: don't need an `else` if there's a `return`


```suggestion
    }
    setActiveViewport(createTreeWidgetViewportInternal(IModelApp.viewManager.selectedView));
    return IModelApp.viewManager.onSelectedViewportChanged.addListener((args) => {
      setActiveViewport(createTreeWidgetViewportInternal(args.current));
    });
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeComponent.tsx
```diff
@@ -42,6 +43,7 @@ interface CategoriesTreeComponentProps
    * ```
    */
   headerButtons?: Array<(props: CategoriesTreeHeaderButtonProps) => React.ReactNode>;
+  treeWidgetViewport?: TreeWidgetViewport;
```
grigasp, line 46: I'd say this should be changed to `viewport?: Viewport | TreeWidgetViewport`. Also, let's document what it means and what happens when it's not provided.

grigasp: same for other tree components

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
```diff
@@ -52,17 +54,21 @@ export interface CategoriesTreeHeaderButtonProps extends TreeToolbarButtonProps
...
   buttonProps: Pick<CategoriesTreeHeaderButtonProps, "categories" | "viewport" | "models">;
   onCategoriesFiltered: (props: { categories: CategoryInfo[] | undefined; models?: Id64Array }) => void;
 } {
   const [filteredCategories, setFilteredCategories] = useState<CategoryInfo[] | undefined>();
   const [filteredModels, setFilteredModels] = useState<Id64Array | undefined>();
-  const categories = useCategories(viewport);
-  const models = useAvailableModels(viewport);
+
+  const treeWidgetViewport = useMemo(() => {
+    return createTreeWidgetViewport(viewport);
```
saskliutas, line 65: Can we handle this somewhere in components higher up instead of doing it in all these buttons? I think `createTreeWidgetViewport` should be called one time above `Tree` and buttons and all buttons and tree renderers should take `TreeWidgetViewport`.

saskliutas: But these changes are in the next branch so we can make breaking changes. 🤔

saskliutas: I don't think these buttons should be used separately without tree component.

grigasp: ah, completely forgot this is on `tree-widget/next`... in that case, I'd like us to consider going even further and taking `TreeWidgetViewport` everywhere, including the top level components, and exposing the `createTreeWidgetViewport` through the public API.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+   * - `true`, model display should be turned on. This does not mean that any elements should be turned on, it only means that
+   * model visibility should not interfere with elements visibility.
+   * - `false`, model display should be turned off. All elements which have that model should not be displayed in the viewport.
+   * @param modelIds The Ids of the models to add or remove.
+   * @param display Whether or not to display the specified models in the viewport.
+   * @returns
+   * - `false` if change is unsuccessful;
+   * - `true` if change is successful;
+   */
+  changeModelDisplay: (props: { modelIds: Id64Arg; display: boolean }) => boolean;
```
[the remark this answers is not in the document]
grigasp: after a second look, I think we should have only one function to change model display in this interface. the use of `addViewedModels` vs `changeModelDisplay` should be an internal detail of the implementation. based on the docs, it sounds like we should use `addViewedModels` every time we enable display, and this is actually not always the case currently.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Should return the per-model category override. */
+  getPerModelCategoryOverride: (props: { modelId: Id64String; categoryId: Id64String }) => PerModelCategoryVisibility.Override;
+  /** Should remove per-model category override for the specified models. */
+  clearPerModelCategoryOverrides: (modelIds?: Id64Arg) => void;
+  /**
+   * Ids of elements which should not be displayed.
+   *
+   * This set takes precedence over category and sub-category display and per-model category override.
+   *
+   * **NOTE** This set should take precedence over the [[alwaysDrawn]] set - if an element is present in both sets, it is never drawn.
```
grigasp, line 192:
```suggestion
   * This set takes precedence over category and sub-category display, per-model category overrides, and the `alwaysDrawn` set.
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+   */
+  neverDrawn: ReadonlySet<Id64String> | undefined;
+  /**
+   * Ids of elements which should not be displayed in the viewport, regardless of category and sub-category visibility.
+   *
+   * @param elementIds The Ids of the elements that should not be displayed.
+   */
+  setNeverDrawn: (elementIds: Set<Id64String>) => void;
```
grigasp, line 200:
```suggestion
  /** Ids of elements which should not be displayed in the viewport, regardless of category and sub-category visibility. */
  setNeverDrawn: (elementIds: Set<Id64String>) => void;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Should clear the set of never-drawn elements. */
+  clearNeverDrawn: () => void;
+  /**
+   * Ids of elements which should be displayed.
+   *
+   * Elements in this set should be displayed only when model is displayed as well.
+   *
+   * This set takes precedence over category and sub-category display and per-model category override.
+   *
+   * **NOTE** The [[neverDrawn]] set takes precedence - if an element is present in both sets, it is never drawn.
```
grigasp, line 210:
```suggestion
   * This set takes precedence over category and sub-category display and per-model category override.
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+   */
+  alwaysDrawn: ReadonlySet<Id64String> | undefined;
+  /**
+   * Ids of elements which should be displayed in the viewport, regardless of category and sub-category visibility.
+   *
+   * These elements should be displayed only when their models are displayed as well.
+   * @param elementIds The Ids of the elements that should be displayed.
+   * @param exclusive If true, *only* the specified elements should be drawn. When this is set to true `isAlwaysDrawnExclusive` should return true, otherwise `isAlwaysDrawnExclusive` should return false.
+   */
+  setAlwaysDrawn: (props: { elementIds: Set<Id64String>; exclusive?: boolean }) => void;
```
grigasp, line 220:
```suggestion
   * These elements should be displayed only when their models are displayed as well.
   *
   * When the `exclusive` prop is set to `true`, only the specified elements should be drawn.
   */
  setAlwaysDrawn: (props: { elementIds: Set<Id64String>; exclusive?: boolean }) => void;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Should clear the set of always-drawn elements. */
+  clearAlwaysDrawn: () => void;
+  /**
+   * Should return true if elements in the [[alwaysDrawn]] set are the *only* elements that are displayed when their models are visible.
+   *
+   * Should be set to true when `setAlwaysDrawn` is called with `exclusive` flag.
+   *
+   * Should be set to false when `setAlwaysDrawn` is called with `exclusive` set to false/undefined or when `clearAlwaysDrawn` is called.
+   */
+  readonly isAlwaysDrawnExclusive: boolean;
```
grigasp, line 230:
```suggestion
   * Should be set to true when `setAlwaysDrawn` is called with `exclusive` flag. Should be set to false 
   * when `setAlwaysDrawn` is called with `exclusive` set to `false` or `undefined`, or when `clearAlwaysDrawn` is called.
   */
  readonly isAlwaysDrawnExclusive: boolean;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+  /** Event that should be raised when set of never-drawn elements changes. */
+  onNeverDrawnChanged: BeEvent<() => void>;
+  /** Event that should be raised when per-model category overrides changes. */
+  onPerModelCategoriesOverridesChanged: BeEvent<() => void>;
+  /** Event that should be raised when category display changes. */
+  onDisplayedCategoriesChanged: BeEvent<() => void>;
+  /** Event that should be raised when model display changes. */
+  onDisplayedModelsChanged: BeEvent<() => void>;
+  /**
+   * Event that should be raised when model, category or sub-category display changes.
```
grigasp, line 242: is this really true, meaning that changing model display fires `onDisplayedModelsChanged` and this one too?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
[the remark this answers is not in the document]
grigasp: do we need to use the `Viewport` or `ViewState` at all? when creating a test view state, we have to provide a ton of values that we don't care about - it would be simpler to just avoid that altogether.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+   *
+   * The returned value can be a custom object that has `Symbol.iterator` defined, or it can be a built in type like `Array` or `Set`.
+   */
+  perModelCategoryOverrides: Readonly<Iterable<{ modelId: Id64String; categoryId: Id64String; visible: boolean }>>;
+  /** Should change the override state of one or more categories for one or more models. */
+  setPerModelCategoryOverride: (props: { modelIds: Id64Arg; categoryIds: Id64Arg; override: PerModelCategoryVisibility.Override }) => void;
+  /** Should return the per-model category override. */
+  getPerModelCategoryOverride: (props: { modelId: Id64String; categoryId: Id64String }) => PerModelCategoryVisibility.Override;
+  /** Should remove per-model category override for the specified models. */
+  clearPerModelCategoryOverrides: (modelIds?: Id64Arg) => void;
```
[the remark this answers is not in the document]
JonasDov: Changed to use props.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
```diff
@@ -0,0 +1,247 @@
...
+  /**
+   * Returns iterable visibility overrides objects.
+   *
+   * The returned value can be a custom object that has `Symbol.iterator` defined, or it can be a built in type like `Array` or `Set`.
+   */
+  perModelCategoryOverrides: Readonly<Iterable<{ modelId: Id64String; categoryId: Id64String; visible: boolean }>>;
+  /** Should change the override state of one or more categories for one or more models. */
+  setPerModelCategoryOverride: (props: { modelIds: Id64Arg; categoryIds: Id64Arg; override: PerModelCategoryVisibility.Override }) => void;
+  /** Should return the per-model category override. */
+  getPerModelCategoryOverride: (props: { modelId: Id64String; categoryId: Id64String }) => PerModelCategoryVisibility.Override;
```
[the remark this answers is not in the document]
JonasDov: Changed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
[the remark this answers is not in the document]
JonasDov: Removed `addViewedModels` from API. Now `changeModelDisplay` is async and underneath it calls `addViewedModels` if `display` is true and `changeModelDisplay` if `display` is false.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/TreeWidgetViewport.ts
JonasDov: Changed `spatial` to `3d`. Changed comment to:
```ts
/**
 * The type of the view. Generally it should be either `3d` or `2d`.
 *
 * The `other` value is reserved for edge cases when viewport doesn't show any elements, e.g. a blank viewport.
 */
```

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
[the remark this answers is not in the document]
grigasp: yes, that is my idea

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
grigasp: now that we take `TreeWidgetViewport` everywhere, do we need `createTreeWidgetViewport` to take `Viewport | TreeWidgetViewport`? maybe we could just take `Viewport`?

apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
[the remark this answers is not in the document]
grigasp: note: only for tests in tree-widget package. perf tests should use the real `Viewport`

change/@itwin-tree-widget-react-606292c5-0d0f-4748-ad83-50279476f61b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Tree Widget: changed `CategoriesTreeComponentProps`, `ClassificationsTreeComponentProps` and `ModelsTreeComponentProps` to now accept optional viewport of `TreeWidgetViewport` type. It will be used instead of the `Viewport` from `@itwin/core-frontend`. Also, changed `TreeToolbarButtonProps`, `useCategoriesTreeButtonProps`, `UseCategoriesTreeProps`, `UseClassificationsTreeProps`, `useModelsTreeButtonProps` and `UseModelsTreeProps` to accept `TreeWidgetViewport` instead of `Viewport`. To migrate, use `createTreeWidgetViewport` and provide previously used `Viewport` from `@itwin/core-frontend`.",
```
grigasp, line 3: - No need to prefix with "Tree widget" - we're in tree widget package.
- I'd rather mention components rather than their props.
- Don't think we need to exlicitly name every component that this change affected. Instead, let's focus on the change itself.

```suggestion
  "comment": "Changed props of `CategoriesTreeComponent`, `ClassificationsTreeComponent` and `ModelsTreeComponent` to accept an optional viewport of `TreeWidgetViewport` type, that would be used instead of the active AppUI's viewport. Also, changed props of other components that used to take `Viewport` to now take `TreeWidgetViewport` instead. Use the new `createTreeWidgetViewport` function to create `TreeWidgetViewport` from `Viewport`.",
```

packages/itwin/tree-widget/src/test/trees/CategoriesVisibilityUtils.test.ts
```diff
@@ -5,9 +5,8 @@
 
 import { expect } from "chai";
 import sinon from "sinon";
-import * as moq from "typemoq";
```
grigasp, line 8: so nice to see `typemoq` going away!

packages/itwin/tree-widget/src/test/trees/common/internal/AlwaysAndNeverDrawnElementInfo.test.ts
```diff
@@ -192,7 +206,14 @@ describe("AlwaysAndNeverDrawnElementInfo", () => {
       expect(result1).to.deep.eq(set);
       expect(vp.iModel.createQueryReader).to.be.calledOnce;
       info.suppressChangeEvents();
-      vp.setAlwaysDrawn(new Set(["0x4"]));
+      const setterFunction = (ids: Set<Id64String>) => {
+        if (setType === "always") {
+          vp.setAlwaysDrawn({ elementIds: ids });
+          return;
+        }
+        vp.setNeverDrawn(ids);
```
grigasp, line 214: why does `setAlwaysDrawn` take props and `setNeverDrawn` - not?

PR #1467, opened 2025-10-06
[Tree widget]: Fix incorrectly changing visibility when always drawn exclusive is true

JonasDov:
closes [#1654](https://github.com/iTwin/itwinjs-backlog/issues/1654)
Also, split `ModelsTreeVisibilityHandler.test.ts` into 3 files:
- `ModelsTreeVisibilityHandlerUnit.test.ts` - contains all unit tests.
- `ModelsTreeVisibilityHandlerFiltered.test.ts` - contains integration tests where models tree is filtered.
- `ModelsTreeVisibilityHandler.test.ts` - contains integration tests when tree isn't filtered.
Did this change because the file was getting too big: it would take multiple seconds to save the file, and maintaining it was getting difficult.

Also, ~added `retry(1)`~ _increased `waitFor` timeout to 5s_ to performance tests' visibility checking and added `defaultIfEmpty` to `ModelsTreeVisibilityHandler.getVisibilityStatus` (this is already in `next` branch). Sometimes visibility change events fire while retrieving nodes' visibility. This causes the `getVisibilityStatus` Observable to abort, and tests fail due to it. These additions will make performance tests more stable.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
@@ -3258,6 +3258,18 @@ describe("ModelsTreeVisibilityHandler", () => {
         viewport.changeModelDisplay(ids.model1, true);
         viewport.setAlwaysDrawn(new Set([ids.element2]), true);
         viewport.renderFrame();
+        await validateHierarchyVisibility({
+          provider,
+          handler,
+          viewport,
+          visibilityExpectations: {
+            subject: () => "partial",
+            model: (id) => (id === ids.model1 ? "partial" : "hidden"),
+            category: ({ categoryId }) => (categoryId === ids.category1 ? "partial" : "hidden"),
+            groupingNode: ({ elementIds }) => (elementIds.includes(ids.element2) ? "partial" : "hidden"),
+            element: ({ elementId }) => (elementId === ids.element2 ? "visible" : "hidden"),
+          },
+        });
         await handler.changeVisibility(createModelHierarchyNode(ids.model1), true);
         viewport.renderFrame();
 
@@ -3294,6 +3306,18 @@ describe("ModelsTreeVisibilityHandler", () => {
         viewport.changeModelDisplay(ids.model, true);
         viewport.setAlwaysDrawn(new Set([ids.element2]), true);
         viewport.renderFrame();
+        await validateHierarchyVisibility({
+          provider,
+          handler,
+          viewport,
+          visibilityExpectations: {
+            subject: () => "partial",
+            model: () => "partial",
+            category: ({ categoryId }) => (categoryId === ids.category1 ? "partial" : "hidden"),
+            groupingNode: ({ elementIds }) => (elementIds.includes(ids.element2) ? "partial" : "hidden"),
+            element: ({ elementId }) => (elementId === ids.element2 ? "visible" : "hidden"),
+          },
+        });
         await handler.changeVisibility(createCategoryHierarchyNode(ids.model, ids.category1, true), true);
         viewport.renderFrame();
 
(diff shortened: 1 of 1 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -693,6 +701,19 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
 
+  private removeAlwaysDrawnExclusive(): Observable<void> {
+    return from(this._idsCache.getAllCategories()).pipe(
+      map((categoryIds) => {
+        this._props.viewport.changeCategoryDisplay(categoryIds, false, false);
+        this._props.viewport.clearNeverDrawn();
+        this._props.viewport.perModelCategoryVisibility.clearOverrides();
+        if (this._props.viewport.alwaysDrawn) {
+          this._props.viewport.setAlwaysDrawn(this._props.viewport.alwaysDrawn);
+        }
```
grigasp, line 712: Can't we have a situation where the exclusive flag is set, but `alwaysDrawn` returns `undefined`? If that's possible, then we wouldn't be resetting the exclusive flag here.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
```diff
@@ -600,53 +602,59 @@ class ModelsTreeVisibilityHandlerImpl implements HierarchyVisibilityHandler {
...
-    }
-
-    return this.changeElementsState({
-      elementIds: new Set([...node.key.instanceKeys.map(({ id }) => id)]),
-      modelId,
-      categoryId,
-      on,
     });
+    if (this._props.viewport.isAlwaysDrawnExclusive) {
+      return this.removeAlwaysDrawnExclusive().pipe(mergeMap(() => changeObs));
```
grigasp, line 655: I feel like it's more appropriate to use `concat` here

change/@itwin-tree-widget-react-91d3d356-4797-4445-8a98-b2e6d73b2665.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Models tree: fixed visibility change incorrectly setting visibility when alwaysDrawn exclusive is set to true. Now, before making any visibility changes to the desired node, always drawn exclusive is set to false, all categories are turned off, never drawn list is cleared and all per-model category overrides are removed. ",
```
grigasp, line 3:
```suggestion
  "comment": "Models tree: Fixed visibility change incorrectly setting visibility when `alwaysDrawn` exclusive flag is set to `true`. Now, before making any visibility changes to the desired node, `alwaysDrawn` exclusive flag is reset to `false`, all categories are turned off, never drawn list is cleared and all per-model category overrides are removed. ",
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
JonasDov: Changed to `this._props.viewport.setAlwaysDrawn(this._props.viewport.alwaysDrawn ?? new Set());`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeVisibilityHandler.ts
JonasDov: Changed to: `return concat(this.removeAlwaysDrawnExclusive(), changeObs);`

packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
```diff
@@ -3235,6 +3237,159 @@ describe("ModelsTreeVisibilityHandler", () => {
...
+          return { model1, category1, element1, element2, otherModel, otherCategory };
+        });
+
+        const { imodel, ...ids } = buildIModelResult;
+        using visibilityTestData = createVisibilityTestData({ imodel });
+        const { handler, provider, viewport } = visibilityTestData;
+        viewport.changeModelDisplay(ids.model1, true);
+        viewport.setAlwaysDrawn(new Set([ids.element2]), true);
+        viewport.renderFrame();
+        await handler.changeVisibility(createModelHierarchyNode(ids.model1), true);
```
grigasp, line 3261: maybe it would be worth checking hierarchy visibility state  before doing the change? just to confirm it matches the state we expect.

grigasp (approved):

PR #1477, opened 2025-10-15
[Tree widget]: Fix element visibility ignoring children elements

JonasDov:
closes https://github.com/iTwin/viewer-components-react/issues/1129

### DIFF
--- a/apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
+++ b/apps/performance-tests/src/tree-widget/VisibilityUtilities.ts
@@ -4,7 +4,22 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { expect } from "chai";
-import { bufferCount, concatMap, defaultIfEmpty, delay, EMPTY, expand, firstValueFrom, from, identity, mergeMap, of, queueScheduler, takeLast, toArray } from "rxjs";
+import {
+  bufferCount,
+  concatMap,
+  defaultIfEmpty,
+  delay,
+  EMPTY,
+  expand,
+  firstValueFrom,
+  from,
+  identity,
+  mergeMap,
+  of,
+  queueScheduler,
+  takeLast,
+  toArray,
+} from "rxjs";
 import { assert } from "@itwin/core-bentley";
 import { Code, ColorDef, IModel, RenderMode } from "@itwin/core-common";
 import { IModelApp, OffScreenViewport, SpatialViewState, ViewRect } from "@itwin/core-frontend";


packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -37,9 +39,15 @@ interface ElementFilteredTreeNode extends BaseFilteredTreeNode {
 
 type FilteredTreeNode = GenericFilteredTreeNode | CategoryFilteredTreeNode | ElementFilteredTreeNode;
 
+interface GetElementsFromUnfilteredChildrenTreeProps {
+  childrenTree: ChildrenTree<any>;
```
grigasp: `unknown` should be preferred over `any` if we don't care about the type

```suggestion
  childrenTree: ChildrenTree<unknown>;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -105,6 +113,137 @@ export class ModelsTreeIdsCache implements Disposable {
     }
   }
 
+  private async *queryChildren({ elementIds }: { elementIds: Id64Array }): AsyncIterableIterator<{ id: Id64String; parentId: Id64String }> {
+    const ctes = [
+      `
+        ElementChildren(id, parentId) AS (
+          SELECT ECInstanceId id, Parent.Id parentId
+          FROM ${this.#hierarchyConfig.elementClassSpecification}
+          WHERE Parent.Id IN (${elementIds.join(", ")})
```
grigasp: what numbers of elements do we expect here? do we have any measures against a huge query that would result from a large number of ids?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -105,6 +113,137 @@ export class ModelsTreeIdsCache implements Disposable {
...
+      result.set(elementId, { children: elementChildrenTree });
+      entry.children.forEach((childId) => {
+        const childrenTreeOfChild = this.getChildrenTreeFromMap({ elementIds: childId });
+        // Need to add children tree created from childId. This tree includes childId as root element
+        if (childrenTreeOfChild.size > 0) {
+          elementChildrenTree.set(childId, { children: childrenTreeOfChild });
+          return;
+        }
+        // If child does not have children, children tree won't be created. Need to add childId with undefined children
+        elementChildrenTree.set(childId, { children: undefined });
```
grigasp:
```suggestion
        const childrenTreeOfChild = this.getChildrenTreeFromMap({ elementIds: childId });
        // Need to add children tree created from childId. This tree includes childId as root element
        // If child does not have children, children tree won't be created. Need to add childId with undefined children
        elementChildrenTree.set(childId, { children: childrenTreeOfChild.size > 0 ? childrenTreeOfChild : undefined });
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -194,20 +195,27 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
...
+      const filterTargets = new Map<Id64String, { childrenCount: number }>();
+      node.children.forEach((child) => {
+        if (child.extendedData?.childrenCount) {
+          childrenCount += child.extendedData.childrenCount;
+        }
         if (child.filtering) {
           if (child.filtering.hasFilterTargetAncestor) {
             hasFilterTargetAncestor = true;
-            break;
+            return;
```
grigasp: if we do this early return here, then `childrenCount` will only cover one child node, and `filterTargets` will be empty - is that intentional?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -107,9 +115,79 @@ export async function createFilteredTree(imodelAccess: ECClassHierarchyInspector
...
+  filteredNodes,
+}: {
+  tree: ChildrenTree<any>;
+  filteredNodes: Array<FilteredTreeNode | FilteredTreeRootNode>;
+}): Id64Set {
+  const getIdsRecursively = (childrenTree: ChildrenTree<any>, parentFilteredNodes: Array<FilteredTreeNode | FilteredTreeRootNode>): Id64Set => {
+    // If one of the parent filtered nodes does not have children, it is filter target and because of this, all elements in the childrenTree are in the filtered tree.
+    if (parentFilteredNodes.some((filteredNode) => !filteredNode.children)) {
+      return getIdsFromChildrenTree({ tree: childrenTree });
+    }
```
grigasp: A node may be a filter target AND have children at the same time. Is that situation covered?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/FilteredTree.ts
```diff
@@ -107,9 +115,79 @@ export async function createFilteredTree(imodelAccess: ECClassHierarchyInspector
...
+      result.add(id);
+      if (!entry.children || entry.children.size === 0) {
+        return;
+      }
+      // Continue recursively for children.
+      const resultFromChildren = getIdsRecursively(entry.children, nodes);
+      if (resultFromChildren.size === 0) {
+        return;
+      }
+      resultFromChildren.forEach((childId) => result.add(childId));
```
grigasp:
```suggestion
      if (entry.children) {
        // Continue recursively for children
        getIdsRecursively(entry.children, nodes).forEach((childId) => result.add(childId));
      }
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -105,6 +113,137 @@ export class ModelsTreeIdsCache implements Disposable {
...
+        // Add child to parent's entry and add child to children map
+        entry.children.push(id);
+        if (!childrenMap.has(id)) {
+          childrenMap.set(id, { children: undefined });
+        }
+      }
+
+      elementsToQuery.forEach((elementId) => childrenLoadingMap.delete(elementId));
+      return;
+    })({ childrenLoadingMap: this.#childrenLoadingMap, childrenMap: this.#childrenMap });
```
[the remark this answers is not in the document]
JonasDov: Yes, removed these args

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
grigasp: > function will be called with all grouped elements (should not be more than 1k unless hierarchy level limit is increased)

So it could easily be 10k, since we offer users to increase the loaded instances count to that by default. I think we need to batch these queries by ~1k. 10k ids here will produce a huge query.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -105,6 +113,137 @@ export class ModelsTreeIdsCache implements Disposable {
...
+      if (entry?.children) {
+        let totalChildrenCount = entry.children.length;
+        this.getChildrenCountMap({ elementIds: entry.children }).forEach((childrenOfChildCount) => (totalChildrenCount += childrenOfChildCount));
+        result.set(elementId, totalChildrenCount);
+      }
+    }
+    return result;
+  }
+
+  private createChildrenLoadingMapEntries({ elementsToQuery }: { elementsToQuery: Id64Array }): { promise: Promise<void> } {
```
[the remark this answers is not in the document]
JonasDov: Changed `promise` to `loadingMapEntry`. Added comment on this function that explains a little bit what this function does.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
JonasDov: Changed to add filterTargets to extended data only when grouping node does not have filter target ancestor. Also, regarding your last comment, `childrenCount` will always include all children count's because we are using `return` inside `forEach`.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/FilteredTree.ts
JonasDov: I just realized what you were asking. 
>do we have cases when nodes with isFilterTarget flag don't have children?

Did you mean to ask: do we have cases when nodes without children don't have isFilterTarget flag set?

In such case: no. All nodes without children should have isFilterTarget flag set. Changed code to this:
```ts
// If one of the parent filtered nodes does not have children, it is filter target and because of this, all elements in the childrenTree are in the filtered tree.
const hasParentFilterTarget = parentFilteredNodes.some((filteredNode) => "isFilterTarget" in filteredNode && filteredNode.isFilterTarget);
if (hasParentFilterTarget) {
  return getIdsFromChildrenTree({ tree: childrenTree });
}
```

change/@itwin-tree-widget-react-36f554d3-eaad-4bf1-997b-f8070462917e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Models tree: Fixed element visibility not taking into account child elements visibility. Also fixed changing element visibility not affecting child elements visibility.",
```
grigasp: please review

packages/itwin/tree-widget/src/test/trees/models-tree/internal/ModelsTreeVisibilityHandler.test.ts
```diff
@@ -168,7 +169,16 @@ describe("ModelsTreeVisibilityHandler", () => {
           imodelAccess: createFakeIModelAccess(),
         });
 
-        const node = createElementHierarchyNode({ modelId: "0x1", categoryId: "0x2", elementId: "0x3" });
+        const node = createElementHierarchyNode({
+          modelId: "0x1",
+          categoryId: "0x2",
+          elementId: "0x3",
+          parentKeys: [
+            { type: "instances", instanceKeys: [{ id: "0x1", className: MODEL_CLASS_NAME }] },
```
grigasp: it's very unlikely that the hierarchy will start with a Model node - could these `parentKeys` be adjusted to have at least one `bis.Subject` key in front?

change/@itwin-tree-widget-react-36f554d3-eaad-4bf1-997b-f8070462917e.json
JonasDov: Changed comment.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -462,10 +506,22 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
         }, new Array<Id64String>()),
       ),
     );
+    const childrenCountWhereClause = (parentAlias: string) => `
+      ${parentAlias}.Category.Id IN (${categoryIds.map(() => "?").join(",")})
+      AND ${parentAlias}.Model.Id IN (${modelIds.map(() => "?").join(",")})
+      AND ${parentAlias}.Parent.Id IS NULL
+    `;
```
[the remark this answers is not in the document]
JonasDov: Updated query to not use the same where clause both in cte and ecsql. Now we use where clause only in cte and select elements based on ids retrieved from the cte.

PR #1488, opened 2025-10-23
Prettier: change import sorter plugin

JonasDov:
At the moment we are using https://github.com/[SoominHan/import-sorter](https://github.com/SoominHan/import-sorter) plugin to sort imports. It has a few flaws: 
1. It can not sort `type` imports (or at least it's not trivial).
2. It sometimes bugs out, and removes imports that it deems as `unused` even though they are used.

Changed to use https://github.com/IanVS/prettier-plugin-sort-imports

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -14,6 +14,6 @@
     }
   ],
   "importOrderSafeSideEffects": [".s?css$"],
-  "importOrderTypeScriptVersion": "5.0.0",
+  "importOrderTypeScriptVersion": "5.6.3",
   "importOrderCaseSensitive": false
 }


.prettierrc
```diff
@@ -2,5 +2,18 @@
...
+  "plugins": ["@ianvs/prettier-plugin-sort-imports"],
+  "importOrder": ["", ".s?css$", "", "^[A-Za-z]", "^@", "^[.]", "", "<TYPES>", "<TYPES>^@", "<TYPES>^[.]", "", "<BUILTIN_MODULES>"],
+  "overrides": [
+    {
+      "files": ["**/learning-snippets/**/*.ts*", "**/test/**/setup.ts", "**/README.md"],
+      "options": {
+        "importOrder": []
+      }
+    }
+  ],
```
grigasp: do I get this correctly - doing auto-format won't ruin all the learning snippet imports anymore?

.prettierrc
```diff
@@ -2,5 +2,18 @@
...
+  "importOrderSafeSideEffects": [".s?css$"],
+  "importOrderTypeScriptVersion": "5.0.0",
```
grigasp: why `5.0.0` and not the one we're actually using?

.prettierrc
JonasDov: Yes, you can save files and run prettier - it won't mess up the imports

PR #1508, opened 2025-11-10
[Measurement-tools] World values in non-civil sheets

Maxime-Brassard:
Refactor of the measure-tools in order to introduce a new way to calculate "world values" of measurements in all models (even non-civil models).

In order to do this, we need to use the model's relations in order to link up our entry point (Biscore.Sheet) to Biscore.Drawing and, finally, Biscore.SpatialViewDefinition.
image
The critical link for this new method is BisCore.SectionDrawingGeneratedFromSpatialView which links up the Biscore.DrawingViewDefinition to the Biscore.SpatialViewDefinition. Currently, this relation is NOT present on every model but we'd eventually like it to be. Therefore, the old method for civil models is still present as a backup and will be used when this relation is not present.
 
A lot of functions were marked as deprecated but this PR should not include any breaking change to the api.


packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -59,26 +60,46 @@ export namespace DrawingMetadata {
...
-    const masterOrigin = obj.sheetToWorldTransform?.masterOrigin.toJSONXYZ();
-    const sheetTov8Drawing = obj.sheetToWorldTransform?.sheetTov8Drawing.toJSON();
-    const v8DrawingToDesign = obj.sheetToWorldTransform?.v8DrawingToDesign.toJSON();
-    if (origin !== undefined)
-      return { origin, extents, worldScale: obj.worldScale, drawingId: obj.drawingId, sheetToWorldTransform: (masterOrigin !== undefined && sheetTov8Drawing !== undefined && v8DrawingToDesign !== undefined) ? {masterOrigin, sheetTov8Drawing, v8DrawingToDesign}: undefined };
+    if (origin !== undefined) {
+      return {
+        origin, extents, drawingId: obj.drawingId,
+        sheetToWorldTransformProps: {
+          DVDOrigin: obj.sheetToWorldTransformProps?.DVDOrigin,
```
a-gagnon: There's probably a way to spread the values instead of doing direct assignment, something like 
`sheetToWorldTransformProps: obj.sheetToWorldTransformProps ? { ...obj.sheetToWorldTransformProps } : undefined`.

Same with the code in `fromJSON`

PR #1520, opened 2025-11-20
[Measurement-tools] Fixed new sheet to world transformation

Maxime-Brassard:
A few fixes for the new transform
1. Cloning the transforms when using them to create new transforms in order to prevent modifying them
2. Using the jsonProp scale when using the civil transforms as these older models might not always have the sheetScale present on the biscore.sheet
3. Corrected adjustPointWithSheetToWoldTransform to not use the inverse


change/@itwin-measure-tools-react-02ad5415-c4c9-4826-9b0b-7fd347387ac1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "a few fixes",
```
aruniverse: ???

PR #1533, opened 2025-12-09
Fix flaky context menu item test by correcting waitFor() usage and adding PropertyGrid load wait

MattHandzel:
The "renders context menu item" test was flaky due to two root causes:

1. Incorrect waitFor() usage: The test had an async callback that executed user.pointer() and immediately checked the assertion without waiting for the DOM to update. The waitFor() callback should be synchronous.

2. Missing PropertyGrid load wait: The test attempted to interact with the PropertyGrid before it finished loading, causing the "Test SpatialCategory" element to not be found.

Changes:

- Enable the test
- Added this.timeout(10000) to increase test timeout to 10 seconds
- Added explicit waitFor() with 5-second timeout to wait for PropertyGrid to load and display the category before interaction
- Separated user interaction (pointer event) from assertion
- Made the assertion callback synchronous to properly wait for context menu

Verification:

- Ran this test 100 times locally without failures
- No regressions in other tests

### DIFF
--- a/apps/learning-snippets/src/test/property-grid/ContextMenuItem.test.tsx
+++ b/apps/learning-snippets/src/test/property-grid/ContextMenuItem.test.tsx
@@ -17,7 +17,7 @@ import type { ContextMenuItemProps } from "@itwin/property-grid-react";
 // __PUBLISH_EXTRACT_START__ PropertyGrid.PropertyGridWithContextMenuItemImports
 import { PropertyGridComponent } from "@itwin/property-grid-react";
 // __PUBLISH_EXTRACT_END__
-import { cleanup, queryByText, render, waitFor } from "@testing-library/react";
+import { cleanup, queryByText, render, waitFor, screen } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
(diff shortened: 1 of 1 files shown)


grigasp (approved):
Thanks for the PR, looks good.

ben-polinsky (approved):

MattHandzel:
> @MattHandzel thanks - could you run `prettier --write` to correct formatting issues?

Hello @ben-polinsky, I ran prettier, not sure if you get notified when I make a commit or if I should write you with a comment like this, but it should be ready to be merged :)

grigasp:
> > @MattHandzel thanks - could you run `prettier --write` to correct formatting issues?
> 
> Hello @ben-polinsky, I ran prettier, not sure if you get notified when I make a commit or if I should write you with a comment like this, but it should be ready to be merged :)

Thanks. I didn't get notified until you commented. Re-ran the pipelines and prettier is still complaining:
```
> viewer-components-react@ prettier /home/runner/work/viewer-components-react/viewer-components-react
> prettier --check .

Checking formatting...
[warn] apps/learning-snippets/src/test/property-grid/ContextMenuItem.test.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

PR #1540, opened 2026-01-07
[tree-widget]: React to breaking Tree/TreeRenderer API changes

saskliutas:
Updated to latest `@itwin/presentation-hierarchies` and `@itwin/presentation-hierarchies-react`. 
Reacted to breaking changes in `@itwin/presentation-hierarchies-react` and exposed ability to pass down props to the underlying `Tree.Root` and `Tree.Item` components.

### DIFF
--- a/packages/itwin/tree-widget/api/tree-widget-react.api.md
+++ b/packages/itwin/tree-widget/api/tree-widget-react.api.md
@@ -605,9 +605,9 @@ interface UseCategoriesTreeProps {
 // @beta (undocumented)
 interface UseCategoriesTreeResult {
     // (undocumented)
-    categoriesTreeProps: Pick<VisibilityTreeProps, "treeName" | "getHierarchyDefinition" | "getSearchPaths" | "visibilityHandlerFactory" | "highlightText" | "emptyTreeContent">;
-    // (undocumented)
     getTreeItemProps: Required<ExtendedVisibilityTreeRendererProps>["getTreeItemProps"];
+    // (undocumented)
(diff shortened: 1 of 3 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTree.tsx
```diff
@@ -32,16 +35,16 @@ export function ModelsTree({
...
   getInlineActions,
   getMenuActions,
   getContextMenuActions,
   emptyTreeContent,
   getSubTreePaths,
+  getTreeItemProps,
   treeLabel,
 }: ModelsTreeProps) {
-  const { modelsTreeProps, rendererProps } = useModelsTree({
+  const { modelsTreeProps, getTreeItemProps: modelsTreeItemProps } = useModelsTree({
```
grigasp: It's a bit strange that we return `{ modelsTreeProps, getTreeItemProps }` - `modelsTreeProps` name is specific to the component, but `getTreeItemProps` is generic. Maybe rename `getTreeItemProps` to `getModelsTreeItemProps`?

grigasp: Or maybe make it return just `{ treeProps, getTreeItemProps }`, and then consumers could do this:

```ts
const modelsTree = useModelsTree({ ... });
return (
  <VisibilityTree
    {...modelsTree.treeProps}
    treeRenderer={(treeProps) => (
      <VisibilityTreeRenderer
        {...treeProps}
        getTreeItemProps={(node) => modelsTree.getTreeItemProps(node, treeProps)}
      />
    )}
  />
);
```

PR #1543, opened 2026-01-12
[Tree widget]: Fix query error causing application crash

JonasDov:
We are using restart tokens to cancel old requests. In cases where old requests are cancelled due to restart token, `BE_SQLITE_INTERRUPT` error is thrown and it could cause application crash. Addressed the issue by:
1. Added safeguards for `BE_SQLITE_INTERRUPT` errors (in such cases the error is no longer thrown).
2. For other types of errors added `useErrorState` -  A hook that helps components throw errors in React's render loop so they can be captured by React error boundaries. So instead of causing the application to crash, only tree widget would show: 
image

### DIFF
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/Utils.ts
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/Utils.ts
@@ -42,9 +42,6 @@ export async function showAll(props: {
   componentId?: GuidString;
 }) {
   const { models, categories, viewport, componentId } = props;
-  viewport.changeModelDisplay({ modelIds: models, display: true });
-  viewport.clearNeverDrawn();
-  viewport.clearAlwaysDrawn();
   if (categories) {
     await showAllCategories(categories, viewport);
   } else {
@@ -55,6 +52,9 @@ export async function showAll(props: {
     const ids = categoryInfos.map((categoryInfo) => categoryInfo.categoryId);
     await enableCategoryDisplay(viewport, ids, true);
   }
+  viewport.changeModelDisplay({ modelIds: models, display: true });
+  viewport.clearNeverDrawn();
+  viewport.clearAlwaysDrawn();
 }
 
 /**
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeButtons.tsx
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeButtons.tsx
@@ -19,7 +19,6 @@ import {
   CLASS_NAME_GeometricModel3d,
   CLASS_NAME_InformationPartitionElement,
 } from "../common/internal/ClassNameDefinitions.js";
-import { useErrorState } from "../common/internal/UseErrorState.js";
 import { useGuid } from "../common/internal/useGuid.js";
 import { areAllModelsVisible, hideAllModels, invertAllModels, showAll, toggleModels } from "../common/Utils.js";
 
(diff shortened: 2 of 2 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/external-sources-tree/ExternalSourcesTreeDefinition.ts
```diff
@@ -302,18 +303,25 @@ export class ExternalSourcesTreeDefinition implements HierarchyDefinition {
...
+        { restartToken: `${this.#componentName}/${this.#componentId}/is-external-source-supported` },
+      )) {
+        return true;
+      }
+      return false;
+    } catch (error) {
+      if (!isBeSqliteInterruptError(error)) {
+        throw error;
+      }
+      return false;
```
grigasp: I'm not sure it's okay to say the component doesn't support given iModel if this query fails. Instead, I think we should say there was a failure and allow to retry.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -89,6 +90,12 @@ export class ModelsTreeIdsCache implements Disposable {
         { rowFormat: "ECSqlPropertyNames", limit: "unbounded", restartToken: `${this.#componentName}/${this.#componentId}/visible-sub-categories` },
       );
     }).pipe(
+      catchError((error) => {
+        if (isBeSqliteInterruptError(error)) {
+          return EMPTY;
+        }
+        throw error;
+      }),
```
grigasp: This is repeated a lot - maybe create a shared operator like `catchBeSQLiteInterrupts` that would internally call `catchError` and handle interrupts?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/UseHierarchyVisibility.ts
```diff
@@ -124,7 +115,13 @@ export function useHierarchyVisibility({ visibilityHandlerFactory }: UseHierarch
...
       // visible should become hidden, partial and hidden should become visible TODO: redo for clarity
       const on = visibilityState === "visible" ? false : true;
-      void handler.changeVisibility(node.nodeData, on);
+      void (async () => {
+        try {
+          await handler.changeVisibility(node.nodeData, on);
+        } catch (error) {
+          setErrorState(error);
+        }
+      })();
```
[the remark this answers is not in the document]
JonasDov: Changed, to only catch the error here and not render "error" state when that happens.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
JonasDov: Added `catchBeSQLiteInterrupts` and removed duplication

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/UseHierarchyVisibility.ts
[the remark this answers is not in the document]
grigasp: If getting the categories is the single point of failure, then maybe we should first get the categories, and only then do the other visibility changes? That way, if getting categories fails - we won't make any changes. If it succeeds, then we know we can fully make the requested change.

PR #1556, opened 2026-01-21
[Tree widget]: Adjust categories tree visibility handling

JonasDov:
Fixed a bug in categories tree, where turning on category visibility of hidden model would also turn on other categories in that model.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesVisibilityHandler.test.ts
@@ -366,17 +366,19 @@ describe("CategoriesVisibilityHandler", () => {
     });
 
     describe("categories", () => {
-      it("showing category of hidden model does not turn on categories without overrides in that model", async function () {
+      it("showing category of hidden model does not enable other categories", async function () {
         await using buildIModelResult = await buildIModel(this, async (builder) => {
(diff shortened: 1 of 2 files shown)


change/@itwin-tree-widget-react-8ba9612f-dcb9-4a41-a749-fe3bde8da80e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "[Tree widget]: Adjust visibility handling in categories tree. Now, showing categories of hidden models does not turn on other categories in that model.",
```
saskliutas: You don't need `[Tree widget]` in change log file that is specific to tree-widget-react

packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesVisibilityHandler.test.ts
```diff
@@ -366,6 +366,44 @@ describe("CategoriesVisibilityHandler", () => {
...
+        viewport.renderFrame();
+        await handler.changeVisibility(createCategoryHierarchyNode(keys.category.id), true);
+        await validateHierarchyVisibility({
+          provider,
+          handler,
+          viewport,
+          expectations: {
+            [keys.category.id]: "visible",
+            [keys.otherCategory.id]: "hidden",
+            [keys.categoryWithOverride.id]: "visible",
```
grigasp: hm, so we're still changing the effective status of `categoryWithOverride`, because we don't change the override. why?

PR #1584, opened 2026-02-05
[Tree widget]: Add element model categories cache

JonasDov:
`ModelsTreeIdsCache`, `CategoriesTreeIdsCache` and `ClassificationsTreeIdsCache` used almost the same implementation to query and store data about element model categories. Moved them to `ElementModelCategoriesCache`, which is now used by them.

### DIFF
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
@@ -1,0 +1,160 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { defer, forkJoin, from, map, mergeMap, reduce, shareReplay } from "rxjs";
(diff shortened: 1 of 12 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
```diff
@@ -0,0 +1,160 @@
...
+      reduce((acc, [modelId, { categoriesOfTopMostElements, isSubModel }]) => {
+        if (!isSubModel && categoriesOfTopMostElements.has(categoryId)) {
+          acc.push(modelId);
+        }
+        return acc;
+      }, new Array<ModelId>()),
+    );
+  }
+
+  public getAllCategoriesOfElements(): Observable<Id64Set> {
```
saskliutas: Why was this renamed to `getAllCategoriesOfElements` from `getAllCategories`? `OfElements` part brings some confusion (at least for me) because there is no arguments that specify of which elements.

JonasDov: This function is only used when always drawn exclusive mode is enabled and visibility handler wants to remove it.
Previous name `getAllCategories` was a little bit misleading: to remove always drawn exclusive mode,  visibility handler needs to disable only categories that are related to geometric elements. That's why the rename to  `getAllCategoriesOfElements`. Maybe you have some suggestion on how this function should be named?

PR #1595, opened 2026-02-12
[Tree widget]: Add performance test for child elements with different categories in `next`

JonasDov:
Added a performance test that does similar hierarchy to `changing element visibility changes only parent nodes visibility with 50k elements` except where child elements have different categories from their parent. This test should help to see how https://github.com/iTwin/viewer-components-react/issues/1100 would affect performance


grigasp:
need to rerun the benchmark when pipelines on 836e720f7c498131f95ba4f1e1bed24aca5ac83a complete

Issue #1512, opened 2025-11-18
Tree widget: Fix main thread blockage on next branch


JonasDov:
Performance tests on next branch show that main thread is blocked for a long time in some cases. This needs to be fixed after https://github.com/iTwin/viewer-components-react/pull/1511 is merged.
```md
Status | Test                                                                                              | Duration    | Blocks                                                 
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PASS   | categories tree creates initial filtered view for 50k items                                       | 2680.35 ms  | count: 6, max: 1831.00, p95: 1831.00, median: 31.00    
PASS   | categories tree changing definition container visibility changes visibility for 50k subCategories | 8536.17 ms  | count: 2, max: 6711.00, p95: 6711.00, median: 3483.00  
PASS   | categories tree changing definition container visibility changes visibility for 50k categories    | 12064.57 ms | count: 41, max: 3127.00, p95: 69.00, median: 47.00     
PASS   | classifications tree loads initial view for iModel with 50k classifications                       | 70.92 ms    | count: 0                                               
PASS   | classifications tree loads first branch for iModel with 50k classifications                       | 1935.96 ms  | count: 6, max: 126.00, p95: 126.00, median: 34.50      
PASS   | models tree creates initial filtered view for 50k target items                                    | 628.51 ms   | count: 5, max: 76.00, p95: 76.00, median: 29.00        
PASS   | models tree validates categories visibility for imodel with 50k categories                        | 26182.41 ms | count: 10, max: 11739.00, p95: 11739.00, median: 199.00
PASS   | models tree changing model visibility changes visibility for 50k elements                         | 2032.02 ms  | count: 0                                               
PASS   | models tree changing category visibility changes visibility for 50k elements                      | 2100.9 ms   | count: 0                                               
PASS   | models tree changing per-model-category override changes visibility for 50k elements              | 2419.74 ms  | count: 0
```

PR #1600, opened 2026-02-19
[Tree widget]: Adjust when always/never drawn elements are requested

JonasDov:
Part of https://github.com/iTwin/viewer-components-react/issues/1512.

Updated visibility computation to request always/never drawn elements only from the set opposite to the current default status (e.g., if default is visible, query only never-drawn; if hidden, query only always-drawn), instead of querying both sets. This reduces unnecessary work and lowers compute cost, especially when both sets are populated.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/TreeUtils.ts
+++ b/packages/itwin/tree-widget/src/test/trees/TreeUtils.ts
@@ -109,12 +109,12 @@ export function createTreeWidgetTestingViewport({
       onDisplayedModelsChanged.shouldFireOnRender = true;
     },
     changeSubCategoryDisplay: (props) => {
-      subCategories.forEach((subCategoryMap) => {
+      for (const subCategoryMap of subCategories.values()) {
         if (subCategoryMap.has(props.subCategoryId)) {
           subCategoryMap.set(props.subCategoryId, { isVisible: props.display });
           return;
         }
-      });
+      }
     },
     clearNeverDrawn: () => {
       neverDrawn = new Set();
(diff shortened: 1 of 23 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/visibility/BaseVisibilityHelper.ts
```diff
@@ -386,72 +386,53 @@ export class BaseVisibilityHelper implements Disposable {
...
-        return of(createVisibilityStatus("hidden"));
-      }
-    } else if (!neverDrawn?.size && !alwaysDrawn?.size) {
-      return of(defaultStatus());
+    const defaultStatus = this.#props.viewport.isAlwaysDrawnExclusive ? createVisibilityStatus("hidden") : props.defaultStatus;
+    const oppositeSetFromDefaultStatus = defaultStatus.state === "visible" ? this.#props.viewport.neverDrawn : this.#props.viewport.alwaysDrawn;
+    if (!oppositeSetFromDefaultStatus?.size) {
+      return of(defaultStatus);
     }
+    const oppositeSetType = defaultStatus.state === "visible" ? "never" : "always";
```
saskliutas: Looks like this can be combined with `oppositeSetFromDefaultStatus` check e.g.
```ts
const { oppositeSet, setType } = defaultStatus.state === "visibile" 
  ? { oppositeSet: this.#props.viewport.neverDrawn, setType: "never" } 
  : { oppositeSet: this.#props.viewport.alwaysDrawn, setType: "always" } 
```

PR #1619, opened 2026-03-05
[Tree widget]: Allow searching for categories under sub-models

JonasDov:
Fixed a bug in Models and Categories trees:
When hierarchy contained sub-models, and user would search for a category that was under sub-model, it would not be found.
E.g. hierarchy:
- Model
  - Category
    - Element (is modeled)
      - Category2
         - element (model is Element)  
         
Searching for `category2` would not find `Category2` node.

### DIFF
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
@@ -676,6 +676,15 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
   }
 }
 
+const SUBJECT_TYPE_AS_NUMBER = 1;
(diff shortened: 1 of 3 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -949,8 +959,19 @@ function createInstanceKeyPathsFromInstanceLabelObs(
...
     }),
     catchBeSQLiteInterrupts,
-    map((row) => ({ className: row[0], id: row[1] })),
-    toArray(),
-    mergeMap((targetKeys) => createInstanceKeyPathsFromTargetItemsObs({ ...props, targetItems: targetKeys })),
+    map((row) => {
+      const key = row[1];
+      switch (row[0]) {
+        case "s":
+          return { key, type: 1 };
```
grigasp: let's create consts for these `type` values

PR #1650, opened 2026-04-14
[Tree widget]: Inconsistent visibility state fix

### LINKED ISSUE #1649: Inconsistent state on visibility in the new tree widget
**Description of the issue and steps to reproduce**

1. Open an iModel containing several elements and categories.
2. In the tree widget, click fast several times on the eye icon of a category to toggle visibility on and off 
3. At some point the toggle is no more reflecting the state of the element in the category (e.g. the eye is open while the element is hidden in the 3D view)
4. Try to click again on the eye icon -> no effect, neither in the 3D view, nor in the state of the eye icon

(see video in attachment)

**Package**
This happens in "@itwin/tree-widget-react": "4.0.0-alpha.27" but had already been seen in previous alpha versions

**Screenshots**

**Desktop (please complete the applicable information):**

- OS: macOS (but it has been reported in an Electron app running on Windows too)
- Browser: Chromium engine
- Version: 147.0.7727.56
- iTwin.js Version 5.8.0

https://github.com/user-attachments/assets/cebea475-9725-42df-8119-b5fb09a474b3

JonasDov:
closes https://github.com/iTwin/viewer-components-react/issues/1649

The main changes:
1. Added `needsRefresh = true` when visibility button is clicked.
2. When clicking visibility icon, the cached visibility immediatly changes to either `visibile` or `hidden`  depending on what state it was in. (Previously only a tooltip would be changed to `Determining visibility...` and state would be adjusted only when visibility status is retrieved from visibility handler).
3. Added `BufferingViewport` - the main purpose of this viewport is to collect visibility changes instead of making them directly on the actual viewport. Then later, all the changes can be applied at once by calling `commit`, or discarded by calling `discard`.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
@@ -1,0 +1,298 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { describe, expect, it } from "vitest";
+import { BeEvent, Id64 } from "@itwin/core-bentley";
+import { type IModelConnection } from "@itwin/core-frontend";
+import { BufferingViewport } from "../../../../tree-widget-react/components/trees/common/internal/BufferingViewport.js";
+
+import type { Id64String } from "@itwin/core-bentley";
+import type { PerModelCategoryOverride, TreeWidgetViewport } from "../../../../tree-widget-react/components/trees/common/TreeWidgetViewport.js";
+
+function createMockViewport(props?: {
+  models?: Map<Id64String, boolean>;
+  categories?: Map<Id64String, boolean>;
+  subCategories?: Map<Id64String, boolean>;
+  alwaysDrawn?: Set<Id64String>;
+  neverDrawn?: Set<Id64String>;
+  isAlwaysDrawnExclusive?: boolean;
+  perModelCategoryOverrides?: Map<Id64String, Map<Id64String, PerModelCategoryOverride>>;
+}): TreeWidgetViewport {
+  const models = props?.models ?? new Map<Id64String, boolean>();
+  const categories = props?.categories ?? new Map<Id64String, boolean>();
+  const subCategories = props?.subCategories ?? new Map<Id64String, boolean>();
+  let alwaysDrawn: Set<Id64String> | undefined = props?.alwaysDrawn;
+  let neverDrawn: Set<Id64String> | undefined = props?.neverDrawn;
+  let isExclusive = props?.isAlwaysDrawnExclusive ?? false;
+  const overrides = props?.perModelCategoryOverrides ?? new Map<Id64String, Map<Id64String, PerModelCategoryOverride>>();
+
+  return {
+    viewType: "3d",
+    iModel: {} as IModelConnection,
+    viewsModel: (modelId) => models.get(modelId) ?? false,
+    changeModelDisplay: ({ modelIds, display }) => {
+      for (const id of Id64.iterable(modelIds)) {
+        models.set(id, display);
+      }
+    },
+    viewsCategory: (categoryId) => categories.get(categoryId) ?? false,
+    changeCategoryDisplay: ({ categoryIds, display }) => {
+      for (const id of Id64.iterable(categoryIds)) {
+        categories.set(id, display);
+      }
+    },
+    viewsSubCategory: (subCategoryId) => subCategories.get(subCategoryId) ?? false,
+    changeSubCategoryDisplay: ({ subCategoryId, display }) => {
+      subCategories.set(subCategoryId, display);
+    },
+    get alwaysDrawn() {
+      return alwaysDrawn;
+    },
+    setAlwaysDrawn: ({ elementIds, exclusive }) => {
+      alwaysDrawn = elementIds;
+      isExclusive = !!exclusive;
+    },
+    clearAlwaysDrawn: () => {
+      alwaysDrawn = undefined;
+      isExclusive = false;
+    },
+    get neverDrawn() {
+      return neverDrawn;
+    },
+    setNeverDrawn: ({ elementIds }) => {
+      neverDrawn = elementIds;
+    },
+    clearNeverDrawn: () => {
+      neverDrawn = undefined;
+    },
+    get isAlwaysDrawnExclusive() {
+      return isExclusive;
+    },
+    get perModelCategoryOverrides(): Iterable<{ modelId: Id64String; categoryId: Id64String; visible: boolean }> {
+      const result: Array<{ modelId: Id64String; categoryId: Id64String; visible: boolean }> = [];
+      for (const [modelId, categoryMap] of overrides) {
+        for (const [categoryId, override] of categoryMap) {
+          if (override !== "none") {
+            result.push({ modelId, categoryId, visible: override === "show" });
+          }
+        }
+      }
+      return result;
+    },
+    setPerModelCategoryOverride: ({ modelIds, categoryIds, override }) => {
+      for (const modelId of Id64.iterable(modelIds)) {
+        for (const categoryId of Id64.iterable(categoryIds)) {
+          if (!overrides.has(modelId)) {
+            overrides.set(modelId, new Map<Id64String, PerModelCategoryOverride>());
+          }
+          overrides.get(modelId)!.set(categoryId, override);
+        }
+      }
+    },
+    getPerModelCategoryOverride: ({ modelId, categoryId }) => {
+      return overrides.get(modelId)?.get(categoryId) ?? "none";
+    },
+    clearPerModelCategoryOverrides: (clearProps) => {
+      if (clearProps?.modelIds !== undefined) {
+        for (const modelId of Id64.iterable(clearProps.modelIds)) {
+          overrides.delete(modelId);
+        }
+      } else {
+        overrides.clear();
+      }
+    },
+    onAlwaysDrawnChanged: new BeEvent<() => void>(),
+    onNeverDrawnChanged: new BeEvent<() => void>(),
+    onDisplayStyleChanged: new BeEvent<() => void>(),
+    onDisplayedModelsChanged: new BeEvent<() => void>(),
+    onDisplayedCategoriesChanged: new BeEvent<() => void>(),
+    onPerModelCategoriesOverridesChanged: new BeEvent<() => void>(),
+  };
+}
+
+describe("BufferingViewport", () => {
+  describe("models", () => {
+    [
+      {
+        describeName: "models",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.viewsModel(id ?? "0x1"),
+        changeDisplay: (vp: BufferingViewport) => vp.changeModelDisplay({ modelIds: "0x1", display: false }),
+        getViewport: () => createMockViewport({ models: new Map([["0x1", true]]) }),
+      },
+      {
+        describeName: "categories",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.viewsCategory(id ?? "0x1"),
+        changeDisplay: (vp: BufferingViewport) => vp.changeCategoryDisplay({ categoryIds: "0x1", display: false }),
+        getViewport: () => createMockViewport({ categories: new Map([["0x1", true]]) }),
+      },
+      {
+        describeName: "sub-categories",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.viewsSubCategory(id ?? "0x1"),
+        changeDisplay: (vp: BufferingViewport) => vp.changeSubCategoryDisplay({ subCategoryId: "0x1", display: false }),
+        getViewport: () => createMockViewport({ subCategories: new Map([["0x1", true]]) }),
+      },
+      {
+        describeName: "always drawn",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.alwaysDrawn?.has(id ?? "0x1") ?? false,
+        changeDisplay: (vp: BufferingViewport) => vp.setAlwaysDrawn({ elementIds: new Set(["0x2"]) }),
+        getViewport: () => createMockViewport({ alwaysDrawn: new Set(["0x1"]) }),
+      },
+      {
+        describeName: "always drawn exclusive",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.alwaysDrawn?.has(id ?? "0x1") && vp.isAlwaysDrawnExclusive,
+        changeDisplay: (vp: BufferingViewport) => vp.setAlwaysDrawn({ elementIds: new Set(["0x1"]), exclusive: false }),
+        getViewport: () => createMockViewport({ alwaysDrawn: new Set(["0x1"]), isAlwaysDrawnExclusive: true }),
+      },
+      {
+        describeName: "never drawn",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.neverDrawn?.has(id ?? "0x1") ?? false,
+        changeDisplay: (vp: BufferingViewport) => vp.setNeverDrawn({ elementIds: new Set(["0x2"]) }),
+        getViewport: () => createMockViewport({ neverDrawn: new Set(["0x1"]) }),
+      },
+    ].forEach(({ describeName, getDisplay, changeDisplay, getViewport }) => {
+      describe(describeName, () => {
+        it("returns viewport viewports result when change function has not been called", () => {
+          const viewport = getViewport();
+          const bufferingViewport = new BufferingViewport(viewport);
+          expect(getDisplay(bufferingViewport)).toBe(true);
+          expect(getDisplay(viewport)).toBe(true);
+          expect(getDisplay(bufferingViewport, "0x2")).toBe(false);
+          expect(getDisplay(viewport, "0x2")).toBe(false);
+        });
+
+        it("returns adjusted values when change function has been called", () => {
+          const viewport = getViewport();
+          const bufferingViewport = new BufferingViewport(viewport);
+          changeDisplay(bufferingViewport);
+          expect(getDisplay(viewport)).toBe(true);
+          expect(getDisplay(bufferingViewport)).toBe(false);
+        });
+
+        it("adjusts viewport viewport when commit is called", () => {
+          const viewport = getViewport();
+          const bufferingViewport = new BufferingViewport(viewport);
+          changeDisplay(bufferingViewport);
+          bufferingViewport.commit();
+          expect(getDisplay(viewport)).toBe(false);
+          expect(getDisplay(bufferingViewport)).toBe(false);
+        });
+
+        it("discards changes when discard is called", () => {
+          const viewport = getViewport();
+          const bufferingViewport = new BufferingViewport(viewport);
+          changeDisplay(bufferingViewport);
+          bufferingViewport.discard();
+          expect(getDisplay(viewport)).toBe(true);
+          expect(getDisplay(bufferingViewport)).toBe(true);
+        });
+      });
+    });
+  });
+
+  describe("per-model category overrides", () => {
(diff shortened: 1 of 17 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/AlwaysAndNeverDrawnElementInfoCache.ts
```diff
@@ -337,22 +338,27 @@ export class AlwaysAndNeverDrawnElementInfoCache implements Disposable {
...
-        if (viewport.alwaysDrawn?.size && alwaysDrawn.size) {
-          viewport.setAlwaysDrawn({ elementIds: setDifference(viewport.alwaysDrawn, alwaysDrawn) });
+        if (this.#viewport.alwaysDrawn?.size && alwaysDrawn.size) {
+          bufferingViewport.setAlwaysDrawn({ elementIds: setDifference(this.#viewport.alwaysDrawn, alwaysDrawn) });
         }
-        if (viewport.neverDrawn?.size && neverDrawn.size) {
-          viewport.setNeverDrawn({ elementIds: setDifference(viewport.neverDrawn, neverDrawn) });
+        if (this.#viewport.neverDrawn?.size && neverDrawn.size) {
+          bufferingViewport.setNeverDrawn({ elementIds: setDifference(this.#viewport.neverDrawn, neverDrawn) });
         }
```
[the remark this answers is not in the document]
JonasDov: Changed to call `setDifference` with the buffering viewport

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/BufferingViewport.ts
```diff
@@ -0,0 +1,271 @@
...
+    if (this.#clearedPerModelCategoryOverrides !== "all") {
+      return this.#clearedPerModelCategoryOverrides;
+    }
+    this.#clearedPerModelCategoryOverrides = new Set<string>();
+    // When converting from "all" to Set, need to iterate over real viewport overrides and add them to cleared set
+    // If they have not been changed yet.
+    for (const { modelId } of this.#realViewport.perModelCategoryOverrides) {
+      if (!this.#changedPerModelCategoryOverrides.has(modelId)) {
+        this.#clearedPerModelCategoryOverrides.add(modelId);
+      }
```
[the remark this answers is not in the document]
JonasDov: Made a different change, instead of checking if changed overrides contain that model, added a check which checks that it has model AND category, this way the correct models are added.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/BufferingViewport.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  public get onPerModelCategoriesOverridesChanged(): BeEvent<() => void> {
+    return this.#realViewport.onPerModelCategoriesOverridesChanged;
+  }
+
+  public commit(): void {
+    for (const callback of this.#onCommitCallbacks) {
+      callback();
+    }
+    this.discard();
```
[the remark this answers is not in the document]
JonasDov: changed

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      },
+      {
+        describeName: "never drawn",
+        getDisplay: (vp: TreeWidgetViewport, id?: string) => vp.neverDrawn?.has(id ?? "0x1") ?? false,
+        changeDisplay: (vp: BufferingViewport) => vp.setNeverDrawn({ elementIds: new Set(["0x2"]) }),
+        getViewport: () => createMockViewport({ neverDrawn: new Set(["0x1"]) }),
+      },
+    ].forEach(({ describeName, getDisplay, changeDisplay, getViewport }) => {
+      describe(describeName, () => {
+        it("returns viewport viewports result when change function has not been called", () => {
```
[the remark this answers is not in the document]
JonasDov: When doing replace all, accidentally replaced `real` with `viewport` in test names, changed back. now it is: `real viewports`

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+        it("returns adjusted values when change function has been called", () => {
+          const viewport = getViewport();
+          const bufferingViewport = new BufferingViewport(viewport);
+          changeDisplay(bufferingViewport);
+          expect(getDisplay(viewport)).toBe(true);
+          expect(getDisplay(bufferingViewport)).toBe(false);
+        });
+
+        it("adjusts viewport viewport when commit is called", () => {
```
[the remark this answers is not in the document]
JonasDov: When doing replace all, accidentally replaced `real` with `viewport` in test names, changed back. now it is: `real viewports`

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+          bufferingViewport.discard();
+          expect(getDisplay(viewport)).toBe(true);
+          expect(getDisplay(bufferingViewport)).toBe(true);
+        });
+      });
+    });
+  });
+
+  describe("per-model category overrides", () => {
+    it("returns viewport viewports result when change function has not been called", () => {
```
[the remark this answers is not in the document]
JonasDov: When doing replace all, accidentally replaced `real` with `viewport` in test names, changed back. now it is: `real viewports`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/AlwaysAndNeverDrawnElementInfoCache.ts
```diff
@@ -337,22 +338,27 @@ export class AlwaysAndNeverDrawnElementInfoCache implements Disposable {
...
     return forkJoin({
       alwaysDrawn: this.#viewport.alwaysDrawn?.size
-        ? this.getAlwaysOrNeverDrawnElements({ modelId: props.modelId, categoryIds: props.categoryIds, setType: "always" })
-        : of(new Set<Id64String>()),
-      neverDrawn: this.#viewport.neverDrawn?.size
-        ? this.getAlwaysOrNeverDrawnElements({ modelId: props.modelId, categoryIds: props.categoryIds, setType: "never" })
+        ? this.getAlwaysOrNeverDrawnElements({ modelId, categoryIds, setType: "always" })
         : of(new Set<Id64String>()),
+      neverDrawn: this.#viewport.neverDrawn?.size ? this.getAlwaysOrNeverDrawnElements({ modelId, categoryIds, setType: "never" }) : of(new Set<Id64String>()),
     }).pipe(
```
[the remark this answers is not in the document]
JonasDov: This is a valid observation in theory, but in practice the current code should be correct for all scenarios. I've added a comment to the code explaining the reasoning, and also added a `bufferingViewport.alwaysDrawn?.size` / `bufferingViewport.neverDrawn?.size` checks to not execute calculations if they're not needed.

The `this.#viewport.alwaysDrawn?.size` check for `getAlwaysOrNeverDrawnElements` is sufficient because the queried elements are only used to `setDifference` them from the buffered sets. Walking through the possible states of the buffered set relative to the real viewport:

1. **Sets are identical** - works as expected.
2. **Buffered set has elements already removed** - `setDifference` either removes the remaining matching elements or they were already gone. No issue.
3. **Buffered set has elements added** - elements added to always/never drawn during a `changeVisibility` call should not be removed within the same call, so `setDifference` correctly leaves them in place.

The deeper issue with switching to `bufferingViewport` here is that this cache is driven by real viewport change events - it rebuilds its data when `onAlwaysDrawnChanged` / `onNeverDrawnChanged` fire. The whole caching logic would need to be adjusted to take into concideration `bufferingViewport`, also it would likely execute additional queries that would in the very end produce the same outcome.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/AlwaysAndNeverDrawnElementInfoCache.ts
```diff
@@ -337,22 +338,38 @@ export class AlwaysAndNeverDrawnElementInfoCache implements Disposable {
...
-        ? this.getAlwaysOrNeverDrawnElements({ modelId: props.modelId, categoryIds: props.categoryIds, setType: "never" })
-        : of(new Set<Id64String>()),
+      alwaysDrawn:
+        this.#viewport.alwaysDrawn?.size && bufferingViewport.alwaysDrawn?.size
+          ? this.getAlwaysOrNeverDrawnElements({ modelId, categoryIds, setType: "always" })
+          : of(new Set<Id64String>()),
+      neverDrawn:
+        this.#viewport.neverDrawn?.size && bufferingViewport.neverDrawn?.size
+          ? this.getAlwaysOrNeverDrawnElements({ modelId, categoryIds, setType: "never" })
+          : of(new Set<Id64String>()),
```
grigasp: I don't understand this. As far as I can tell, `this.getAlwaysOrNeverDrawnElements` uses only `this.#viewport` to get the elements, so why are we checking `bufferingViewport` at all here?

grigasp (commented):
Overall I think the idea is good, but it's implemented in a very confusing way. We now have 2 viewports and I would have a very hard time deciding which one to use when checking visibility (changing should always use `BufferedViewport`?).

When we discussed possible ways to solve the problem, I was imagining a slightly different approach - refactoring our visibility handlers to capture the changes instead of making them on the viewport. The change in this PR to use a viewport object to capture those changes is maybe a good idea, but then we really need to get rid of the other viewport object, IMO.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/AlwaysAndNeverDrawnElementInfoCache.ts
JonasDov: No need to check it, removed.

JonasDov:
> Overall I think the idea is good, but it's implemented in a very confusing way. We now have 2 viewports and I would have a very hard time deciding which one to use when checking visibility (changing should always use `BufferedViewport`?).
> 
> When we discussed possible ways to solve the problem, I was imagining a slightly different approach - refactoring our visibility handlers to capture the changes instead of making them on the viewport. The change in this PR to use a viewport object to capture those changes is maybe a good idea, but then we really need to get rid of the other viewport object, IMO.

Adjusted the implementation. Now instead of passing around the `bufferingViewport`, we create a new VisibilityHandler on each `get/change` call.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/BufferingViewport.ts
```diff
@@ -0,0 +1,274 @@
...
+    }
+    return this.#neverDrawn === "cleared" ? undefined : this.#neverDrawn;
+  }
+
+  public get alwaysDrawn(): ReadonlySet<Id64String> | undefined {
+    if (this.#alwaysDrawn === undefined) {
+      return this.#realViewport.alwaysDrawn;
+    }
+    return this.#alwaysDrawn === "cleared" ? undefined : this.#alwaysDrawn;
+  }
```
grigasp, line 85: don't we have to merge sets of the real viewport and the buffered one?

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+import { describe, expect, it } from "vitest";
+import { BeEvent, Id64 } from "@itwin/core-bentley";
+import { type IModelConnection } from "@itwin/core-frontend";
+import { BufferingViewport } from "../../../../tree-widget-react/components/trees/common/internal/BufferingViewport.js";
+
+import type { Id64String } from "@itwin/core-bentley";
+import type { PerModelCategoryOverride, TreeWidgetViewport } from "../../../../tree-widget-react/components/trees/common/TreeWidgetViewport.js";
+
+function createMockViewport(props?: {
```
grigasp: why not reuse `createTreeWidgetTestingViewport`?

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      bufferingViewport.clearPerModelCategoryOverrides();
+      bufferingViewport.setPerModelCategoryOverride({ modelIds: "0x5", categoryIds: "0x6", override: "show" });
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("none");
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x3", categoryId: "0x4" })).toBe("none");
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x5", categoryId: "0x6" })).toBe("show");
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("show");
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x3", categoryId: "0x4" })).toBe("hide");
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x5", categoryId: "0x6" })).toBe("none");
+    });
+    it("clears overrides for specified models when clearPerModelCategoryOverrides is called", () => {
```
grigasp, line 190:
```suggestion
    });

    it("clears overrides for specified models when clearPerModelCategoryOverrides is called", () => {
```

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      }
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("hide");
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("show");
+    });
+
+    it("clears overrides when clearPerModelCategoryOverrides is called", () => {
+      const viewport = createMockViewport({
+        perModelCategoryOverrides: new Map([
+          ["0x1", new Map([["0x2", "show" as PerModelCategoryOverride]])],
+          ["0x3", new Map([["0x4", "hide" as PerModelCategoryOverride]])],
```
grigasp: do we need `as PerModelCategoryOverride` here?

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      const bufferingViewport = new BufferingViewport(viewport);
+      bufferingViewport.setPerModelCategoryOverride({ modelIds: "0x1", categoryIds: "0x2", override: "hide" });
+      bufferingViewport.clearPerModelCategoryOverrides({ modelIds: "0x3" });
+      bufferingViewport.commit();
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("hide");
+      expect(viewport.getPerModelCategoryOverride({ modelId: "0x3", categoryId: "0x4" })).toBe("none");
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x1", categoryId: "0x2" })).toBe("hide");
+      expect(bufferingViewport.getPerModelCategoryOverride({ modelId: "0x3", categoryId: "0x4" })).toBe("none");
+    });
+    it("discards changes when discard is called", () => {
```
grigasp, line 244:
```suggestion
    });

    it("discards changes when discard is called", () => {
```

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
JonasDov: Changed to use `createTreeWidgetTestingViewport` this is no longer needed

packages/itwin/tree-widget/src/test/trees/common/internal/BufferingViewport.test.ts
JonasDov: Reused `createTreeWidgetTestingViewport`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/BufferingViewport.ts
JonasDov: No, Drawn elements are set all at once. When i want to add an element to always/neverDrawn set, I have to create a new set and then call `setAlways/NeverDrawn`. So `#always/neverDrawn` inside the buffering viewport already contains all the necessary elements

packages/itwin/tree-widget/src/test/trees/common/internal/useTreeHooks/UseCachedVisibility.test.ts
```diff
@@ -0,0 +1,388 @@
...
+
+    it("suppresses and resumes event listeners during change", async () => {
+      const changeSubject = new Subject<void>();
+      const { handler: handler2, viewport: viewport2 } = createHandler({
+        visibilityHandler: {
+          changeVisibilityStatus: vi.fn(() => changeSubject),
+        },
+      });
+      const spy2 = vi.fn();
+      handler2.onVisibilityChange.addListener(spy2);
```
grigasp: why do these vars have the `2` suffix?

packages/itwin/tree-widget/src/test/trees/common/internal/useTreeHooks/UseCachedVisibility.test.ts
```diff
@@ -0,0 +1,388 @@
...
+      // Complete the change
+      changeSubject.complete();
+      await changePromise;
+
+      // After change, events should be resumed
+      viewport2.onDisplayedModelsChanged.raiseEvent();
+      await vi.advanceTimersByTimeAsync(0);
+      expect(spy2).toHaveBeenCalled();
+
+      handler2[Symbol.dispose]();
```
grigasp: shouldn't other tests dispose it, too?

maybe change the `createHandler` function to return a disposable object that disposes the handler, then do the following?
```ts
using setup = setupTest({ ... }); // renamed createHandler to setupTest
const { handler, viewport } = setup;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/useTreeHooks/UseCachedVisibility.ts
```diff
@@ -141,19 +143,21 @@ export class HierarchyVisibilityHandlerImpl<TSearchTargets> implements Hierarchy
...
-    );
+    this.#overrideHandler = new HierarchyVisibilityOverrideHandler(this);
   }
 
   public get onVisibilityChange() {
     return this.#eventListener.onVisibilityChange;
   }
 
   public async getVisibilityStatus(node: HierarchyNode): Promise<VisibilityStatus> {
+    const visibilityHelper = this.#props.getTreeSpecificVisibilityHandler({
```
grigasp: helper or handler? :)

packages/itwin/tree-widget/src/test/trees/common/internal/useTreeHooks/UseCachedVisibility.test.ts
JonasDov: Removed

packages/itwin/tree-widget/src/test/trees/common/internal/useTreeHooks/UseCachedVisibility.test.ts
JonasDov: Adjusted

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/useTreeHooks/UseCachedVisibility.ts
JonasDov: Renamed to handler

grigasp (approved):

PR #1659, opened 2026-04-20
Align measure-tools with iTwin.js 5.9 FormatSpecHandle API

### LINKED ISSUE #1658: Align `@itwin/measure-tools-react` with iTwin.js 5.9 quantity formatting APIs
The upcoming iTwin.js 5.9 release will introduce `FormatSpecHandle` and `getSpecsByNameAndUnit` in `@itwin/core-quantity`, replacing the manual `populateFormattingSpecsRegistry` / `getSpecsByName` pattern for quantity formatting (see [iTwin/itwinjs-core#9143](https://github.com/iTwin/itwinjs-core/pull/9143)). The new handles auto-refresh when the formatter reloads, eliminating the need for consumers to manually re-register format specs.

`@itwin/measure-tools-react` currently manages its own format spec registry population and listens to four separate QuantityFormatter events to stay in sync. Once 5.9 is released, this should be migrated to use `FormatSpecHandle` on each measurement subclass, simplifying event wiring in `MeasurementManager` and removing the `populateFormattingSpecsRegistry` override pattern entirely.

This will be a **minor** (breaking for pre-1.0) change — peer dependencies on `@itwin/core-*` will be bumped to `^5.9.0`.

hl662:
Closes #1658

> Tested locally by running the test-viewer and making sure nothing regressed after refactoring.

## Why

`measure-tools` was carrying its own formatter-spec lifecycle logic instead of following the upstream quantity-formatting model. That made measurement formatting harder to reason about, easier to desynchronize from formatter reload behavior, and more dependent on custom registry maintenance inside the package.

The test viewer also needed to exercise the same model without adding app-specific refresh machinery around measurement rendering. Format edits should continue to propagate into measurement tools while schema-defined units remain available and schema KoQs keep their original meaning.

This work also closes a correctness gap in the on-canvas measurement labels. Existing distance and area text markers were rebuilding asynchronously after format changes, but refresh timing could leave canvas labels stale even when the measurement widget had already updated.

## What

| Asset | Why it exists |
|---|---|
| `measure-tools` package manifest | Consumers need the updated iTwin.js peer range so the package can rely on the newer quantity-formatting APIs instead of preserving compatibility with older formatter behavior. |
| Measurement classes (`AngleMeasurement`, `AreaMeasurement`, `DistanceMeasurement`, `LocationMeasurement`, `RadiusMeasurement`) | Measurement instances need to own auto-refreshing `FormatSpecHandle`s so formatting follows the formatter reload model instead of duplicating registry maintenance in each tool path. |
| Formatting refresh pipeline (`MeasurementManager`, `Polygon`) | Existing canvas labels were repainting at the wrong time, so measure-tools now owns the deferred formatting refresh timing and aligns polygon text markers with the same formatter-spec path used by measurement widgets. |
| `measure-tools` test coverage | The formatting migration needed regression coverage for async refresh sequencing, area-label fallback behavior, and no longer needs schema-file bootstrapping just to resolve standard BIS units in tests. |
| `test-viewer` package manifest | The demo app needs dependency alignment with the updated quantity-formatting APIs used by `measure-tools`. |
| `test-viewer` units-provider setup (`Viewer.tsx`) | The viewer should compose schema units with bundled BIS units so iModel-defined units stay available without discarding the default provider behavior. |
| `test-viewer` format-set bridging (`FormatManager`, `QuantityFormatButton`, `FormatPropertyContextMenuItem`) | Quantity format edits need to repoint canonical measurement aliases to edited schema entries so test-viewer format changes continue to propagate into measurement tools without collapsing unrelated schema KoQs into one shared definition. |

Updated lockfiles to resolve the dependency changes.

## Tests

<details>
<summary>New and updated test coverage</summary>

| Test | What it covers |
|---|---|
| `Measurement.test.ts` — `Test measurement manager awaits async formatting refresh before invalidating` | Guards the regression where measurement decorations could invalidate before async label rebuilds completed after quantity format changes. |
| `Measurement.test.ts` — `Test measurement manager, stopDecorator does not cleanup managed measurements` | Guards the decorator lifecycle regression where stopping decoration could dispose still-managed measurements. |
| `AreaMeasurement.test.ts` — `Test fallback from getSpecFromHandle in setTextToMarker` | Guards the area text-marker path so polygon labels use the same fallback formatter-spec behavior as area measurements. |
| `pnpm test` in `packages/itwin/measure-tools` | Verifies the formatting migration still preserves existing measurement formatting and serialization behavior across the full suite. |
| `pnpm build` in `packages/itwin/measure-tools` | Verifies the refresh timing changes and package updates compile cleanly in the published package. |
| `pnpm build` in `apps/test-viewer` | Verifies the dependency set and canonical-alias format editing flow typecheck in the test application. |
| `pnpm lint` | Verifies the repo remains lint-clean after the formatting migration and test-viewer format-manager cleanup. |

</details>

---
*Nambot 🤖*


packages/itwin/measure-tools/src/measurements/AngleMeasurement.ts
```diff
@@ -157,22 +160,29 @@ export class AngleMeasurement extends Measurement {
...
+    if (!this._angleHandle) {
+      this._angleHandle = IModelApp.quantityFormatter.getFormatSpecHandle(
+        this._angleKoQ, this._anglePersistenceUnitName
+      );
     }
+    return this._angleHandle;
+  }
+
+  private _disposeHandles(): void {
+    this._angleHandle?.[Symbol.dispose]();
```
bsy-nicholasw: Prefer `this._disposable = dispose(this._disposable)`.  This is just the first instance I saw, it's a recommendation everywhere.

packages/itwin/measure-tools/src/measurements/AngleMeasurement.ts
bsy-nicholasw: `createTextMarker` can be made synchronous now. I see in the `FormatterUtils` many of them will optionally take a formatter spec. I dont think they're really needed, the format can be called directly from the handle (or using the handle - forget if the handle has a nice `handle.format(...)` helper or not.

I assume this will clean up _a lot_ of floating promises all around, across all the measurement types.

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -843,12 +843,9 @@ export abstract class Measurement {
   protected onLockToggled() { }
 
   /** Notify subclasses when the display units have changed. */
-  public onDisplayUnitsChanged(): void { }
+  public async onDisplayUnitsChanged(): Promise<void> { }
```
bsy-nicholasw: I'm thinking this does not actually need to be async.

If all formats are resolved, and our helpers to create text markers can become synchronous, I dont think any measurement will actually need this to be async.

I think a new callback similar to `populateFormattingSpecsRegistry` may still be needed though, to allow measurements to register any custom formats they may need though.

packages/itwin/measure-tools/src/api/Polygon.ts
```diff
@@ -29,6 +30,7 @@ export class Polygon {
   private _styleSet: StyleSet;
   private _areaKoQ: string;
```
bsy-nicholasw: Isn't the polygon a helper class used by other measurements? Wouldn't whoever use it just set a format specs handle to it? If none is set, then can fallback to `DefaultToolsUnits.AREA`?

It would eliminate the need to supply so many extra values, which should all be self contained by handle now.

packages/itwin/measure-tools/src/api/Measurement.ts
hl662: 🤖 **Nambot** — Agreed. After making the text-marker refresh paths synchronous, I changed `Measurement.onDisplayUnitsChanged` back to a sync hook and simplified `MeasurementManager._onFormattingRefresh()` to iterate measurements directly before invalidating/notifying.

I did not add a new registration callback in this PR, since none of the current measurements need extra custom format pre-registration beyond the existing handle-based path. If we want to harden the extensibility story further, that feels like a separate follow-up.

Nambot 🤖

packages/itwin/measure-tools/src/measurements/AngleMeasurement.ts
hl662: 🤖 **Nambot** — Good catch. I took this a bit further than just making `onDisplayUnitsChanged` async here: `createTextMarker` is synchronous now in the measurement types that use `FormatSpecHandle`, so the floating promise at this call site is gone entirely rather than being awaited through an async helper. That also let me simplify the formatting refresh path in `MeasurementManager`.

Nambot 🤖

packages/itwin/measure-tools/src/measurements/AngleMeasurement.ts
hl662: 🤖 **Nambot** — Yep, this was the right direction. I switched the text-marker paths backed by `FormatSpecHandle` over to synchronous formatting via the handle in `AngleMeasurement`, `DistanceMeasurement`, `LocationMeasurement`, and `RadiusMeasurement`, and removed the fire-and-forget marker refresh calls that fell out of that.

I left the broader widget/data formatting helpers alone for now, since those still have legitimate async call sites.

Nambot 🤖

packages/itwin/measure-tools/src/api/Measurement.ts
bsy-nicholasw: And if it can remain synchronous, then that eliminates the need for that schedule update timer in the manager.

saskliutas (approved):
Approving test-viewer changes

a-gagnon (approved):
LGTM. Nicholas already did a good job with adding comments :)

hl662:
🤖 **Nambot** —

Since the earlier review pass, I pulled in the `5.9.2` core packages and updated `measure-tools` to rely on the `FormatSpecHandle` stale-state fix there.

### What changed

- bumped the local `5.9` exact pins used here to `5.9.2`
- bumped `@itwin/measure-tools-react`’s relevant `5.9` peer floor to `^5.9.2`
- removed the deferred formatting refresh timer from `MeasurementManager`
  - with `5.9.2`, `FormatSpecHandle` now reads current spec state on access, so the old listener-ordering workaround is no longer needed
- kept the broader formatting cleanup from this branch:
  - sync marker/text refresh paths where handle/immediate formatting is sufficient
  - `Polygon` simplification so it is provider/handle-driven instead of duplicating area KoQ/unit state
  - widget-data internal cleanup to reduce fake async
  - bearing refactor so it uses async prep/registration once, then sync handle-based formatting afterward

### Validation

Re-ran:
- `cd packages/itwin/measure-tools && pnpm test`
- `cd packages/itwin/measure-tools && pnpm build`
- `cd packages/itwin/measure-tools && pnpm lint`
- `cd apps/test-viewer && pnpm build`
- `cd apps/test-viewer && pnpm lint`

Also verified the test app behavior locally after consuming `5.9.2`. (Human edit)

change/@itwin-measure-tools-react-766165da-34b9-4dc5-b71c-2e570d666b28.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Update peer dependencies to iTwin.js 5.9, migrate measurement formatting to the FormatSpecHandle-based API, and fix existing measurement labels to refresh immediately after quantity format changes.",
```
JonasDov, line 3: I think this should be a major version bump since it involves peer dependency changes, which are breaking for consumers.

saskliutas: This package is still in 0.x version so `minor = major`

packages/itwin/measure-tools/src/api/FormatterUtils.ts
```diff
@@ -223,22 +280,62 @@ export namespace FormatterUtils {
...
+            persistenceUnitName,
+            formatProps,
+            system,
+          });
+        }
+      })();
+      bearingRegistrationPromises.set(key, promise);
+    }
+
+    try {
```
a-gagnon: This is only needed for the first call that initiates the promise, but not a dealbreaker if subsequent calls try/catch too..

packages/itwin/measure-tools/src/api/Measurement.ts
```diff
@@ -649,20 +649,28 @@ export abstract class Measurement {
...
    */
-  protected async getDataForMeasurementWidgetInternal(): Promise<MeasurementWidgetData | undefined> { return undefined; }
+  protected getDataForMeasurementWidgetInternal(): MeasurementWidgetData | undefined | Promise<MeasurementWidgetData | undefined> { return undefined; }
+
+  /** Allows subclasses to asynchronously prepare any custom formatting state before widget data is gathered. */
+  protected async prepareFormatting(): Promise<void> { }
+
+  private async applyDataForMeasurementWidgetHandlers(data: MeasurementWidgetData): Promise<void> {
+    for (const handler of Measurement._dataForMeasurementWidgetHandlers.sort((h) => h.priority))
+      await handler.handlerFunction(this, data);
```
a-gagnon, line 672: I'm a bit surprised the compiler doesn't complain since it's either sync or async...
The right way of handling both is to do `await Promise.resolve(handler.handlerFunction(this, data))`.

packages/itwin/measure-tools/src/api/Measurement.ts
hl662: I (and my bot) went a little overboard and cleaned a lot of the async stuff in measure-tools, including the schedule update timer in the manager : P

packages/itwin/measure-tools/src/api/Polygon.ts
hl662: good point, I resturctured it to be more self contained now.

packages/itwin/measure-tools/src/api/Measurement.ts
hl662: I think the current form already handles both cases correctly... `handlerFunction` is typed as `void | Promise<void>`, and `await` already normalizes both sync and async returns, so `await Promise.resolve(handler.handlerFunction(this, data))` would be equivalent rather than functionally different.

I'd prefer to keep the simpler form unless we want the extra explicitness for style reasons?

packages/itwin/measure-tools/src/api/FormatterUtils.ts
hl662: good point, I seplit the bearing registration into its own method and the promise cleanup as a separate to make it clearer and clean up dis

packages/itwin/measure-tools/src/api/Measurement.ts
bsy-nicholasw: Unless if I missed something, doesn't look overboard at all.

bsy-nicholasw (approved):

PR #1660, opened 2026-04-21
[Tree widget[: bump presentation dependencies on `next`

JonasDov:


grigasp (approved):
Being the first consumer of the select clause factories' breaking changes, how do you feel about them? Anything you think could be improved?

JonasDov:
> Being the first consumer of the select clause factories' breaking changes, how do you feel about them? Anything you think could be improved?

Nothing comes to mind, the transition was simple.
I think the addition of clause factories in `DefineHierarchyLevelProps` was a good choice, they simplified the code base a little bit.

PR #1665, opened 2026-04-24
Switch to @dnd-kit in Map Layers

### LINKED ISSUE #1449: Migrate from react-beautiful-dnd to dnd-kit
The `react-beautiful-dnd` package used in this repo is now [archived](https://github.com/atlassian/react-beautiful-dnd) on github and [deprecated](https://www.npmjs.com/package/react-beautiful-dnd) on npm. It is no longer being maintained, yet it has some known issues examples of which are discussed in these two issues:

https://github.com/atlassian/react-beautiful-dnd/issues/2396
https://github.com/iTwin/viewer-components-react/issues/1390

We need to drop our dependency on it by migrating to [dnd-kit](https://dndkit.com/) which in our current opinion is the most stable and future-proof option and is [recommended by Atlassian](https://github.com/atlassian/react-beautiful-dnd/issues/2672#issue-2587187019) as a way forward.

arome:
Updated map layers to @dnd-kit to fix console errors happening due to https://github.com/atlassian/react-beautiful-dnd being deprecated.
image 
image
Also applied minimal styling to the layer 

and to the loading


packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+      style={{
+        zIndex: props.sortable.isDragging ? 1 : undefined,
+        boxShadow: props.sortable.isDragging ? "10px 5px 5px rgba(0, 0, 0, 0.15)" : undefined,
+      }}
+    >
+      <Checkbox
+        data-testid={"select-item-checkbox"}
+        checked={props.activeLayer.selected}
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+          props.activeLayer.selected = event.target.checked;
```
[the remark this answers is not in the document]
arome: done

packages/itwin/map-layers/src/ui/widget/map-layer/SortableMapLayerItem.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function SortableMapLayerItem(props: SortableMapLayerItemProps) {
+  assert(props.layer !== undefined);
+
+  const sortable = useSortable({
+    id: props.layer.id,
+    index: props.index,
+    type: 'item',
+    accept: 'item',
```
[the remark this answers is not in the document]
arome: done

packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx
```diff
@@ -534,32 +258,32 @@ export function MapLayerManager(props: MapLayerManagerProps) {
...
-
-      if (isOverlay === undefined || isOverlay) {
-        overlayMapLayers?.forEach((layer) => changeLayerVisibility(!layer.visible, layer.layerIndex, layer.isOverlay));
+      if (isOverlay === undefined) {
+        if (isOverlay) {
+          overlayLayers.forEach((layer) => changeLayerVisibility(!layer.visible, layer.layerIndex, layer.isOverlay));
+        } else {
+          backgroundLayers.forEach((layer) => changeLayerVisibility(!layer.visible, layer.layerIndex, layer.isOverlay));
+        }
       }
```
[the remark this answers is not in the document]
arome: done

packages/itwin/map-layers/src/ui/widget/MapLayerDroppable.tsx
```diff
@@ -4,36 +4,29 @@
...
-import { Checkbox, IconButton } from "@itwin/itwinui-react";
+import { IModelApp, NotifyMessageDetails, OutputMessagePriority } from "@itwin/core-frontend";
 import { MapLayersUI } from "../../mapLayers";
-import { AttachLayerButtonType, AttachLayerPopupButton } from "./AttachLayerPopupButton";
-import { MapLayerSettingsMenu } from "./MapLayerSettingsMenu";
-import { MapUrlDialog } from "./MapUrlDialog";
-import { SubLayersPopupButton } from "./SubLayersPopupButton";
+import { AttachLayerButtonType, AttachLayerPopupButton } from "./map-layer/AttachLayerPopupButton";
+import { backgroundMapLayersId, overlayMapLayersId } from "./map-layer/MapLayerDragDrop";
+import { useDroppable } from '@dnd-kit/react';
```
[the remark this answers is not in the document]
arome: done

packages/itwin/map-layers/src/ui/widget/SortableMapLayerItem.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useSortable } from "@dnd-kit/react/sortable";
+import type { StyleMapLayerSettings } from "../../Interfaces";
+import type { backgroundMapLayersId, overlayMapLayersId } from "./MapLayerDragDrop";
+import { assert } from "@itwin/core-bentley";
+
+interface SortableMapLayerItemProps {
+  layer: StyleMapLayerSettings;
+  droppableId: typeof overlayMapLayersId | typeof backgroundMapLayersId;
+  index: number;
+  renderItem: (sortable: ReturnType<typeof useSortable>) => React.ReactNode;
```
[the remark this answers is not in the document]
arome: done

packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+      >
+        {props.activeLayer.visible ? <SvgVisibilityShow data-testid="layer-visibility-icon-show" /> : <SvgVisibilityHide data-testid="layer-visibility-icon-hide" />}
+      </IconButton>
+
+      <span
+        className={props.disabled || outOfRange ? "map-manager-item-label-disabled" : "map-manager-item-label"}
+        title={outOfRange ? props.outOfRangeTitle : undefined}
+      >
+        {props.activeLayer.name}
+        {/* eslint-disable-next-line @itwin/no-internal */}
```
eringram:
```suggestion
```

I think this `eslint-disable` is unnecessary now, `MapLayerImageryProvider.status` has become public preview and `MapLayerImageryProviderStatus` is public

packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+            expandMode="rootGroupOnly"
+            subLayers={props.activeViewport ? props.activeLayer.subLayers : undefined}
+            singleVisibleSubLayer={props.activeLayer.provider?.mutualExclusiveSubLayer}
+            onSubLayerStateChange={(subLayerId: SubLayerId, isSelected: boolean) => {
+              props.onSubLayerStateChange(props.activeLayer, subLayerId, isSelected);
+            }}
+          />
+        )}
+      </div>
+      {/* eslint-disable-next-line @itwin/no-internal */}
```
eringram:
```suggestion
```

eringram (commented):
Code looks good to me but didn't manually test yet

Ignorant question since I don't work on this repo, where is the ui-test-app mentioned in the README (or what's the current way to test these components)?

arome:
> Code looks good to me but didn't manually test yet
> 
> Ignorant question since I don't work on this repo, where is the ui-test-app mentioned in the README (or what's the current way to test these components)?

It's in viewer-components-react/apps/test-viewer you can run `pnpm start` from there (probably have to first run `pnpm i` and `pnpm build` on the whole repo).
Also make sure to have your .env file, there's the apps/test-viewer/.env.template as a template

packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx
arome: Updated the devDependency of itwin/core-frontend to v5.6.0 in order to be able to remove this comment

packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx
arome: fixed same comment as above

eringram:
> It's in viewer-components-react/apps/test-viewer you can run `pnpm start` from there (probably have to first run `pnpm i` and `pnpm build` on the whole repo). Also make sure to have your .env file, there's the apps/test-viewer/.env.template as a template

Thanks, I was able to run it. I think the test-viewer pnpm-lock.yaml also needs to be updated but after that I'll approve

packages/itwin/map-layers/src/ui/widget/MapLayerDragDrop.ts
```diff
@@ -0,0 +1,102 @@
...
+  const sourceDroppableId = draggedId.slice(0, firstColon);
+  if (!isMapLayerDroppableId(sourceDroppableId)) {
+    return false;
+  }
+
+  // Locate the dragged item in the post-drag-over state to find its destination.
+  for (const destinationDroppableId of [backgroundMapLayersId, overlayMapLayersId] as const) {
+    const layers = mapLayers[destinationDroppableId];
+    const destinationUiIndex = layers.findIndex((l) => l.id === draggedId);
+    if (destinationUiIndex < 0) {
```
simihartstein: 🤖 _simi-bot_ | 🐛 🔴 **[Drop Handling]** — This can persist a reorder even when the user releases outside any droppable.

[MapLayerManager.tsx:180-195](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx#L180-L195) eagerly updates `mapLayersRef.current` during `onDragOver`, and this helper then commits by scanning that optimistic state instead of checking the final `event.operation.target`. If the pointer last hovered another slot/list and the user drops over empty space, the last hovered arrangement is still written to `DisplayStyle`.

```ts
for (const destinationDroppableId of [backgroundMapLayersId, overlayMapLayersId] as const) {
  const layers = mapLayers[destinationDroppableId];
  const destinationUiIndex = layers.findIndex((l) => l.id === draggedId);
```

The old `react-beautiful-dnd` path explicitly no-op'd when there was no final destination, so this changes drop semantics.

packages/itwin/map-layers/package.json
simihartstein, line 1: 🤖 _simi-bot_ | 🐛 🟡 **[Published Compatibility]** — This package now requires a `core-frontend` API that only became public in `5.6.0`, but the published peer range still advertises `^5.0.0`.

[MapLayerItem.tsx:6-7](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/ui/widget/MapLayerItem.tsx#L6-L7) now imports and uses `MapLayerImageryProviderStatus`, and the PR also raises the local dev floor at [package.json:51-53](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/package.json#L51-L53). Keeping the published peer range at [package.json:95-101](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/package.json#L95-L101) means consumers on `5.0.x` through `5.5.x` can still satisfy the manifest even though this build no longer targets them.

packages/itwin/map-layers/src/test/MapLayerManager.test.tsx
simihartstein, line 1: 🤖 _simi-bot_ | 🧪 🟡 **[Coverage Gap]** — The PR replaces the core reorder and cross-list implementation but adds no direct tests for that new path.

[MapLayerManager.tsx:180-225](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/ui/widget/MapLayerManager.tsx#L180-L225) now owns optimistic `onDragOver`, deferred `onDragEnd`, cancel rollback, and cross-list persistence through [MapLayerDragDrop.ts:36-90](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/ui/widget/MapLayerDragDrop.ts#L36-L90). But [MapLayerManager.test.tsx:90-310](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/test/MapLayerManager.test.tsx#L90-L310) still only covers rendering, source updates, selection, detach, and visibility changes.

I did not find any package tests covering same-list reorder, cross-list moves, empty-list drops, outside-drop no-op behavior, or canceled-drag rollback, so regressions in the new DnD path are currently unguarded.

simihartstein (commented):
🤖 _simi-bot_ | Posted 3 findings covering one behavior regression, one published compatibility risk, and one missing-test gap.

packages/itwin/map-layers/package.json
arome: Addressed in d9019593. Updated peer dependencies to require @itwin/core-{bentley,common,frontend}@5.6.0+ to match the actual runtime requirements now that MapLayerImageryProviderStatus is used.

packages/itwin/map-layers/src/test/MapLayerManager.test.tsx
arome: Addressed in the latest changes. I added manager-level drag/drop coverage in MapLayerManagerDragDrop.test.tsx for rollback when leaving all droppables, rollback when dropping outside, and projection into a list that was empty at drag start. I also added focused tests around the extracted drag controller behavior so the dnd-kit interpretation/projection decisions are covered directly.

packages/itwin/map-layers/src/test/MapLayerDragModel.test.ts
```diff
@@ -0,0 +1,235 @@
...
+          { id: `${backgroundMapLayersId}:Layer2`, layerIndex: 1 },
+        ],
+        [overlayMapLayersId]: [],
+      };
+
+      const result = commitMapLayerDrop(displayStyle, mapLayers, event);
+      expect(result).toBe(true);
+      // destinationUiIndex = 0, so destinationDisplayStyleIndex = 2 - 1 - 0 = 1
+      expect(displayStyle.moveMapLayerToIndex).toHaveBeenCalledWith(0, 1, true);
+    });
```
simihartstein, line 154: 🤖 _simi-bot_ | 🧪 🟡 **Test correctness** — The third argument to `moveMapLayerToIndex` is the `isOverlay` flag. The source ID is `backgroundMapLayers:Layer1`, so `sourceIsOverlay = false` and `destinationIsOverlay = false` — both background. The assertion passes `true`, which will cause this test to fail against the current implementation.

```suggestion
      expect(displayStyle.moveMapLayerToIndex).toHaveBeenCalledWith(0, 1, false);
```

packages/itwin/map-layers/src/test/MapLayerDragModel.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      const mapLayers = {
+        [backgroundMapLayersId]: [],
+        [overlayMapLayersId]: [
+          { id: itemId, layerIndex: 1 },
+          { id: `${overlayMapLayersId}:Layer2`, layerIndex: 0 },
+        ],
+      };
+
+      const result = commitMapLayerDrop(displayStyle, mapLayers, event);
+      expect(result).toBe(true);
```
simihartstein, line 174: 🤖 _simi-bot_ | 🧪 🟡 **Test correctness — overlay reorder setup causes a no-op** — Layer1 is placed at `uiIndex=0` with `layerIndex=1`. The implementation computes `destinationDisplayStyleIndex = length(2) - 1 - uiIndex(0) = 1`, and `sourceDisplayStyleIndex = layerIndex = 1`. Because `1 === 1`, [MapLayerDragDrop.ts](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/ui/widget/MapLayerDragDrop.ts#L84-L88) returns `false` and skips the move. The test expects `true` and a `moveMapLayerToIndex` call — both will fail.

To exercise a genuine reorder the layer must start at a display-style index that differs from where the UI positions it after the drag. For example, place Layer1 at `uiIndex=0, layerIndex=0` (Layer2 at `uiIndex=1, layerIndex=1`): source index = 0, destination index = 2 - 1 - 0 = 1 — they differ, the move fires.

```suggestion
      const mapLayers = {
        [backgroundMapLayersId]: [],
        [overlayMapLayersId]: [
          { id: itemId, layerIndex: 0 },
          { id: `${overlayMapLayersId}:Layer2`, layerIndex: 1 },
        ],
      };
```
And update the assertion:
```suggestion
      expect(displayStyle.moveMapLayerToIndex).toHaveBeenCalledWith(0, 1, true);
```

packages/itwin/map-layers/src/test/MapLayerDragDrop.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        [overlayMapLayersId]: [],
+      };
+
+      const result = commitMapLayerDrop(displayStyle, mapLayers, event);
+      // destinationUiIndex = 1, destinationDisplayStyleIndex = 2 - 1 - 1 = 0
+      // sourceDisplayStyleIndex = 1, so sourceDisplayStyleIndex !== destinationDisplayStyleIndex
+      // Actually this will NOT be a no-op. Let me make a true no-op where the layer doesn't move
+      expect(result).toBe(true);
+      // This is actually a move from index 1 to 0
+      expect(displayStyle.moveMapLayerToIndex).toHaveBeenCalledWith(1, 0, true);
```
simihartstein: 🤖 _simi-bot_ | 🧪 🟡 **Test correctness — "no-op" test is not a no-op and uses the wrong `isOverlay` flag** — Two separate problems here:

1. **Not a no-op.** Layer1 sits at `uiIndex=1, layerIndex=1`. `destinationDisplayStyleIndex = 2 - 1 - 1 = 0`, `sourceDisplayStyleIndex = 1`. They differ, so the implementation calls `moveMapLayerToIndex` and returns `true`. The test name and the comment inside it acknowledge this ("Actually this will NOT be a no-op") but the name was never updated. This is dead-weight confusion, not a tested invariant.

2. **Wrong `isOverlay` flag.** The source ID is `backgroundMapLayers:Layer1`, so `sourceIsOverlay = false`. The assertion passes `true` as the third arg — it will fail against the actual implementation call.

The [existing "no-op" test at line 210 of MapLayerDragDrop.test.ts](https://github.com/iTwin/viewer-components-react/blob/omar/switch-to-dnd-kit/packages/itwin/map-layers/src/test/MapLayerDragDrop.test.ts#L210) (the one that correctly expects `false`) already covers the same-position case. This block should either be renamed and corrected, or removed.

Suggested fix — turn it into a genuine move test with the right flag:
```suggestion
    it("should reorder background layer to a lower display-style index", () => {
      const displayStyle = createMockDisplayStyle();
      const event = createMockEvent();
      const itemId = `${backgroundMapLayersId}:Layer1`;
      // Layer1 at uiIndex=1, layerIndex=1 → destinationDisplayStyleIndex = 2-1-1 = 0
      // 1 !== 0, so a real move occurs.
      const mapLayers = {
        [backgroundMapLayersId]: [
          { id: `${backgroundMapLayersId}:Layer2`, layerIndex: 0 },
          { id: itemId, layerIndex: 1 },
        ],
        [overlayMapLayersId]: [],
      };

      const result = commitMapLayerDrop(displayStyle, mapLayers, event);
      expect(result).toBe(true);
      expect(displayStyle.moveMapLayerToIndex).toHaveBeenCalledWith(1, 0, false);
    });
```

simihartstein (commented):
🤖 _simi-bot_ | Follow-up on the test fixes for the three findings addressed in this PR. The two source-code fixes (outside-drop guard and peer range bump) look correct. Three of the new `MapLayerDragDrop.test.ts` tests have incorrect setups or assertions that will cause them to fail or silently validate the wrong contract — see the inline comments.

arome:
> 🤖 _simi-bot_ | Follow-up on the test fixes for the three findings addressed in this PR. The two source-code fixes (outside-drop guard and peer range bump) look correct. Three of the new `MapLayerDragDrop.test.ts` tests have incorrect setups or assertions that will cause them to fail or silently validate the wrong contract — see the inline comments.

actually im still working on the tests and the fixes

packages/itwin/map-layers/src/ui/widget/MapLayerDragDrop.ts
arome: Addressed in the latest changes. commitMapLayerDrop now explicitly no-ops when the final event.operation.target is missing, so the optimistic hover state is never persisted when the user releases outside a droppable. The drag hook also restores the drag-start snapshot when leaving/dropping outside, preserving the old react-beautiful-dnd semantics.

packages/itwin/map-layers/src/test/MapLayerDragModel.test.ts
arome: fixed

packages/itwin/map-layers/src/test/MapLayerDragModel.test.ts
arome: done

packages/itwin/map-layers/src/test/MapLayerDragDrop.test.ts
arome: done

packages/itwin/map-layers/package.json
```diff
@@ -95,9 +96,9 @@
     "@itwin/appui-abstract": "^5.0.0",
     "@itwin/appui-react": "^5.5.0",
     "@itwin/components-react": "^5.5.0",
-    "@itwin/core-bentley": "^5.0.0",
-    "@itwin/core-common": "^5.0.0",
-    "@itwin/core-frontend": "^5.0.0",
+    "@itwin/core-bentley": "^5.6.0",
```
saskliutas, line 99: Bumping peer deps is a breaking change and should result in major version bump.

saskliutas (approved):
Approving test-viewer changes

packages/itwin/map-layers/package.json
arome: Addressed in 2089add8 — updated the change file type to `major` to reflect the breaking peer dep version bump.

eringram (approved):

change/@itwin-map-layers-ba2ed48c-750d-4f72-a3f2-e17bc24f03b6.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
aruniverse, line 2: why was this a major bump?
isnt this just an underlying implementation change?

saskliutas: Peer dep bump https://github.com/iTwin/viewer-components-react/pull/1665#discussion_r3193733045

PR #1675, opened 2026-05-08
Bump pnpm to new major version

saskliutas:
Bumped pnpm to v11. Moved pnpm config from `.npmrc` to `pnpm-workspace.yaml` file

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -3,3 +3,29 @@ packages:
   - apps/test-viewer
   - apps/learning-snippets
   - apps/performance-tests
+
+packageConfigs:
+  "test-viewer":
+    overrides:
+      "@itwin/itwinui-react@^2": "2.12.26"
+      "@itwin/itwinui-react@^3": "3.20.1"
+
+
+allowBuilds:
+  '@bentley/imodeljs-native': true
+  '@itwin/certa': true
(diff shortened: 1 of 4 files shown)


pnpm-workspace.yaml
```diff
@@ -3,3 +3,29 @@ packages:
...
+  '@bentley/imodeljs-native': true
+  '@itwin/certa': true
+  '@itwin/oidc-signin-tool': true
+  '@parcel/watcher': false
+  esbuild: true
+  puppeteer: true
+  typemoq: true
+
+syncInjectedDepsAfterScripts:
+  - "build"
```
grigasp, line 24: This looks interesting. Am I right to think it should allow us to remove the `linkWorkspaceDeps` scripts?

saskliutas: That's the theory behind the option. However, vite hot reload does not work with it. However running `pnpm build` in tree-widget and restarting vite devServer manually works.

saskliutas: However, I think other link deps like tree-widget where we don't need hot reload can be removed I think.

saskliutas: Didn't work on initial lage build.

grigasp (approved):

aruniverse (approved):

PR #1676, opened 2026-05-11
Remove usage of deprecated iModel editing APIs

saskliutas:
Update test-utilities, tree-widget, learning-snippets and performance tests to use new EditTxn API to create iModel for test.


aruniverse (approved):

JonasDov (approved):

packages/itwin/test-utilities/package.json
```diff
@@ -18,29 +18,31 @@
     "./package.json": "./package.json"
   },
   "peerDependencies": {
+    "@itwin/core-backend": "^5.0.0",
     "@itwin/core-bentley": "^5.0.0",
     "@itwin/core-common": "^5.0.0",
     "@itwin/core-frontend": "^5.0.0",
+    "@itwin/presentation-backend": "^5.0.0",
```
grigasp, line 25: I intentionally didn't place these into peers to avoid having to add them as dev-deps to tree-widget-react package. 
- Was there a reason for this move?
- Doesn't `pnpm install` complain about missing peers when installing tree-widget-react package?

grigasp (commented):
These changes should allow us to remove some of `/* eslint-disable @typescript-eslint/no-deprecated */` comments, right?

packages/itwin/test-utilities/package.json
saskliutas: `pnpm install` does not complain because we have `core-backend` and `presentation-backend` in dev deps https://github.com/iTwin/viewer-components-react/blob/6b6f98aa108bca7ebb718f294dfa264a1d55f310/packages/itwin/tree-widget/package.json#L95

I guess there is no reason to add `presentation-backend` but for `core-backend` we expose classes like IModelDb and with a new API we need to use `withEditTxn` on consuming side (tree-widget) so we will need to have `core-backend` dep.

grigasp (approved):

PR #1681, opened 2026-05-18
[Tree widget]: add abstract batching cache on `next`

JonasDov:
When reviewing https://github.com/iTwin/viewer-components-react/pull/1678,
Realized I hadn't included the `ChildElementsCache` rewrite in the original phases. Thinking we could insert it between phase 1 and phase 2.

While trying to implement `ChildElementsCache`, considered unifying how it works with `DescendantCountsCache`. Currently `ChildElementsCache` executes a query per request, while `DescendantCountsCache` batches requests over a 20ms window. The original plan was to keep the structure as-is, but decided that it's better to batch `ChildElementsCache` as well.

This PR includes:
1. Abstract `BatchingCache` class which is used by `DescendantCountsCache`.
2. `DescendantCountsCache` changes to use `BatchingCache`.

Afterwards will create a separate PR for adjusting `ChildElementsCache`.

### DIFF
--- a/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BatchingCache.ts
+++ b/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BatchingCache.ts
@@ -27,14 +27,14 @@ export interface BatchingCacheProps {
  *
  * @template TRequest - A single logical request made by a consumer (e.g. `{ modelId, categoryId, parentElementId }`).
  * @template TResult - The value returned to the consumer for a given request (e.g. `Id64Array`).
- * @template TQueryData - Query data produced by decomposing a batch of requests via `getIterable`
+ * @template TQueryData - Query data produced by decomposing a batch of requests via `getQueryData`
  *   (e.g. a WHERE clause fragment). Items are buffered (up to `bufferSize`) and passed to `executeQuery`.
  * @template TRow - A single result row emitted by `executeQuery`, cached via `insertRow`
  *   (e.g. `{ modelId, reqParent, reqCategory, ownCategory, count }`).
  *
  * Pipeline:
  * 1. Requests arriving within `timerDelay` ms are collected into a batch (`TRequest[]`).
- * 2. `getIterable(batch)` decomposes the batch into a stream of `TQueryData` items.
+ * 2. `getQueryData(batch)` decomposes the batch into a stream of `TQueryData` items.
  * 3. Items are buffered (up to `bufferSize`) and passed to `executeQuery(items)`.
  * 4. Each `TRow` emitted by `executeQuery` is cached via `insertRow`.
  * 5. After completion, `ensureDefaultCacheEntries` fills in empty entries for
@@ -79,7 +79,7 @@ export abstract class BatchingCache<TRequest, TResult, TQueryData, TRow> {
    * those values take shape of a WHERE clause fragments.
    * These TQueryData items are buffered and passed to `executeQuery`.
    */
-  protected abstract getIterable(batch: TRequest[]): Observable<TQueryData>;
+  protected abstract getQueryData(batch: TRequest[]): Observable<TQueryData>;
 
   /** Execute a query for the given query data buffer. Returns an observable of result rows. */
   protected abstract executeQuery(queryData: TQueryData[]): Observable<TRow>;
@@ -150,7 +150,7 @@ export abstract class BatchingCache<TRequest, TResult, TQueryData, TRow> {
   }
 
   private executeBatchQuery(batch: TRequest[]): Observable<TRow> {
-    return this.getIterable(batch).pipe(
+    return this.getQueryData(batch).pipe(
       bufferCount(this.#bufferSize),
       mergeMap((queryData: TQueryData[]) => this.executeQuery(queryData).pipe(catchBeSQLiteInterrupts)),
       releaseMainThreadOnItemsCount(this.#releaseOnCount),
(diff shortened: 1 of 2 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BatchingCache.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    // Check if request is fully covered by an in-flight batch
+    let requestNotInBatch: TRequest = request;
+    for (const { values, sharedObs } of this.#requestedValues.values()) {
+      const partOfRequestNotInBatch = this.getValuesToRequest(request, values);
+      if (partOfRequestNotInBatch === undefined) {
+        return this.getResultAfterObservable(request, sharedObs);
+      }
+      requestNotInBatch = partOfRequestNotInBatch;
+    }
```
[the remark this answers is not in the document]
JonasDov: Changed:
- renamed `getValuesToRequest` -> `getValuesNotInBatch`
- old: `this.getValuesToRequest(request, values)`; new `this.getValuesNotInBatch(requestNotInBatch, values)`
- `getValuesNotInBatch` now returns an object, which indicates if some part of request was in batch or not. This is used to collect `sharedObs` instead of just using `this.#valuesToRequest.sharedObs`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/DescendantsCountCache.ts
```diff
@@ -29,60 +26,86 @@ interface DescendantsCountElementRequest extends DescendantsCountBaseRequest {
...
-        const entry = this.#cachedValues.get(modelId)?.get(parentElementId)?.get(categoryId);
-        assert(entry !== undefined);
-        return entry;
-      }),
-    );
+  protected getCachedValue(request: DescendantsCountRequest): DescendantsCountResult | undefined {
+    return this.#cachedValues.get(request.modelId)?.get(request.parentElementId)?.get(request.categoryId);
+  }
+
+  protected getGuaranteedCachedValue(request: DescendantsCountRequest): DescendantsCountResult {
```
saskliutas: Do we really need these 2: `getCachedValue` and `getGuaranteedCachedValue`? Maybe we can call  `getCachedValue` in `getResultAfterObservable` and assert there? I struggle to imagine scenario where implementation of both these functions will be different.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/DescendantsCountCache.ts
```diff
@@ -29,60 +26,86 @@ interface DescendantsCountElementRequest extends DescendantsCountBaseRequest {
...
+    return { valuesNotInBatch: request, batchContainsValues: false };
   }
 
-  private executeBatchQuery(valuesToRequest: RequestEntry) {
-    return from(valuesToRequest.entries()).pipe(
+  protected createBatch(): Batch {
+    return new Map();
+  }
+
+  protected addRequestToBatch(request: DescendantsCountRequest, batch: Batch): void {
```
saskliutas: What do you think about changing `BatchingCache` to take care of batching? Instead of leaving it up to concrete implementation to implement `createBatch`, `addToBatch`, etc, it looks like batch could be just `Array<TRequest>` in `BatchingCache`. 

For concrete implementation it should be enough to implement: `compareBatchItem` and `getIterable` where it would convert from Array<TReqeust> into input for `executeQuery`.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/DescendantsCountCache.ts
JonasDov: Removed `getGuaranteedCachedValue`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/DescendantsCountCache.ts
JonasDov: Removed `createBatch()` and `addRequestToBatch(request: TRequest, batch: TBatch)`;
now batch is of type `TRequest[]`

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BatchingCache.ts
```diff
@@ -0,0 +1,154 @@
...
+   * Returns `undefined` if the entire request is already in the batch.
+   * For caches without partial requests, return `undefined` if in batch, or the full request if not.
+   */
+  protected abstract getValuesNotInBatch(
+    request: TRequest,
+    batch: TRequest[],
+  ): { valuesNotInBatch: TRequest; batchContainsValues: boolean } | { valuesNotInBatch: undefined; batchContainsValues: true };
+
+  /** Produce the items from a batch that will be buffered and passed to `executeQuery`. */
+  protected abstract getIterable(batch: TRequest[]): Observable<TItem>;
```
grigasp: To me `TItem` and the comment here are too vague. What I think could help:
- finding a better name for this template arg,
- documenting template args at the class level,
- documenting the pipeline at the class level.

JonasDov: Changed, please have a look.

grigasp: Changes look good, I'd also consider changing `getIterable` name to `getQueryData` or `getQueryInput(s)`

JonasDov: Renamed to `getQueryData`

grigasp (approved):

PR #1700, opened 2026-05-28
Remove abandoned packages

saskliutas:
@aruniverse As discussed removed abandoned packages. After this I think we could also reduce nesting from `packages/itwin/some-package` -> `packages/some-package`


aruniverse (approved):
lgtm, the only one we might want to keep is the iot pkg after the recent org changes we had...

saskliutas:
> lgtm, the only one we might want to keep is the iot pkg after the recent org changes we had...

I don't think it's worth keeping it. It's been deprecated for a few year already and built on itwinjs-core 3.x. However, if for some reason source code is needed it should be possible to find it through git history.

PR #1719, opened 2026-06-15
[geo-tools] GoogleAddressProvider now handles more API variants for location requests

johannes-goltz:
**Issue**
The GoogleAddressProvider does not always resolve a given place ID to coordinates correctly. It seems the API
supports different variants of query responses but only one variant (legacy ???) is currently accpted.

see screen shoot from my debug session:
image

**Solution**
[GoogleAddressProvider.ts](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html): broadened location extraction to accept alternative response shapes (top-level [location](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), nested [result.geometry.location.lat/lng](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and other geometry variants). Kept existing behavior while adding fallbacks and validation.

### DIFF
--- a/change/@itwin-geo-tools-react-6582ffe5-6858-4b66-9546-d92b561112ee.json
+++ b/change/@itwin-geo-tools-react-6582ffe5-6858-4b66-9546-d92b561112ee.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "GoogleAddressProvider does not always resolve location requests correctly. Now handling different API variants",
+  "packageName": "@itwin/geo-tools-react",
+  "email": "Johannes.Goltz@bentley.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/itwin/geo-tools/src/GoogleAddressProvider.ts
+++ b/packages/itwin/geo-tools/src/GoogleAddressProvider.ts
@@ -73,10 +73,19 @@ export class GoogleAddressProvider implements AddressProvider {
       } });
     const json: any = await response.json();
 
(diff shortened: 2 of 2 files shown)


jason-crow:
The fix looks correct, but the inline comment could be more descriptive for future maintainers. Consider mapping each fallback arm to the specific API that produces that response shape:

```typescript
// Google Address Validation API response shape: { location: { latitude, longitude } }
// @see https://developers.google.com/maps/documentation/address-validation/reference/rest/v1/TopLevel/validateAddress#Location
const lat = json?.location?.latitude
  // Google Geocoding API response shape: { result: { geometry: { location: { lat, lng } } } }
  // @see https://developers.google.com/maps/documentation/geocoding/requests-geocoding#Results
  ?? json?.result?.geometry?.location?.lat
  // Legacy/wrapper format: { location: { geometry: { latitude, longitude } } }
  ?? json?.location?.geometry?.latitude;

const long = json?.location?.longitude
  ?? json?.result?.geometry?.location?.lng
  ?? json?.location?.geometry?.longitude;
```

This makes it clear which Google API produces each shape, so anyone encountering a new variant in the future knows exactly where to add the next fallback.

PR #1724, opened 2026-06-17
[Property grid]: don't hide non empty values

### LINKED ISSUE #1715: Property grid hide null values feature removes boolean switch.
[`PropertyGridComponent ShowHideNullValuesSettingsMenuItem`](https://github.com/iTwin/viewer-components-react/blob/9da9db520af6219c579522f709276b39083c5c98/apps/test-viewer/src/UiProvidersConfig.tsx#L228) feature hides boolean switches. 
Unless I dont understand the feature properly, a boolean is not a null value and should not be hidden. 

See in the image below, the "Add central parking" boolean toggles get's removed.
Image

Thanks.

JonasDov:
closes https://github.com/iTwin/viewer-components-react/issues/1715


grigasp (approved):

packages/itwin/property-grid/src/property-grid-react/components/FilteringPropertyGrid.tsx
```diff
@@ -130,10 +130,12 @@ export class NonEmptyValuesPropertyDataFilterer extends PropertyRecordDataFilter
       };
     }
 
+    // merged values are not null, so they should always match the filter
+    const hasValue = node.isMerged || node.value.value !== undefined;
```
saskliutas: Should we filter out empty strings?

grigasp: I would say no, but I think we should clarify on https://github.com/iTwin/viewer-components-react/issues/1715. It's a matter of how users perceive empty string values vs null values.

JonasDov: I am a bit confused now: 
- `ShowHideNullValuesSettingsMenuItem`: name indicates that the `hide`/`show` applies to null values. Since empty string is not a null value it should not be hidden?
- `ShowHideNullValuesSettingsMenuItem` component renders `Show/Hide Empty Values` text - it indicates that it hides empty string values as well.
- `Primitives.Value` has such type: `type Value = Text | String | ShortDate | Boolean | Numeric | Enum | Point | Composite | InstanceKey;` where `Primitives.Boolean` is of type `string | boolean | object | []` so boolean could be empty string (meaning false)? And I don't see a way how to determine if it's a boolean value or a string value.

Not sure if we should change the component name or the rendered text, but both changes are breaking ones, so I am hesitant to make them.

packages/itwin/property-grid/src/property-grid-react/components/FilteringPropertyGrid.tsx
```diff
@@ -130,10 +130,12 @@ export class NonEmptyValuesPropertyDataFilterer extends PropertyRecordDataFilter
...
     return {
```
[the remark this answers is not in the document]
JonasDov: I think that non primitive values are meant to not match filter and only merged primitive values should match the filter. This is due to: `isMerged` is applied when values differ (since they differ - it means that they are not empty). 
Can I resolve this @grigasp @saskliutas ?

packages/itwin/property-grid/src/property-grid-react/components/FilteringPropertyGrid.tsx
saskliutas: Current changes that were made are also breaking (empty strings are not hidden anymore). 

`PropertyRecord` has `typename` (`record.property.typename`) property that could be used determine how to interpret value. Although data is coming from presentation data provider so it should be somewhat safe to assume that boolean property won't have empty string as value.

grigasp: > * `Primitives.Boolean` is of type `string | boolean | object | []` so boolean could be empty string (meaning false)? And I don't see a way how to determine if it's a boolean value or a string value.

I wouldn't worry about this too much, IMO adding `string`, `object` and `[]` to this union type was a mistake, and I don't think anyone is ever expecting that to act as a `boolean`. I'm 100% sure none of the core property data providers set booleans as either of those types.

> Not sure if we should change the component name or the rendered text, but both changes are breaking ones, so I am hesitant to make them.

That's way I'm suggesting to first confirm the expectations. When we know what's expected, then we know how to proceed. To avoid a breaking change we can introduce a new component and deprecate the old one...

JonasDov: @grigasp 
Decided to add a new component `ShowHideEmptyValuesSettingsMenuItem` and deprecate the old one.
>That's way I'm suggesting to first confirm the expectations.

The expectation I would have is that the component name would match it's behaviour. Since the old name did not match, I think a new component with a correct name is the right approach (as you suggested).

If consumers want to have a separate `ShowHideNullValuesSettingsMenuItem` which actually filters out only null values, it could be added in the future.

@saskliutas Changed the behaviour, now empty strings are filtered out.

JonasDov: > I would say no, but I think we should clarify on #1715. It's a matter of how users perceive empty string values vs null values.

Did not see this comment, will first clarify under the issue.

JonasDov: Seems that Noel answered this question, keeping the deprecation change and hiding the empty strings.

Can this comment be resolved @grigasp @saskliutas?

saskliutas (approved):

packages/itwin/property-grid/src/property-grid-react/components/FilteringPropertyGrid.tsx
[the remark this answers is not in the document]
saskliutas: Looks like we were always hiding non primitive values so I am not sure that we need to change anything in this case.

packages/itwin/property-grid/src/property-grid-react/hooks/UseEmptyValuesSetting.tsx
```diff
@@ -0,0 +1,107 @@
...
+ * @public
+ */
+export function EmptyValuesSettingContext({ children }: PropsWithChildren<object>) {
+  const { showEmptyValues, setShowEmptyValues } = useEmptyValuesSetting();
+
+  return <emptyValuesSettingContext.Provider value={{ showEmptyValues, setShowEmptyValues }}>{children}</emptyValuesSettingContext.Provider>;
+}
+
+/**
+ * Provides context for `Show\Hide Empty Values` setting.
```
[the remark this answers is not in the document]
JonasDov: changed

packages/itwin/property-grid/src/property-grid-react/hooks/UseEmptyValuesSetting.tsx
```diff
@@ -0,0 +1,107 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { createContext, useCallback, useContext, useEffect, useState } from "react";
+import { usePreferencesContext } from "../PropertyGridPreferencesContext.js";
+import { useTelemetryContext } from "./UseTelemetryContext.js";
+
+import type { PropsWithChildren } from "react";
+
+/**
+ * Provides context for `Show\Hide Empty Values` setting.
```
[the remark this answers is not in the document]
JonasDov: changed

packages/itwin/property-grid/src/property-grid-react/hooks/UseEmptyValuesSetting.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { storage } = usePreferencesContext();
+
+  const getShowEmptyValuesPreference = useCallback(async () => {
+    const serializedValue = await storage.get(SHOW_EMPTY_KEY);
+    if (serializedValue !== undefined) {
+      return JSON.parse(serializedValue);
+    }
+    // default to `true`
+    return true;
+  }, [storage]);
```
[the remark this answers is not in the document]
JonasDov: @grigasp @saskliutas: do you agree with this suggestion? I don't want to add this, because I feel like it will stay here forever.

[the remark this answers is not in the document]
grigasp: I agree with Copilot. From the users' perspective it's the same menu item, so they would expect the state to be preserved. 

I think we should first check for the new key and use it if it's there. If not, then we should check the old key and, if found, persist the setting with the new key. That way we'll gradually transition everyone on the new key.

packages/itwin/property-grid/src/property-grid-react/hooks/UseEmptyValuesSetting.tsx
```diff
@@ -0,0 +1,107 @@
...
+  );
+
+  return {
+    showEmptyValues,
+    setShowEmptyValues: updateShowEmptyValues,
+  };
+}
+
+/** @internal */
+export const SHOW_EMPTY_KEY = "showEmptyValues";
```
grigasp: maybe prefix this with the package name to avoid collisions with other components?


```suggestion
export const SHOW_EMPTY_KEY = "@itwin/property-grid-react/showEmptyValues";
```

packages/itwin/property-grid/src/property-grid-react/hooks/UseEmptyValuesSetting.tsx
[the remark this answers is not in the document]
JonasDov: Changed

saskliutas (approved):

PR #1735, opened 2026-07-01
[Tree widget]: add option to exclude classes from trees on `next`

### LINKED ISSUE #1702: [Tree Widget]: Filter out unwanted elements in the tree?
**Package**
@itwin/tree-widget-react

In the classification tree, we have some child elements that we do not want to show to the user, because these sub-elements are not something the user deals with directly. An example is a Pipe element - it has PipingPort elements as children elements, but those are not directly edited by the user so it would be good to be able to hide them.

Is there was a way to provide a filter to the elements that are being shown in the tree, or is this something that could be considered as a future enhancement?

Thanks!

JonasDov:
closes #1702.
Added a hierarchy-config option to the Models, Categories, and Classifications trees that excludes element nodes whose class (or derived class) matches any in the provided list. 

Visibility: omitted elements still count toward visibility, toggling one via the viewport propagates to its ancestor nodes, and toggling an ancestor still adjusts the filtered-out child.

Models vs Classifications vs Categories: the models tree drops categories/models left empty by the exclusion (unless `showEmptyModels` flag is set). 
Categories and classifications trees drop only the filtered out elements (categories tree can have sub-categories and it is not clear when categories should/should not be shown).

### DIFF
--- a/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
+++ b/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Models, Categories and Classifications trees: Added option to exclude element nodes with specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to omit.",
+  "packageName": "@itwin/tree-widget-react",
+  "email": "100586436+JonasDov@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
+++ b/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Models, Categories and Classifications trees: Added an option to exclude element nodes of specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to exclude.",
+  "comment": "Models, Categories and Classifications trees: Added an option to exclude element nodes of specified classes. Use `hierarchyConfig.excludedElementClassNames` property for specifying the classes to exclude.",
   "packageName": "@itwin/tree-widget-react",
   "email": "100586436+JonasDov@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
+++ b/change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Models, Categories and Classifications trees: Added option to exclude element nodes with specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to omit.",
+  "comment": "Models, Categories and Classifications trees: Added an option to exclude element nodes of specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to exclude.",
   "packageName": "@itwin/tree-widget-react",
   "email": "100586436+JonasDov@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/packages/itwin/tree-widget/api/tree-widget-react.api.md
+++ b/packages/itwin/tree-widget/api/tree-widget-react.api.md
@@ -119,7 +119,7 @@ type CategoriesTreeHeaderButtonType = (props: CategoriesTreeHeaderButtonProps) =
 // @beta
 interface CategoriesTreeHierarchyConfiguration {
     hideSubCategories: boolean;
-    omittedElementClassNames?: Array<EC.FullClassName>;
+    excludedElementClassNames?: Array<EC.FullClassName>;
     showElements: boolean;
     showEmptyCategories: boolean;
 }
@@ -214,7 +214,7 @@ interface ClassificationsTreeComponentProps extends Pick<ClassificationsTreeProp
 
 // @alpha (undocumented)
 interface ClassificationsTreeHierarchyConfiguration {
-    omittedElementClassNames?: Array<EC.FullClassName>;
+    excludedElementClassNames?: Array<EC.FullClassName>;
     rootClassificationSystemCode: string;
 }
 
@@ -451,7 +451,7 @@ interface ModelsTreeHierarchyConfiguration {
     elementClassSpecification: EC.FullClassName;
     hideRootSubject: boolean;
     hierarchyLevelFiltering: "enable" | "disable";
-    omittedElementClassNames?: Array<EC.FullClassName>;
+    excludedElementClassNames?: Array<EC.FullClassName>;
     showEmptyModels: boolean;
 }
 
--- a/packages/itwin/tree-widget/api/tree-widget-react.api.md
+++ b/packages/itwin/tree-widget/api/tree-widget-react.api.md
@@ -118,8 +118,8 @@ type CategoriesTreeHeaderButtonType = (props: CategoriesTreeHeaderButtonProps) =
 
 // @beta
 interface CategoriesTreeHierarchyConfiguration {
-    hideSubCategories: boolean;
     excludedElementClassNames?: Array<EC.FullClassName>;
+    hideSubCategories: boolean;
     showElements: boolean;
     showEmptyCategories: boolean;
 }
@@ -449,9 +449,9 @@ type ModelsTreeHeaderButtonType = (props: ModelsTreeHeaderButtonProps) => ReactE
 interface ModelsTreeHierarchyConfiguration {
     elementClassGrouping: "enable" | "enableWithCounts" | "disable";
     elementClassSpecification: EC.FullClassName;
+    excludedElementClassNames?: Array<EC.FullClassName>;
     hideRootSubject: boolean;
     hierarchyLevelFiltering: "enable" | "disable";
-    excludedElementClassNames?: Array<EC.FullClassName>;
     showEmptyModels: boolean;
 }
 
--- a/packages/itwin/tree-widget/api/tree-widget-react.api.md
+++ b/packages/itwin/tree-widget/api/tree-widget-react.api.md
@@ -119,6 +119,7 @@ type CategoriesTreeHeaderButtonType = (props: CategoriesTreeHeaderButtonProps) =
 // @beta
 interface CategoriesTreeHierarchyConfiguration {
     hideSubCategories: boolean;
+    omittedElementClassNames?: Array<EC.FullClassName>;
     showElements: boolean;
     showEmptyCategories: boolean;
 }
@@ -213,6 +214,7 @@ interface ClassificationsTreeComponentProps extends Pick<ClassificationsTreeProp
 
 // @alpha (undocumented)
 interface ClassificationsTreeHierarchyConfiguration {
+    omittedElementClassNames?: Array<EC.FullClassName>;
     rootClassificationSystemCode: string;
 }
 
@@ -449,6 +451,7 @@ interface ModelsTreeHierarchyConfiguration {
     elementClassSpecification: EC.FullClassName;
     hideRootSubject: boolean;
     hierarchyLevelFiltering: "enable" | "disable";
+    omittedElementClassNames?: Array<EC.FullClassName>;
     showEmptyModels: boolean;
 }
 
--- a/packages/itwin/tree-widget/src/test/trees/categories-tree/CategoriesTreeDefinition.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/categories-tree/CategoriesTreeDefinition.test.ts
@@ -984,12 +984,12 @@ describe("Categories tree", () => {
           });
         });
       });
-      describe(`omittedElementClassNames in '${viewType}' view`, () => {
+      describe(`excludedElementClassNames in '${viewType}' view`, () => {
         const elementClassName: EC.FullClassName = viewType === "3d" ? "Generic.PhysicalObject" : "BisCore.DrawingGraphic";
         const modeledElementClassName: EC.FullClassName = `${TestSchema.Name}.${viewType === "3d" ? TestSchema.ModeledElement3dClassName : TestSchema.ModeledElement2dClassName}`;
         const unrelatedElementClassName: EC.FullClassName = viewType === "3d" ? "BisCore.GeometricElement2d" : "BisCore.GeometricElement3d";
         const subModeledElementBaseClassName: EC.FullClassName = "BisCore.ISubModeledElement";
-        it("does not filter out elements when they don't belong to any of the omitted classes", async () => {
+        it("does not filter out elements when they don't belong to any of the excluded classes", async () => {
           await using buildIModelResult = await buildIModel(async (imodel) =>
             withEditTxn(imodel, (txn) => {
               const physicalModel = insertElementsModel({ txn, codeValue: "model" });
@@ -1002,7 +1002,7 @@ describe("Categories tree", () => {
           const { imodelConnection, ...keys } = buildIModelResult;
           using provider = createCategoryTreeProvider(imodelConnection, viewType, {
             showElements: true,
-            omittedElementClassNames: [unrelatedElementClassName],
+            excludedElementClassNames: [unrelatedElementClassName],
           });
 
           await validateHierarchy({
@@ -1026,8 +1026,8 @@ describe("Categories tree", () => {
           });
         });
 
-        describe("root elements of omitted classes", () => {
-          it("filters out elements of omitted classes", async () => {
+        describe("root elements of excluded classes", () => {
+          it("filters out elements of excluded classes", async () => {
             await using buildIModelResult = await buildIModel(async (imodel) =>
               withEditTxn(imodel, (txn) => {
                 const physicalModel = insertElementsModel({ txn, codeValue: "model" });
@@ -1045,7 +1045,7 @@ describe("Categories tree", () => {
             const { imodelConnection, ...keys } = buildIModelResult;
             using provider = createCategoryTreeProvider(imodelConnection, viewType, {
               showElements: true,
-              omittedElementClassNames: [modeledElementClassName],
+              excludedElementClassNames: [modeledElementClassName],
             });
 
             await validateHierarchy({
@@ -1069,7 +1069,7 @@ describe("Categories tree", () => {
             });
           });
 
-          it("filters out elements of classes derived from omitted classes", async () => {
+          it("filters out elements of classes derived from excluded classes", async () => {
             await using buildIModelResult = await buildIModel(async (imodel) =>
               withEditTxn(imodel, (txn) => {
                 const physicalModel = insertElementsModel({ txn, codeValue: "model" });
@@ -1087,7 +1087,7 @@ describe("Categories tree", () => {
             const { imodelConnection, ...keys } = buildIModelResult;
             using provider = createCategoryTreeProvider(imodelConnection, viewType, {
               showElements: true,
-              omittedElementClassNames: [subModeledElementBaseClassName],
+              excludedElementClassNames: [subModeledElementBaseClassName],
             });
 
             await validateHierarchy({
@@ -1111,7 +1111,7 @@ describe("Categories tree", () => {
             });
           });
 
-          it("shows category when it contains only omitted elements", async () => {
+          it("shows category when it contains only excluded elements", async () => {
             await using buildIModelResult = await buildIModel(async (imodel) =>
               withEditTxn(imodel, (txn) => {
                 const definitionContainer = insertDefinitionContainer({ txn, codeValue: "DefinitionContainer" });
@@ -1126,7 +1126,7 @@ describe("Categories tree", () => {
             const { imodelConnection, ...keys } = buildIModelResult;
             using provider = createCategoryTreeProvider(imodelConnection, viewType, {
               showElements: true,
-              omittedElementClassNames: [elementClassName],
+              excludedElementClassNames: [elementClassName],
             });
 
             await validateHierarchy({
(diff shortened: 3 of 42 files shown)


packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BaseIdsCache.ts
```diff
@@ -139,8 +145,33 @@ export class BaseIdsCache {
...
+        return this.#elementModelCategoriesCache.getCachedData().pipe(
+          map(({ modelsCategoriesInfo }) => {
+            const result = new Set<ElementId>();
+            for (const subModelId of allSubModels) {
+              const modelInfo = modelsCategoriesInfo.get(subModelId);
+              // Keep the sub-model unless it has elements that are all of omitted classes. Empty sub-models (no entry) are kept
+              // so that behavior with an empty omitted list is identical to not passing `excludeIfOnlyOmittedClasses`.
+              if (modelInfo && modelInfo.nonOmittedCategories.size > 0) {
+                result.add(subModelId);
+              }
```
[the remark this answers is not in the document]
JonasDov: comment was incorrect, removed it

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -412,7 +428,13 @@ export class CategoriesTreeDefinition implements HierarchyDefinition {
...
             JOIN IdSet(?) modelIdSet ON this.Model.Id = modelIdSet.id
             ${elementInstanceFilterClauses.joins}
-            ${createWhereClause({ conditions: ["this.Parent.Id IS NULL", `this.Category.Id = ${modeledElementCategory}`, elementInstanceFilterClauses.where] })}
+           ${createWhereClause({
+             conditions: [
+               "this.Parent.Id IS NULL",
+               elementInstanceFilterClauses.where,
+               createOmittedClassesExclusionClause({ alias: "this", omittedClassNames: this.#omittedElementClassNames }),
+             ],
+           })}
```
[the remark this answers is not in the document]
JonasDov: added back

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/ModelsTreeIdsCache.ts
```diff
@@ -293,24 +397,32 @@ export class ModelsTreeIdsCache extends BaseIdsCacheImpl {
...
+    modelId,
+    excludeIfOnlyOmittedClasses,
+  }: {
+    modelId: Id64String;
+    excludeIfOnlyOmittedClasses?: boolean;
+  }): Observable<HierarchyNodeIdentifiersPath> {
     return getOrCreate({
       map: this.#upToModelInstanceKeyPaths,
       key: modelId,
       createFunc: () =>
```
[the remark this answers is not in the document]
JonasDov: adjusted the key

packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/BaseIdsCache.ts
```diff
@@ -139,8 +145,31 @@ export class BaseIdsCache {
...
+          map(({ modelsCategoriesInfo }) => {
+            const result = new Set<ElementId>();
+            for (const subModelId of allSubModels) {
+              const modelInfo = modelsCategoriesInfo.get(subModelId);
+              if (modelInfo && modelInfo.nonOmittedCategories.size > 0) {
+                result.add(subModelId);
+              }
+            }
+            return result;
+          }),
```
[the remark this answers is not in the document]
JonasDov: Does not matter, in this context modeled element has the same id as the sub-model. So using `ModelId` (represents sub-model), or `ElementId` (represents the modeled element) is the same. resolving

packages/itwin/tree-widget/api/tree-widget-react.api.md
```diff
@@ -119,6 +119,7 @@ type CategoriesTreeHeaderButtonType = (props: CategoriesTreeHeaderButtonProps) =
 // @beta
 interface CategoriesTreeHierarchyConfiguration {
     hideSubCategories: boolean;
+    omittedElementClassNames?: Array<EC.FullClassName>;
```
grigasp: I feel like "excluded" is a better term here

change/@itwin-tree-widget-react-b60d095b-d9cf-4293-9960-db4487f47da5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Models, Categories and Classifications trees: Added option to exclude element nodes with specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to omit.",
```
grigasp:
```suggestion
  "comment": "Models, Categories and Classifications trees: Added an option to exclude element nodes of specified classes. Use `hierarchyConfig.omittedElementClassNames` property for specifying the classes to exclude.",
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -412,7 +428,14 @@ export class CategoriesTreeDefinition implements HierarchyDefinition {
             FROM ${elementInstanceFilterClauses.from} this
             JOIN IdSet(?) modelIdSet ON this.Model.Id = modelIdSet.id
             ${elementInstanceFilterClauses.joins}
-            ${createWhereClause({ conditions: ["this.Parent.Id IS NULL", `this.Category.Id = ${modeledElementCategory}`, elementInstanceFilterClauses.where] })}
+           ${createWhereClause({
```
grigasp: indentation

grigasp:
> Categories and classifications trees drop only the filtered out elements (categories tree can have sub-categories and it is not clear when categories should/should not be shown).

Why is it not clear? I think it should work the following way:
- Categories tree: unless `showEmptyCategories` is set, don't show categories that have no child nodes (element or sub-category).
- Classifications tree: shows categories no matter if they have elements or not, and have no option to configure that. So just leave as-is.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/internal/visibility/CategoriesTreeVisibilityHandler.ts
```diff
@@ -173,7 +173,7 @@ export class CategoriesTreeVisibilityHandler implements Disposable, TreeSpecific
       modelId: node.extendedData.modelId,
       categoryId: node.extendedData.categoryId,
       parentElementsPath: node.extendedData.parentElementsPath,
-      computeOnlyOwnStatus: node.children ? undefined : true,
+      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() ? undefined : node.children ? undefined : true,
```
grigasp:
```suggestion
      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() || node.children ? undefined : true,
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -114,6 +115,8 @@ export interface CategoriesTreeHierarchyConfiguration {
   showElements: boolean;
   /** Should categories without elements be shown. Defaults to `false`. */
   showEmptyCategories: boolean;
+  /** Classes to omit from the hierarchy. Element nodes whose class derive from the omitted classes will not be shown in the hierarchy. Defaults to `undefined`. */
+  omittedElementClassNames?: Array<EC.FullClassName>;
```
grigasp:
```suggestion
  /** Element classes to exclude from the hierarchy. Elements, whose class is or derives from one of the classes in this list, are not loaded into the hierarchy. Defaults to `[]`. */
  omittedElementClassNames?: Array<EC.FullClassName>;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -866,18 +910,18 @@ export class CategoriesTreeDefinition implements HierarchyDefinition {
...
-                `this.ECInstanceId IN (
-                  SELECT DISTINCT ce.Category.Id
-                  FROM ${this.#categoryElementClass} ce
-                  JOIN IdSet(?) parentIdSet ON ce.Parent.Id = parentIdSet.id
-                  ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES
-                )`,
-                categoryInstanceFilterClauses.where,
-              ],
-            })}
+            WHERE
```
grigasp: why not use `createWhereClause` here?

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/internal/visibility/ClassificationsTreeVisibilityHandler.ts
```diff
@@ -133,7 +133,7 @@ export class ClassificationsTreeVisibilityHandler implements Disposable, TreeSpe
       modelId: node.extendedData.modelId,
       categoryId: node.extendedData.categoryId,
       parentElementsPath: node.extendedData.parentElementsPath,
-      computeOnlyOwnStatus: node.children ? undefined : true,
+      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() ? undefined : node.children ? undefined : true,
```
grigasp:
```suggestion
      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() || node.children ? undefined : true,
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTreeDefinition.ts
```diff
@@ -72,6 +86,8 @@ export interface ClassificationsTreeHierarchyConfiguration {
    * root `ClassificationSystem`.
    */
   rootClassificationSystemCode: string;
+  /** Classes to omit from the hierarchy. Element nodes whose class derive from the omitted classes will not be shown in the hierarchy. Defaults to `undefined`. */
+  omittedElementClassNames?: Array<EC.FullClassName>;
```
grigasp:
```suggestion
  /** Element classes to exclude from the hierarchy. Elements, whose class is or derives from one of the classes in this list, are not loaded into the hierarchy. Defaults to `[]`. */  
  omittedElementClassNames?: Array<EC.FullClassName>;
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/internal/visibility/ModelsTreeVisibilityHandler.ts
```diff
@@ -193,7 +193,7 @@ export class ModelsTreeVisibilityHandler implements Disposable, TreeSpecificVisi
       modelId: node.extendedData.modelId,
       categoryId: node.extendedData.categoryId,
       parentElementsPath: node.extendedData.parentElementsPath,
-      computeOnlyOwnStatus: node.children ? undefined : true,
+      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() ? undefined : node.children ? undefined : true,
```
grigasp:
```suggestion
      computeOnlyOwnStatus: this.#props.idsCache.canHaveHiddenChildren() || node.children ? undefined : true,
```

packages/itwin/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -84,6 +85,8 @@ export interface ModelsTreeHierarchyConfiguration {
   hideRootSubject: boolean;
   /** Should hierarchy level be filterable. Defaults to `enable` */
   hierarchyLevelFiltering: "enable" | "disable";
+  /** Classes to omit from the hierarchy. Element nodes whose class derive from the omitted classes will not be shown in the hierarchy. Defaults to `undefined`. */
+  omittedElementClassNames?: Array<EC.FullClassName>;
```
grigasp:
```suggestion
  /** Element classes to exclude from the hierarchy. Elements, whose class is or derives from one of the classes in this list, are not loaded into the hierarchy. Defaults to `[]`. */
  omittedElementClassNames?: Array<EC.FullClassName>;
```

packages/itwin/tree-widget/src/test/trees/categories-tree/CategoriesTreeDefinition.test.ts
```diff
@@ -983,6 +984,430 @@ describe("Categories tree", () => {
...
+            });
+
+            await validateHierarchy({
+              provider,
+              expect: [
+                NodeValidators.createForInstanceNode({
+                  instanceKeys: [keys.definitionContainer],
+                  children: [
+                    NodeValidators.createForInstanceNode({
+                      instanceKeys: [keys.category],
```
grigasp, line 1139: `showEmptyCategories` is not set, so I would expect this category to not be displayed in the hierarchy

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeDefinition.test.ts
```diff
@@ -1684,6 +1684,650 @@ describe("Models tree", () => {
...
+                insertPhysicalElement({
+                  txn,
+                  userLabel: `omitted element 2`,
+                  modelId: model.id,
+                  categoryId: omittedCategory.id,
+                });
+                const emptyModel = insertPhysicalModelWithPartition({ txn, codeValue: `empty model`, partitionParentId: childSubject.id });
+                insertPhysicalElement({
+                  txn,
+                  userLabel: `empty element 1`,
```
grigasp: maybe this should be called "omitted" or "excluded" rather than "empty"?

grigasp: same for the one below

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeDefinition.test.ts
```diff
@@ -1684,6 +1684,650 @@ describe("Models tree", () => {
...
+                    }),
+                  ],
+                }),
+              ],
+            });
+          });
+        });
+
+        describe("child elements of omitted classes", () => {
+          it("sets hasChildren to false parent contains only omitted elements", async () => {
```
grigasp:
```suggestion
          it("sets hasChildren to false when parent contains only omitted elements", async () => {
```

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeDefinition.test.ts
```diff
@@ -1684,6 +1684,650 @@ describe("Models tree", () => {
...
+                        }),
+                      ],
+                    }),
+                  ],
+                }),
+              ],
+            });
+          });
+
+          it("sets hasChildren to false parent modeled element contains only omitted elements", async () => {
```
grigasp:
```suggestion
          it("sets hasChildren to false when parent modeled element contains only omitted elements", async () => {
```

grigasp (commented):
Could we add a perf test for each hierarchy, where we exclude some class? Just to help us visualize the perf difference between regular case and the "excluded" one. I guess we could exclude some class whose instances aren't even in the iModel, just to make sure we hit these more expensive paths.

packages/itwin/tree-widget/src/tree-widget-react/components/trees/classifications-tree/ClassificationsTreeDefinition.ts
JonasDov: changed

packages/itwin/tree-widget/src/test/trees/models-tree/ModelsTreeDefinition.test.ts
JonasDov: Renamed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
JonasDov: changed

packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeDefinition.ts
JonasDov: fixed

packages/itwin/tree-widget/api/tree-widget-react.api.md
JonasDov: changed

JonasDov:
> Could we add a perf test for each hierarchy, where we exclude some class? Just to help us visualize the perf difference between regular case and the "excluded" one. I guess we could exclude some class whose instances aren't even in the iModel, just to make sure we hit these more expensive paths.

Added perf test for each tree.

JonasDov:
> Why is it not clear?

I am not completely sure what should be the correct approach. A few things to consider:
1. Visibility controls should remain intact. So if we don't show any ancestors of elements which are excluded, there won't be a way to change their visibility in the tree. E.g. model -> category -> excludedElement. If we don't show anything in the hierarchy, there could be a scenario where excludedElement is visible in the viewport but you can't turn it off using the trees.
2. With categories tree in particular it is not entirely clear what is the best approach. This is because categories tree has multiple configuration options which overlap with `excludedElementClassNames`:
```ts
hideSubCategories: boolean;
/** Should Elements be shown. Defaults to `false` */
showElements: boolean;
/** Should categories without elements be shown. Defaults to `false`. */
showEmptyCategories: boolean;
/** Element classes to exclude from the hierarchy. Elements, whose class is or derives from one of the classes in this list, are not loaded into the hierarchy. Defaults to `[]`. */
excludedElementClassNames?: Array<EC.FullClassName>;
```
lets take these categories:
- categoryA
  - elementA1
  - excluded elementA2
  - subCategoryA1
  - subCategoryA2
- categoryB
  - excluded elementB1
  - subCategoryB1
  - subCategoryB2
- categoryC
  - excluded elementC1
  - subCategoryC1
- categoryD
  - subCategoryD1
  - subCategoryD2
  
`showElements` flag determines if `elementA1`, `excluded elementA2`, `excluded elementB1`, `excluded elementC1` are shown or not.
`hideSubCategories` flag determines `subCategoryA1`, `subCategoryA2`, `subCategoryB1`, `subCategoryB2` should be hidden.
`showEmptyCategories` determines if `categoryD`  is shown.
categories tree hierarchy with the following config:
```ts
{ showElements: false, hideSubCategories: false, showEmptyCategories: false, excludedElementClassNames: undefined }
```
- categoryA
  - subCategoryA1
  - subCategoryA2
- categoryB
  - subCategoryB1
  - subCategoryB2
- categoryC
As I understand your suggestion hierarchy with the following config:
```ts
{ showElements: false, hideSubCategories: false, showEmptyCategories: false, excludedElementClassNames: [...] }
```
- categoryA
  - subCategoryA1
  - subCategoryA2
- categoryB
  - subCategoryB1
  - subCategoryB2

so adding `excludedElementClassNames` made `categoryC` dissapear while `categoryB` is still there. Now if `hideSubCategories` is set to false, then as I understand `categoryB` should also disappear in such case?

grigasp:
> > Could we add a perf test for each hierarchy, where we exclude some class? Just to help us visualize the perf difference between regular case and the "excluded" one. I guess we could exclude some class whose instances aren't even in the iModel, just to make sure we hit these more expensive paths.
> 
> Added perf test for each tree.

Interesting, I'm surprised that performance is nearly the same as without excluded classes. Is that expected?

grigasp:
> I am not completely sure what should be the correct approach. A few things to concsder: <...>

Sounds like we understand the "empty category" term differently - to me it's "a category node that has no children (doesn't matter if they're elements or sub-categories), to you it's "a category node that has no elements". Is that right?

I wouldn't worry too much about the 1st point - I don't think the scenario of iModel having only excluded elements, thus making the hierarchy empty, is realistic. Maybe this situation is worth documenting next to the `excludedElementClassNames` prop, but not more than that.

JonasDov:
> Interesting, I'm surprised that performance is nearly the same as without excluded classes. Is that expected?

In terms of visibility handling yes, it is expected to be the same. The only part which might/could differ is the `collect nodes` part since the queries might be slightly slower when excluded classes are present. I am guessing: since our test iModels don't have that many elements, the performance difference is very small.

JonasDov:
> Sounds like we understand the "empty category" term differently - to me it's "a category node that has no children (doesn't matter if they're elements or sub-categories), to you it's "a category node that has no elements". Is that right?
> 
> I wouldn't worry too much about the 1st point - I don't think the scenario of iModel having only excluded elements, thus making the hierarchy empty, is realistic. Maybe this situation is worth documenting next to the `excludedElementClassNames` prop, but not more than that.

The `showEmptyCategories` name might be misleading. Originally this feature was added to show categories which don't have any elements (comment on this property explains that): 
```ts
/** Should categories without elements be shown. Defaults to `false`. */
showEmptyCategories: boolean;
```
I will create a separate PR to adjust the names of `hierarchyConfig` properties, since they are misleading.

I decided to keep the ancestor nodes of the excluded elements present in the hierarchy:
- models tree: models which have only excluded elements will be shown (user won't be able to expand this node).
- categories tree: categories which contain only excluded elements will be shown.
- classifications tree: classifications which contain only excluded elements will be shown. 

I think this is the desired behaviour. If users see a need to hide such nodes, then we can add such feature in the future. For now this approach seems to be simpler to implement.

packages/itwin/tree-widget/src/test/trees/categories-tree/CategoriesTreeDefinition.test.ts
JonasDov: as explained in the comment below, keeping the implementation as is.

grigasp (approved):

PR #1739, opened 2026-07-15
[Tree widget]: add tests for sub-models on `next`

### LINKED ISSUE #1553: [Tree widget]: Visibility of hierarchies containing sub-models behaves incorrectly with search paths
Noticed a bug in categories tree while adding searched hierarchy tests.
With the following hierarchy:
- category1
  - element1
     - category2
        -  element1_1 (model is modeled element)
        -  element1_2 (model is modeled element)

and search paths: [[category1, element1, element1SubModel, category2, element1_1]],
switching `category1` visibility, results in it having `partial` visibility, even though it should be `visible`. 
This happens because when checking `category1` elements (in this case `element1`) visibility, visibility of models that are elements is also checked (and while checking these sub-models visibility handler does not take into account search paths).
It should be enough to fix this by:
1. Modify `BaseIdsCache.getCategories` to accept search paths, and then filter out categories based on search paths
2. Modify `BaseIdsCache.getElementsCount` to accept search paths, and then return correct count for categories based on it's search paths.
 
It should be noted that additional tests for searched hierarchy should be added that check searched and non searched categories that are under sub-models.

JonasDov:
closes https://github.com/iTwin/viewer-components-react/issues/1553

This was fixed while implementing child categories changes. Added tests and removed `.skip`.

### DIFF
--- a/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
+++ b/packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
@@ -5147,8 +5147,7 @@ describe("CategoriesTreeVisibilityHandler", () => {
       });
     });
 
-    // TODO: Enable with https://github.com/iTwin/viewer-components-react/issues/1553
-    describe.skip("category with modeled elements hierarchy", () => {
+    describe("category with modeled elements hierarchy", () => {
       let createIModelResult: Awaited<ReturnType<typeof createIModel>>;
       let visibilityTestData: Awaited<ReturnType<typeof createFilteredVisibilityTestData>>;
       async function createIModel() {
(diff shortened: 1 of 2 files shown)


packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
```diff
@@ -5429,12 +5463,17 @@ describe("CategoriesTreeVisibilityHandler", () => {
...
+
+            [otherSubModelCategory.id]: "hidden",
+              [defaultSubCategoryOfOtherSubModelCategory.id]: "hidden",
 
             [siblingCategory.id]: "hidden",
               [siblingElement.id]: "hidden",
               [defaultSiblingSubCategory.id]: "hidden",
 
-            [subModelCategory.id]: "partial",
+            [subModelCategory.id]: "hidden",
```
grigasp, line 5476: just to confirm, this is not "partial" because the above visibility change made a per-model category visibility change and when determining root catregories' visibility we're only looking at selector?

grigasp (approved):

packages/itwin/tree-widget/src/test/trees/categories-tree/internal/CategoriesTreeVisibilityHandler.test.ts
JonasDov: You are correct.

PR #1746, opened 2026-07-24
[Tree widget]: review public and beta tags on next

JonasDov:
- removed `export` where they were not needed.
- Removed `public` tag from non public apis.
- Removed functions which did nothing.


packages/itwin/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
```diff
@@ -204,3 +212,41 @@ function useAvailableModels(viewport: TreeWidgetViewport): Array<ModelId> {
...
+
+/**
+ * Invert display of all given categories.
+ * Categories are inverted like this:
+ * - If category is visible, it will be hidden.
+ * - If category is hidden, it will be visible.
+ * - If category is partially visible, it will be fully visible.
+ * @internal
+ */
+export async function invertAllCategories(categories: CategoryInfo[], viewport: TreeWidgetViewport) {
```
saskliutas: Whis is this exported? Is it used somewhere else? If this is the case why not placing it in VisibilityUtils?

JonasDov: moved into `VisibilityUtils`

saskliutas (approved):

PR #1749, opened 2026-07-27
Reduce nesting

saskliutas:
Reduced unnecessary nesting by removing intermediate `itwin` folder: `packages/itwin/some-package` -> `packages/some-package`


saskliutas:
@aruniverse This is repo wide change let me know if you are ok with it and we can by pass approvals to avoid distracting all code owners.

PR #1751, opened 2026-07-28
[Tree widget]: header buttons adjustments on `next`

JonasDov:
- Changed Models and Categories tree header buttons to be more consistent.
- Those actions now use `BufferingViewport`.
- PlanProjection models are now detected and corretly used in view2d/view3d buttons. This fixes the 2D toggle being always disabled.
- Updated 2D/3D toggle state handling to react to viewport display changes.
- Expanded visibility utility test coverage.


packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
```diff
@@ -120,77 +119,248 @@ export function getCategoryVisibilityFromAlwaysAndNeverDrawnElementsImpl(props:
...
+          categoriesToEnable.set(categoryId, subCategoryIds);
+          continue;
+        }
+        // Check if category is in partial state
+        if (subCategoryIds?.some((subCategory) => !bufferingViewport.viewsSubCategory(subCategory))) {
+          categoriesToEnable.set(categoryId, subCategoryIds);
+        } else {
+          categoriesToDisable.set(categoryId, subCategoryIds);
+        }
+      }
```
JonasDov: @saskliutas I would like to use visibility handlers in those buttons to determine visibility of models/categories and then call the `visibilityHandler.changeVisibility` with on/off instead.  At the moment we don't really know if the nodes are in `partial`/`visible`/`hidden` states.  This would allow us to "correctly" change visibility, but buttons would work slower.

packages/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeButtons.tsx
```diff
@@ -154,47 +172,76 @@ export function HideAllButton(props: ModelsTreeHeaderButtonProps) {
...
+  const onClick = useCallback(() => {
+    onFeatureUsed?.("models-tree-view2d");
+    cancelChangesInProgress.next();
+    const modelsToTurnOn = models2d.filter((id) => !viewport.viewsModel(id));
+    if (modelsToTurnOn.length > 0) {
+      viewport.changeModelDisplay({ modelIds: modelsToTurnOn, display: true });
+    } else {
+      // All models are already visible, so turn them all off
+      viewport.changeModelDisplay({ modelIds: models2d, display: false });
+    }
```
JonasDov: @saskliutas Is it correct to turn on all of the PlanProjection models if at least one of them is not visible and to turn them of if they are all visible? This is how it previously worked, just want to make sure that this is intended.

Also, when clicking 2d/3d buttons should we turn off models in the other category (2D button would turn off non PlanProjection models, and 3D button would turn off PlanProjection models)?

saskliutas: It makes sense to turn on all when some are hidden and turn off when all are visible. 

I don't think clicking 2d button should affect 3d models. It looks like this is how it worked before.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
saskliutas: I don't think this would work. I think it would freeze up for too long. We already sometimes encounter freezing handling visibility change of single node but this would require to perform this action on all categories/models.

JonasDov: I guess changing hundreds/thousands of categories/models using visibility handler would not be performant. But we could use the handler to determine visibility of those nodes and then change visibility accordingly. Right now we can have cases where a node which is partial would become `hidden`.

saskliutas: But we don't have nodes in header buttons. That would also require to check all nodes that are loaded, wouldn't it?

JonasDov: Well, we would have to create a node for each categoryId/modelId which are provided to the buttons. They would look like:
```ts
const node: HierarchyNode = {
    key: {
      type: "instances",
      instanceKeys: [{ className: categoryClass/modelClass, id: categoryId/modelId }],
    },
    children: true,
    label: "",
    parentKeys: [],
    extendedData: {
      type: "category"/"model",
      // For categories in category tree
      modelIds: [],
      hasSubCategories: true
    },
  }
  ```

Comment (saskliutas) [reply packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts]: This looks like a hacky solution. Also whole API will probably need to change significantly to lift visibilityHandler to component that renders buttons and tree itself. Or can we create new visibilityHandler inside button on each click?

Comment (JonasDov) [reply packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts]: I guess we could create a new instance, but I see how this becomes a hacky solution. Resolving the comment.

File: packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts [hunk 3]
```diff
@@ -120,77 +119,248 @@ export function getCategoryVisibilityFromAlwaysAndNeverDrawnElementsImpl(props:
...
+          const subCategoryIds = categoryInfos.get(categoryId);
+          for (const subCategoryId of subCategoryIds ?? []) {
+            viewport.changeSubCategoryDisplay({ subCategoryId, display });
+          }
         }
       }),
-    ),
-  );
+      takeLast(1),
+    );
```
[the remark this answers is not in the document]
JonasDov: `takeLast(1)` completes without emitting when the upstream is empty; it does not throw. Since `commitViewportChange` subscribes through `toVoidPromise` it works as expected. Added tests for empty input. Resolving

packages/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeButtons.tsx
```diff
@@ -154,47 +172,76 @@ export function HideAllButton(props: ModelsTreeHeaderButtonProps) {
...
+  }, [viewport, cancelChangesInProgress, models2d, onFeatureUsed]);
```
saskliutas: What happens if we call `changeModelDisplay` with `display: true` for models that are already visible? If it's no-op is there a reason to calculate hidden models subset? Can we just call display: true/false with all models each time? We could use `is2dToggleActive` to determine if we need to hide/show all models.

packages/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
```diff
@@ -212,3 +259,28 @@ function useAvailableModels(viewport: TreeWidgetViewport): Array<ModelId> {
...
+  baseIdsCache,
+  cancel,
+}: {
+  categoriesInfo: CategoryInfo[];
+  baseIdsCache: BaseIdsCache;
+  cancel: Observable<void>;
+}): Promise<CategoryInfosMap | undefined> {
+  return firstValueFrom(
+    defer(async () => {
+      const categoriesWithSubCategories = await Promise.all(
```
saskliutas: Can we use `from(categoriesInfo).mergeMap(() => {...})` instead of `Promise.all`?

packages/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
```diff
@@ -90,21 +92,37 @@ export type CategoriesTreeHeaderButtonType = (props: CategoriesTreeHeaderButtonP
...
       onClick={() => {
-        // cspell:disable-next-line
-        props.onFeatureUsed?.(`categories-tree-showall`);
-        cancelChangesInProgress.next();
-        void showAll({
-          models: props.models,
-          viewport: props.viewport,
-          categories: props.categories.map((category) => category.categoryId),
-        }).catch(() => {});
+        void onClick();
```
saskliutas: No need to wrap in arrow function.
```suggestion
      onClick={onClick}
```

Same in other buttons. Also there is no need to wrap `onClick` in `useCallback`.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
```diff
@@ -97,6 +98,36 @@ export class ElementModelCategoriesCache {
     );
   }
 
+  private queryPlanProjectionModels(): Observable<ModelId> {
```
saskliutas: Do we need this separate query? Could we just include `IsPlanProjection` flag where we are querying models? It looks like we are querying all models anyway.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
```diff
@@ -120,77 +119,248 @@ export function getCategoryVisibilityFromAlwaysAndNeverDrawnElementsImpl(props:
...
-    }
-    viewport.setPerModelCategoryOverride({ modelIds: modelsContainingOverrides, categoryIds: bufferedCategories, override: "none" });
-  };
-  const disableSubCategories = async (bufferedCategories: Id64Array) => {
-    // changeCategoryDisplay only enables subcategories, it does not disabled them. So we must do that ourselves.
-    const categoryInfo = await viewport.iModel.categories.getCategoryInfo(bufferedCategories);
-    for (const info of categoryInfo.values()) {
-      for (const value of info.subCategories.values()) {
-        viewport.changeSubCategoryDisplay({ subCategoryId: value.id, display: false });
+export function modifyCategoryDisplay({
```
saskliutas: Maybe `changeCategoryDisplay` to make it more consistent with `viewport` API?

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
```diff
@@ -120,77 +119,248 @@ export function getCategoryVisibilityFromAlwaysAndNeverDrawnElementsImpl(props:
...
+      }),
+    ),
+  );
+}
+
 /**
- * Changes category display in the viewport.
+ * Turns display of given categories and their sub-categories on or off, clearing any per-model category overrides.
+ *
+ * Categories are processed in batches, releasing the main thread in between, so the returned observable may emit over multiple frames.
```
saskliutas, line 126: Why this function does not use `commitViewportChange`?

packages/tree-widget/src/test/trees/common/internal/VisibilityUtils.test.ts
```diff
@@ -71,34 +90,307 @@ describe("VisibilityUtils", () => {
...
+        cancel: new Subject<void>(),
+      });
+
+      expect(viewport.clearAlwaysDrawn).toHaveBeenCalled();
+      expect(viewport.clearNeverDrawn).toHaveBeenCalled();
+      expect(viewport.changeModelDisplay).toHaveBeenCalledWith({ modelIds: ["0x10"], display: true });
+      expect(viewport.changeCategoryDisplay).not.toHaveBeenCalled();
+    });
+
+    it("applies changes when it is not cancelled", async () => {
```
saskliutas: It this test needed? It looks like this is covered by previous tests?

packages/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeButtons.tsx
JonasDov: It is a no-op. Changed to call based on your suggestion.

packages/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
JonasDov: Changed

packages/tree-widget/src/tree-widget-react/components/trees/categories-tree/CategoriesTreeButtons.tsx
JonasDov: Changed

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
JonasDov: Well, my thought was if it had the same name as the `viewport` api, then it might be confusing why the function is even needed. But decided to be consistent and changed to `changeCategoryDisplay` as you suggested.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
JonasDov: `queryElementModelCategories` uses `BisCore.Model` class to see which models relate to either `GeometricElement3d` or `GeometricElement2d`. The problem is: we can't use `GeometricModel2d/3d` in that query, since `IsPlanProjection` flag exists only on `GeometricModel3d` class.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
JonasDov: Because it is not used by any of the buttons directly. All of them call either ShowAll/HideAll/InvertAll functions which use the `commitViewportChange`. This function is exported only for testing purposes.

packages/tree-widget/src/test/trees/common/internal/VisibilityUtils.test.ts
JonasDov: This test uses enough categories to trigger the release main thread functionality. It verifies that buffered changes are committed after asynchronous completion. Renamed the test.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
saskliutas: Could we use instance access like `$->IsPlanProjection`?

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/VisibilityUtils.ts
saskliutas: Do we need categories batching and main thread releasing if this is always called inside `commitViewportChange`?

JonasDov: Removed it.

packages/tree-widget/src/tree-widget-react/components/trees/common/internal/caches/ElementModelCategoriesCache.ts
JonasDov: changed

JonasDov:
@saskliutas I have made changes to when buttons are disabled and their tooltips. Previously we would allow clicking header buttons even when models/categories arrays were empty, now they are disabled. Tooltip of disabled buttons now shows `Loading` when data is not yet loaded. What do you think?

saskliutas:
> @saskliutas I have made changes to when buttons are disabled and their tooltips. Previously we would allow clicking header buttons even when models/categories arrays were empty, now they are disabled. Tooltip of disabled buttons now shows `Loading` when data is not yet loaded. What do you think?

I don't like disabled buttons with loading tooltip. Can we leave it as is? It will be noop until data is loaded? Ideally we would show skeleton buttons while loading but I think active noop button should be fine because if I understand correctly in that case where buttons are loading data tree will be loading too?

JonasDov:
> but I think active noop button should be fine because if I understand correctly in that case where buttons are loading data tree will be loading too?

Well, not exactly. Tree nodes can be loaded without loading that data in. Models/Categories trees don't use that data to load the root hierarchy level, they use it to validate visibility. So, there is a point (especially noticeable on larger imodels) where root nodes are loaded, but `element-model-categories` query has not yet finished. So I see root nodes, I can click header buttons, but nothing happens. That's why I wanted to add the `disabled` state for such scenarios.

JonasDov:
> I don't like disabled buttons with loading tooltip. Can we leave it as is? It will be noop until data is loaded? Ideally we would show skeleton buttons while loading but I think active noop button should be fine because if I understand correctly in that case where buttons are loading data tree will be loading too?

Added skeleton buttons which are displayed when models/categories arrays are empty AND cache has not yet loaded items. 
Buttons are now disabled if those arrays are empty. Here is how it looks:
image

saskliutas (approved):

PR #1752, opened 2026-07-28
Archive unmaintained packages

saskliutas:
"archived" packages that are not maintained:
- `@itwin/grouping-mapping-widget`
- `@itwin/ec3-widget-react`
- `@itwin/one-click-lca-react`
- `@itwin/reports-config-widget-react`

 These packages were not removed just moved under `archive` folder so they can be easily restored if needed. They are also excluded from build/release system. This should improve build times for maintainers of other packages housed in this repo.


JonasDov (approved):

aruniverse:
@arnob-mallick fyi

arnob-mallick:
I think archiving is fine. I'm assuming we can still manually publish if need be once in a blue moon?

aruniverse:
> I think archiving is fine. I'm assuming we can still manually publish if need be once in a blue moon?

We would need to move the pkgs out of archive, but yes we could

aruniverse (approved):

saskliutas:
> > I think archiving is fine. I'm assuming we can still manually publish if need be once in a blue moon?
> 
> We would need to move the pkgs out of archive, but yes we could

Not necessarily. It should be enough to update `beachball` and `pnpm-workspace` to include path pointing to `archive/package-to-release`. After release this could be reverted if no more releases are needed.

If packages become actively maintained then they should be moved back out of archive.

PR #1754, opened 2026-07-28
[Tree-widget]: update hierarchies to 16

MartynasStrazdas:


packages/tree-widget/pnpm-lock.yaml
```diff
@@ -683,6 +714,12 @@ packages:
   '@itwin/core-bentley@5.9.1':
     resolution: {integrity: sha512-avysdRO0rT1mkYGPh8g3cxFZb/fWpYyo2zBxrsQcFDUpkTt4cF6/UkdAKylQ8h4xd19/AKucR2Vh3K7Eeez++Q==}
```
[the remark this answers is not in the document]
MartynasStrazdas: it should be fine

packages/tree-widget/src/tree-widget-react/components/trees/imodel-content-tree/IModelContentTreeDefinition.ts
```diff
@@ -241,7 +241,7 @@ export class IModelContentTreeDefinition implements HierarchyDefinition {
             JOIN ${modelFilterClauses.from} this ON this.ECInstanceId = model.ECInstanceId
             ${modelFilterClauses.joins}
 
-            ${createWhereClause({ conditions: [modelFilterClauses.where && `model.${NodeSelectClauseColumnNames.HideNodeInHierarchy} OR ${modelFilterClauses.where}`] })}
+            ${createWhereClause({ conditions: [modelFilterClauses.where && `model.HideNodeInHierarchy OR ${modelFilterClauses.where}`] })}
```
saskliutas: I think we need proper solution instead of magic strings. Probably `NodeSelectClauseColumnNames` should be exported. @JonasDov any thoughts?

packages/tree-widget/src/tree-widget-react/components/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -498,7 +498,7 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
             ) model
             JOIN ${modelFilterClauses.from} this ON this.ECInstanceId = model.ECInstanceId
             ${modelFilterClauses.joins}
-            ${createWhereClause({ conditions: [modelFilterClauses.where && `model.${NodeSelectClauseColumnNames.HideNodeInHierarchy} OR ${modelFilterClauses.where}`] })}
+            ${createWhereClause({ conditions: [modelFilterClauses.where && `model.HideNodeInHierarchy OR ${modelFilterClauses.where}`] })}
```
saskliutas: Same as previous should avoid magic strings.

packages/tree-widget/src/tree-widget-react/components/trees/imodel-content-tree/IModelContentTreeDefinition.ts
MartynasStrazdas: We could just have our own internal version of `NodeSelectClauseColumnNames`, which would include what we need.

saskliutas: Where would this internal NodeSelectClauseColumnNames would live?

JonasDov: Did not think of this when reviewing pr that removed the `NodeSelectClauseColumnNames`. How it was replaced in models tree in that pr: https://github.com/iTwin/presentation/pull/1311/changes#diff-09a781328fb84417f0b995ee49c5df8de3a7a3ad2c10e1d6f123cd95a88d0a1bR325.
I think we have two approaches here:
- Either make the `NodeSelectClauseColumnNames` exported in hierarchies package. I don't see why not (that interface already has a `public` tag), but not sure why it was removed in the original pr.
- create a separate file with a copy of `NodeSelectClauseColumnNames`. Can go with this option also, since that api has not changed in over 2 years and I don't think that it will any time soon.

Prefer option 1 but either one is ok with me.

saskliutas: I think we can take same approach as in PR you linked. That would remove magic string and would not require to export `NodeSelectClauseColumnNames`.

MartynasStrazdas: Changed, please review iModelsTree definition if it makes sence

JonasDov: LGTM

saskliutas (approved):

packages/tree-widget/src/e2e-tests/CategoriesTree.test.ts-snapshots/Categories-tree-node-with-active-filtering---information-message-2-chromium-linux.png
grigasp, line 1: The issue message now has an extra unnecessary space between node label and `.` 😢

JonasDov: This is the same way how it is in `hierarchies-react` package https://github.com/iTwin/presentation/pull/1425/changes#diff-3b7c3d147a39be727569a2bea2138fc2bc924a7f56e516f6a9ca4adbd957e038.
Do you think we should remove the `" "` from [MessageWithLink](https://github.com/iTwin/presentation/blob/f82b3203fa4d3ee2ae3eabcc91615bf7e9d2beac/packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/ErrorItemRenderer.tsx#L157)?

grigasp: Yes, it looks like the whole line 157 should be removed there 🤔

JonasDov: https://github.com/iTwin/presentation/pull/1475#discussion_r3749268809

PR #1762, opened 2026-08-04
Migrate `change` files ->  `changeset` on `next`

JonasDov:
Moved all files under `.changeset` folder and renamed to have `.md` extension. 
Removed property grid change file.
After merge with master README needed to be reformatted by `check-extractions`.
Added `changeset` folder to prettier ignore list.

In a follow up PR: 
- Will remove added prettier ignore.
- Update changeset files to match the format of changesets.
- Add what is needed for releasing `tree-widget/next`.


saskliutas:
We need to enter changesets pre mode and mark all change files as handled so there won't be new release PR for already release changes

JonasDov:
> We need to enter changesets pre mode and mark all change files as handled so there won't be new release PR for already release changes

Yes, will do that in a separate pr. This pr only renames the change files and moves them to the appropriate location.

saskliutas (approved):

PR #1787, opened 2026-08-20
[Tree widget]: add migration guide on `next`

### LINKED ISSUE #1737: Tree widget: Docs for v4
Need docs related to v4 vs v3:
- What changed: enhancements made, bugs fixed, visual changes, etc.
- Breaking changes & migration guide.

JonasDov:
Part of #1737


packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,841 @@
...
+    // iTwin.js packages with the same version range
+    "@itwin/core-bentley": "^5.8.0",
+    "@itwin/core-common": "^5.8.0",
+    "@itwin/core-frontend": "^5.8.0",
+    "@itwin/ecschema-metadata": "^5.8.0",
+    "@itwin/presentation-common": "^5.8.0",
+    "@itwin/presentation-frontend": "^5.8.0",
+
+    // Other iTwin.js packages
+    "@itwin/appui-abstract": "^5.8.0",
```
grigasp: `@itwin/appui-abstract` should go up

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,841 @@
+# Migrating from tree-widget v3 to v4
+
+Version 4 moves tree-widget to the current iTwin.js hierarchy and UI packages. This is a breaking release: update the application's dependencies and replace v3 initialization and rendering APIs before upgrading individual trees.
```
grigasp: This is very unclear, sounds like the major release merely bumps dependency versions.

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,841 @@
...
+
+`LOCALIZATION_NAMESPACES` is exported for namespace registration, not as a public catalog of translation keys.
+
+## 3. Update widget registration and standard tree components
+
+`createTreeWidget` and `TreeWidgetComponent` now require `localization`. In v4, `getLabel` receives `standardLabels`, which contains the localized default names for the built-in trees. The `render` callback no longer receives `density`; it now receives a required `treeLabel` and an optional `searchText`.
+
+`treeLabel` is the resolved label of the tree definition and must be forwarded to the tree component. `searchText` is the widget header's current search string and is supplied when the definition sets `isSearchable`.
+
+`ModelsTreeComponent`, `CategoriesTreeComponent`, `IModelContentTreeComponent`, and `ExternalSourcesTreeComponent` now require an accessible `treeLabel`. The v3 `getSchemaContext` and `density` props were removed from those standard tree components. `density` was also removed from `createTreeWidget`, `TreeWidgetComponent`, `SelectableTree`, `TreeWithHeader`, `Tree`, and `VisibilityTree`.
```
grigasp: I don't understand the criteria that these 3 paragraphs are split by. 
- `treeLabel` is mentioned in all 3 of them,
- `searchText` is mentioned in 1 and 2,
- `density` is mentioned in 1 and 3.

Maybe a bullet list would work better than paragraphs here?

JonasDov: updated

packages/tree-widget/MigrationGuide.md
JonasDov: updated

packages/tree-widget/MigrationGuide.md
JonasDov: moved

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,841 @@
...
+    "@mui/material": "^9.0.0",
+    "@stratakit/icons": "^0.4.0",
+    "@stratakit/mui": "^0.5.3",
+    "@stratakit/structures": "^0.5.6",
+
+    // React packages
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+  },
+}
```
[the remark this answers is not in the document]
JonasDov: updated

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,988 @@
...
+| `@itwin/presentation-shared` v1            | `@itwin/presentation-shared` v2            |
+| `@itwin/unified-selection` v1              | `@itwin/unified-selection` v2              |
+
+When these v2 packages are prereleases, use versions compatible with the installed tree-widget version rather than selecting their versions independently.
+
+## 2. Replace global initialization with `TreeWidgetContextProvider`
+
+`TreeWidget.initialize()` and `TreeWidget.terminate()` are removed. Delete both calls. Register localization namespaces during application startup, then place one `TreeWidgetContextProvider` above the tree-widget components rendered directly by the application. The provider owns the shared resources for its lifetime, so no replacement shutdown call is needed.
+
+`createTreeWidget` and `TreeWidgetComponent` add this provider around their content automatically when they receive `localization`. No additional provider is needed solely for either API. If the application already has one shared provider near its root, they may be rendered beneath it.
```
grigasp: I'm not a fan of this language style... Would prefer something like this:

> 1. Register tree widget's localization namespaces:
>    ```ts
>    import { LOCALIZATION_NAMESPACES } from "@itwin/tree-widget-react";
>    for (const namespace of LOCALIZATION_NAMESPACES) {
>      await IModelApp.localization.registerNamespace(namespace);
>    }
>    ```
> 2. Replace `TreeWidget` initialization and termination with `TreeWidgetContextProvider` React context:
>    - Remove `TreeWidget.initialize()` and `TreeWidget.terminate` calls.
>    - When using `createTreeWidget` or `TreeWidgetComponent`, make sure to pass `localization: IModelApp.localization` prop.
>    - When rendering lower level components directly, make sure they have `TreeWidgetContextProvider` React context available to them.

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,988 @@
...
+    </TreeWidgetContextProvider>
+  );
+}
+```
+
+### Replace `TreeWidget` localization helpers
+
+The `TreeWidget.i18n`, `TreeWidget.i18nNamespace`, and `TreeWidget.translate` static APIs were removed. There is no direct replacement: tree-widget translation keys are package implementation details and should not be resolved by consumers.
+
+Continue registering `LOCALIZATION_NAMESPACES` and pass the application localization service to `TreeWidgetContextProvider`, `createTreeWidget`, or `TreeWidgetComponent`. Tree-widget components then resolve their own strings internally. Use the `standardLabels` supplied to `TreeDefinition.getLabel` for standard tree names.
```
grigasp: The first sentence in this paragraph is repeating what's already described in the above section.

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,988 @@
...
+```tsx
+// v3
+const { buttonProps } = useCategoriesTreeButtonProps({ viewport });
+
+<CategoriesTreeComponent.ShowAllButton viewport={buttonProps.viewport} categories={buttonProps.categories} />;
+
+// v4 - `models` was added
+const { buttonProps } = useCategoriesTreeButtonProps({ viewport: treeWidgetViewport });
+
+<CategoriesTreeComponent.ShowAllButton viewport={buttonProps.viewport} categories={buttonProps.categories} models={buttonProps.models} />;
```
grigasp, line 308: maybe also provide an alternative, simpler example with `{...buttonProps}`?

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,988 @@
...
+```
+
+Custom header-button renderers no longer receive `density` or `onFeatureUsed`. Remove either property when destructuring or forwarding the renderer props, and do not pass them when rendering standard buttons directly.
+
+`onCategoriesFiltered` now takes a single object argument holding the filtered `categories` and the `models` those categories belong to, instead of the categories array. Both values may be `undefined` when filtering is cleared:
```
grigasp, line 321: feels like this should be a separate section

packages/tree-widget/MigrationGuide.md
```diff
@@ -0,0 +1,988 @@
...
+    ];
+
+    return HierarchySearchTree.createFromPathsList(paths);
+  },
+});
+```
+
+`createFromPathsList` merges shared path prefixes and returns a `Promise<HierarchySearchTree[]>`, so it can be returned directly from `getSearchPaths` or `getSubTreePaths`.
+
+Do not pass a v1 `HierarchyFilteringPath[]` to `createFromPathsList` unchanged. Re-type the paths as v2 `HierarchySearchPath` values and migrate their options. The identifier shapes are unchanged, but the option meanings differ:
```
grigasp: need to do something with versions - now we have v1, v2, v3 and v4, which is very confusing... it should be made clear that v3 widget is paired with v1 hierarchies and v4 widget is paired with v2 hierarchies.

JonasDov: Updated

packages/tree-widget/MigrationGuide.md
JonasDov: Moved to a separate section.

packages/tree-widget/MigrationGuide.md
JonasDov: Added

packages/tree-widget/MigrationGuide.md
JonasDov: Removed

packages/tree-widget/MigrationGuide.md
JonasDov: Updated

grigasp (approved):

PR #1794, opened 2026-08-28
[Tree widget]: models tree loading improvement on `next`

### LINKED ISSUE #1793: Tree widget: Don't wait for all model-categories when loading tree nodes
Models, Categories and Classifications trees use a lot of similar data. This data is queried once and then stored in a shared cache. 
The shared cache contains more data than what is needed to load tree nodes. The goal is:
1. If cached data exists - use it (this way queries are the fastest).
2. If cached data does not exist - modify node loading queries to not need it (this way queries are slower, but the overall time for initial load is faster).

Theoretical example: Shared cache takes 5s, to query and cache shared data. This cached data is then used when loading tree nodes, so node loading queries ~ 300ms.
With these changes if shared data is not yet loaded, execute a slightly different query to get tree nodes: this slower query takes 1s. Overall, the tree nodes are now loaded faster on first load: previously 5.3s, now 1s. But when cached data exists, queries are still as performant as before.

JonasDov:
Part of #1793.
Updated models tree. 
Categories and classifications tree root nodes still require element-models-categories query to finish. They are more tightly coupled with this cache. @grigasp should we try to adjust categories/classifications trees also?

### DIFF
--- a/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
+++ b/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
@@ -3,7 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { EMPTY, filter, forkJoin, from, identity, map, mergeMap, of, reduce, shareReplay } from "rxjs";
+import { EMPTY, filter, forkJoin, from, identity, map, mergeMap, of, reduce, shareReplay, tap } from "rxjs";
 import { Guid } from "@itwin/core-bentley";
 import { fromWithRelease } from "../Rxjs.js";
 import { ChildrenTree, getOrCreate } from "../Utils.js";
@@ -37,6 +37,7 @@ export class BaseIdsCache {
   readonly #subCategoriesCache: SubCategoriesCache;
   #elementClassName: EC.FullClassNameDotNotation;
   #modeledElementsCache: Observable<ModeledElementsCache> | undefined;
+  #modeledElementsLoaded = false;
   readonly #elementModelCategoriesCache: ElementModelCategoriesCache;
   #categoryModelsInfoWithoutSubModels:
     | Observable<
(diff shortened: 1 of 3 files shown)


packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
```diff
@@ -93,7 +94,12 @@ export class BaseIdsCache {
       ),
       shareReplay(),
     );
-    return this.#modeledElementsCache.pipe(mergeMap((modeledElementsCache) => modeledElementsCache.getModeledElementsInfo()));
+    return this.#modeledElementsCache.pipe(
+      mergeMap((modeledElementsCache) => modeledElementsCache.getModeledElementsInfo()),
+      tap(() => {
+        this.#modeledElementsLoaded = true;
+      }),
```
grigasp, line 101: shouldn't this be done before `shareReplay` above?

packages/tree-widget/src/tree-widget-react/trees/models-tree/ModelsTreeDefinition.ts
```diff
@@ -603,20 +605,30 @@ export class ModelsTreeDefinition implements HierarchyDefinition {
...
+                : `
+                  JOIN ${CLASS_NAME_GeometricElement3d} ce ON ce.Category.Id = this.ECInstanceId
+                  JOIN IdSet(?) modelIdSet ON ce.Model.Id = modelIdSet.id
+                `
+            }
             ${categoryInstanceFilterClauses.joins}
-            ${createWhereClause({ conditions: [categoryInstanceFilterClauses.where] })}
+            ${createWhereClause({
+              conditions: [categoryInstanceFilterClauses.where, !categoriesToShow && "ce.Parent.Id IS NULL"],
+            })}
```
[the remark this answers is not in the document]
JonasDov: fixed

packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
JonasDov: Don't think so, that `shareReplay` only caches the `ModeledElementsCache` instance. The flag should be set after `getModeledElementsInfo()` emits; otherwise it could report that modeled-element data is loaded before the inner cache has resolved.

grigasp: ah, right, at first glance I thought we only want to wait for `this.getAllModels()`... the current implementation will call the `tap` callback many times, but I guess that's not a big deal.

grigasp (approved):

PR #1803, opened 2026-09-08
[measure-tools-react] Fix measurements in profile drawings

Maxime-Brassard:
On some profile drawings, measurements were incorrect. For example, a 10ft measurement would come out as 0.08ft.
This would happen in sheets with more than 1 drawing as, when trying to get the transform, the algorithm would output the first encountered drawing. If that drawing was not a profile drawing (and therefore did not contain the profileToWorld transform) we'd be stuck in sheet coordinates with an undefined transform.

Also, now, we give priority to element detection when starting a measurement in a drawing. This means that, when there is a element, we check the associated drawing instead of doing it by using the drawing box.


a-gagnon (approved):
LGTM

PR #1804, opened 2026-09-09
[Tree widget]: preload tree caches on `next`

### LINKED ISSUE #1793: Tree widget: Don't wait for all model-categories when loading tree nodes

JonasDov:
Closes #1793 
Based on https://github.com/iTwin/viewer-components-react/pull/1801#discussion_r3957155834

### DIFF
--- a/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
+++ b/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
@@ -104,10 +104,15 @@ export class BaseIdsCache {
 
   // Implement methods using each cache
 
-  public async preloadCachedData(): Promise<void> {
+  public async preloadModeledElements(): Promise<void> {
     try {
-      // Loading modeled elements also preloads element-model-category data.
       await toVoidPromise(this.getModeledElementsInfo());
+    } catch {}
+  }
+
+  public async preloadElementModelCategories(): Promise<void> {
+    try {
+      await toVoidPromise(this.#elementModelCategoriesCache.getCachedData());
     } catch {}
   }
 
@@ -452,8 +457,12 @@ export class BaseIdsCacheImpl {
 
   // Implement IBaseIdsCache by re-exporting BaseIdsCache methods
 
-  public async preloadCachedData(): ReturnType<BaseIdsCache["preloadCachedData"]> {
-    return this.#baseIdsCache.preloadCachedData();
+  public async preloadModeledElements(): ReturnType<BaseIdsCache["preloadModeledElements"]> {
+    return this.#baseIdsCache.preloadModeledElements();
+  }
+
+  public async preloadElementModelCategories(): ReturnType<BaseIdsCache["preloadElementModelCategories"]> {
+    return this.#baseIdsCache.preloadElementModelCategories();
   }
 
   public getChildElements(props: Props<BaseIdsCache["getChildElements"]>): ReturnType<BaseIdsCache["getChildElements"]> {
--- a/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
+++ b/packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
@@ -5,7 +5,7 @@
 
 import { EMPTY, filter, forkJoin, from, identity, map, mergeMap, of, reduce, shareReplay, tap } from "rxjs";
 import { Guid } from "@itwin/core-bentley";
-import { fromWithRelease } from "../Rxjs.js";
+import { fromWithRelease, toVoidPromise } from "../Rxjs.js";
 import { ChildrenTree, getOrCreate } from "../Utils.js";
 import { ChildElementsCache } from "./ChildElementsCache.js";
 import { DescendantsCountCache } from "./DescendantsCountCache.js";
@@ -103,6 +103,13 @@ export class BaseIdsCache {
   }
 
   // Implement methods using each cache
+
+  public async preloadCachedData(): Promise<void> {
+    try {
+      // Loading modeled elements also preloads element-model-category data.
+      await toVoidPromise(this.getModeledElementsInfo());
+    } catch {}
+  }
 
   // ModeledElementsCache methods
   public getCategoryModeledElements({
@@ -444,6 +451,10 @@ export class BaseIdsCacheImpl {
   }
 
   // Implement IBaseIdsCache by re-exporting BaseIdsCache methods
+
+  public async preloadCachedData(): ReturnType<BaseIdsCache["preloadCachedData"]> {
+    return this.#baseIdsCache.preloadCachedData();
+  }
 
   public getChildElements(props: Props<BaseIdsCache["getChildElements"]>): ReturnType<BaseIdsCache["getChildElements"]> {
     return this.#baseIdsCache.getChildElements(props);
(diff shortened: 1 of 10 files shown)


packages/tree-widget/src/tree-widget-react/trees/classifications-tree/ClassificationsTreeDefinition.ts
```diff
@@ -132,7 +132,12 @@ export class ClassificationsTreeDefinition implements HierarchyDefinition {
         childNodes: [
           {
             parentInstancesNodePredicate: CLASS_NAME_ClassificationTable,
-            definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) => this.#createClassificationTableChildrenQuery(requestProps),
+            definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) => {
+              if (requestProps.parentNodeInstanceIds.length > 0) {
+                void this.#props.getIdsCache(this.#props.imodelAccess.imodelKey).preloadCachedData();
+              }
+              return this.#createClassificationTableChildrenQuery(requestProps);
+            },
```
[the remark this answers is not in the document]
JonasDov: changed

packages/tree-widget/src/tree-widget-react/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -336,15 +336,24 @@ export class CategoriesTreeDefinition implements HierarchyDefinition {
               : [
                   {
                     parentInstancesNodePredicate: this.#categoryClass,
-                    definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) => this.createCategoryChildrenQuery(requestProps),
+                    definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) => {
+                      if (requestProps.parentNodeInstanceIds.length > 0) {
+                        void this.#idsCache.preloadCachedData();
+                      }
```
grigasp: this will execute the model-categories query even through the tree doesn't need that (doesn't show element nodes), right?

packages/tree-widget/src/tree-widget-react/trees/categories-tree/CategoriesTreeDefinition.ts
```diff
@@ -336,15 +336,24 @@ export class CategoriesTreeDefinition implements HierarchyDefinition {
...
                   },
                 ]),
             ...(isDefinitionContainerSupported
               ? [
                   {
                     parentInstancesNodePredicate: CLASS_NAME_DefinitionContainer,
-                    definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) =>
-                      this.createDefinitionContainersAndCategoriesQuery(requestProps),
+                    definitions: async (requestProps: DefineInstanceNodeChildHierarchyLevelProps) => {
+                      if (requestProps.parentNodeInstanceIds.length > 0) {
```
grigasp: why do we need the `requestProps.parentNodeInstanceIds.length > 0` check?

packages/tree-widget/src/tree-widget-react/trees/classifications-tree/ClassificationsTreeDefinition.ts
```diff
@@ -238,6 +238,9 @@ export class ClassificationsTreeDefinition implements HierarchyDefinition {
       return [];
     }
     const cache = this.#props.getIdsCache(imodelKey);
+    if (classificationTableIds.length > 0) {
+      void cache.preloadCachedData();
+    }
```
grigasp: why not at a similar place like other hierarchy definitions, within the call to `createPredicateBasedHierarchyDefinition`?

packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
```diff
@@ -104,6 +104,13 @@ export class BaseIdsCache {
 
   // Implement methods using each cache
 
+  public async preloadCachedData(): Promise<void> {
```
grigasp: How can we decide what to preload at the base ids cache level? Some trees may not need element-model-category data (or am I wrong?), so we'd be executing this for no reason. I'm thinking that maybe each tree-specific cache should make that decision. Or maybe not even a cache, but something that initiates preloading, could call cache-specific methods based on hierarchy configuration.

grigasp (commented):
I'm having some doubts about whether hierarchy definition should initiate preloading. Maybe that should be done by the `use*Tree` hook on node expansion? @saskliutas, opinion?

packages/tree-widget/src/tree-widget-react/trees/categories-tree/CategoriesTreeDefinition.ts
JonasDov: Changed, now we preload DefinitionContainers and subModels if elements are shown in the categories tree. 
>the tree doesn't need that 

We do use model-categories when loading definition containers data.

packages/tree-widget/src/tree-widget-react/trees/categories-tree/CategoriesTreeDefinition.ts
JonasDov: On a second thought we don't, removed

packages/tree-widget/src/tree-widget-react/trees/classifications-tree/ClassificationsTreeDefinition.ts
JonasDov: We need to get the cache based on imodel (cache is retrieved on a line above). Other trees use a single imodel.

packages/tree-widget/src/tree-widget-react/shared/internal/caches/BaseIdsCache.ts
JonasDov: Agree. 
- Added separate methods.
- Unecessary data is no longer preloaded.

packages/tree-widget/src/tree-widget-react/trees/categories-tree/UseCategoriesTree.tsx
```diff
@@ -108,6 +109,18 @@ export function useCategoriesTree({
     hierarchyConfig: hierarchyConfiguration,
   });
 
+  const onNodeExpanded = useCallback(
+    (node: TreeNode) => {
+      if (node.nodeData.parentKeys.length === 0) {
```
saskliutas: Why do we need this check?

JonasDov: Well, we are calling `cache.preload*()`. It in turn converts observable into void promise and awaits it. By having this check none of that is done.
I am fine removing it if you think it's better without

saskliutas: Well, it does not ensure that preload is called only once. There might be multiple root nodes so expanding each would result in preload call.

JonasDov: Changed, check is now in `preload` instead.

saskliutas (approved):

PR #1810, opened 2026-09-15
[Test viewer]: add imodel selector and theme switch on `next`

JonasDov:
- Added `IModelSelector`: it allows switching between iModels of an iTwin without the need to manually change the url.
- Added `ColorThemeDropdown`: it allows switching active color theme in the test app: `system`/`light`/`dark`

Changes were made on `next` branch due to stratakit+mui. We could make this available in `master` with iTwinUI but then it would have to be migrated to strata+mui


apps/test-viewer/src/components/IModelSelector.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  if (!iTwinId || !iModelId) {
+    return null;
+  }
+
+  const currentState = state?.iTwinId === iTwinId ? state : undefined;
+  const loading = !currentState || currentState.status === "loading";
+  const failed = currentState?.status === "error";
+  const options = currentState?.status === "ready" ? currentState.options : [];
+  const currentOption = { value: iModelId, label: iModelId };
```
saskliutas: When is this needed? Is there a case where selected iModel is not in loaded iModels list?

apps/test-viewer/src/components/Viewer.tsx
```diff
@@ -54,14 +55,15 @@ export function Viewer() {
   return (
     <ViewerOptionsProvider>
       <ViewerWithOptions />
+      <IModelSelector />
```
saskliutas: Maybe make it status bar button instead of rendering on top of AppUI? https://github.com/iTwin/viewer-components-react/blob/e689f553ae36660ad0f1fdec922d9b9aee5f93b4/apps/test-viewer/src/components/ViewerOptions.tsx#L48

apps/test-viewer/src/components/IModelSelector.tsx
JonasDov: That was a fallback for a when provided iModelId in the url was missing from the fetched list. Normally this shouldn't happen, removed.

apps/test-viewer/src/components/Viewer.tsx
JonasDov: Moved `IModelSelector` to statusBarActions

packages/tree-widget/src/e2e-tests/CategoriesTree.test.ts-snapshots/Categories-tree-node-with-active-filtering---information-message-1-chromium-linux.png
saskliutas, line 1: @MartynasStrazdas could you check if this change is expected and screenshots where outdated?

MartynasStrazdas: The new screenshots seem off the padding in the new ones don't exist (are different) in presentation. It seems for some reason button css has priority over errorHeader css (both come from strata). Would ignore the screenshot updates (let them pass with these ones) and come back to this when we update to v1.

saskliutas (approved):

